PR #67, opened 2022-01-13
they_said_so: Switched to the RSS feed.

henryso:
This uses the FeedBurner RSS feeds instead of the API.  This also has the advantage that it gets all categories in one fetch and caches all of them at once.


henryso:
The repository links got messed up, so I will close this and reopen from a new fork of tidbyt/community.

betterengineering:
Oh no! So sorry about that @henryso - I had no idea that would happen. If I had, I would have gotten this landed before opening up the repo 😢

henryso:
Heh.  I had no idea that would happen either...

PR #69, opened 2022-01-14
Added "Clock By Henry" app

henryso:
At the risk of yet another clock application, here's mine. I wanted to show date and time with big numbers for the time. This one changes the time color for day and night (though it's currently hard-coded for 6 AM-6 PM for day and 6 PM-6 AM for night) and the date color by season.

clock_by_henry

This one probably needs quite a bit of configurability which I don't currently know how to do properly. It will use the $tz thing for the time zone but should probably take a location like the other clock apps. The colors are also configurable by parameter.

In the future, I'll probably add a 12h/24h control, day/night hour control, and some date format options. It'd also be nice to have some i18n for the date, but I don't know how to do that (yet?).

(On a more selfish note, this is one I'd like available so I don't need to keep pushing to the Tidbyt every minute even though it cycles through my apps approximately every 3 minutes.)


henryso:
I put this on in my rush to re-connect my repository fork, and now realize that I need to work on the Schema stuff before this can proceed.   Should I leave this here or close it for now and re-open when I have the schema work done?

betterengineering:
Feel free to keep it open and add the schema in when you get a chance! Will keep checking back on it

apps/clockbyhenry/clock_by_henry.star
```diff
@@ -0,0 +1,499 @@
...
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Location(
+                id = P_LOCATION,
+                name = "Location",
+                desc = "Location for the display of date and time.",
+                icon = "place",
+            ),
+            schema.Text(
```
betterengineering, line 155: These colors should probably be a dropdown, which sucks because we'd have to limit the number of colors as well and presumably that's why you did it this way 😅 . We can work on getting you a color picker but it's going to take some time on our end. I'm hesitant to publish this with a text input because I imagine it will be pretty confusing to someone trying to enter the colors

apps/clockbyhenry/clock_by_henry.star
```diff
@@ -0,0 +1,499 @@
...
+    if season >= 320 and season < 621:
+        season = 1
+    elif season >= 621 and season < 923:
+        season = 2
+    elif season >= 923 and season < 1221:
+        season = 3
+    else:
+        season = 0
+
+    phase = int(now.format("15"))
```
betterengineering, line 78: nit: you could use the new sunrise/sunset timer!

betterengineering (changes requested):
Hey @henryso! So sorry for the delay on this one. I love that this clock changes colors for the seasons ❤️ 

We've been dragging our feet because we're trying to figure out what is style with these apps and what's going to add a support burden on our team. 

I think the colors being a text entry are going to be hard to navigate as an end user and we should update it to a smaller list of colors as a drop down for each of the fields. We can work on getting you a color picker which we've been needing anyways that we can upgrade to once we have it to provide even more options to the user.

Other then that it looks good and we can land once we get the changes on how colors are selected.

apps/clockbyhenry/clock_by_henry.star
henryso: Ok, lets leave this hanging until the color picker materializes

apps/clockbyhenry/clock_by_henry.star
henryso: I actually *don't* want it to go by sunrise and sunset.  I should probably make this configurable.  I have time to think about it since the color picker is still up in the air.

henryso:
@betterengineering If you want, I can close this and re-create it once the color picker is available.

henryso:
Another option: How would you feel about not having individual color configurations until the color picker is available? I'd leave the theme selection but remove the individual color fields.

henryso:
This takes out the individual color configurations.  I'll put it back in when the color picker is available.

apps/clockbyhenry/clock_by_henry.star
henryso: I made this configurable.

apps/clockbyhenry/clock_by_henry.star
henryso: I took out these configurations for now.

betterengineering (approved):
Thanks for the changes @henryso! Going to get this landed and will work on getting you a color picker!

PR #70, opened 2022-01-14
digital_rain: More rain and colorful rain

henryso:
This adds a new parameter to the schema for the `char_size` to select smaller size characters. This also fixes a bug with the selection of the blue color.
 
It gets slower as the characters get smaller (and more images need to be rendered). Please let me know if the performance is unacceptable, and I'll remove the `tinier` option.

This also adds the ability to produce `multicolor` rain.


apps/digitalrain/digital_rain.star
```diff
@@ -101,44 +100,50 @@ def main(config):
...
     seed = [seed]
 
+    # get the color; do it this way so that setting the color from the
+    # config doesn't spoil the pseudo-random number sequence
+    color_options = (
+        [i for i in range(COLOR_COUNT)] +
+        [rand(seed, COLOR_COUNT + 1) - 1, rand(seed, COLOR_COUNT)]
+    )
+    color_number = COLOR_NAMES.get(config.get("color"), COLOR_NAMES["random"])
+    if color_number >= 0:
```
betterengineering, line 110: Is there ever a case this would be less then zero? Or just in case someone tries to through pixlet?

betterengineering (approved):
This looks fine to me! Though I didn't look too closely at the generation code - I figured you have that part figured out :D

apps/digitalrain/digital_rain.star
henryso: The `multicolor` has a value of -1.

PR #71, opened 2022-01-14
they_said_so: Switched to the RSS feed.

henryso:
This uses the FeedBurner RSS feeds instead of the API. This also has the advantage that it gets all categories in one fetch and caches all of them at once.


betterengineering:
Hey @henryso! We're still willing to pay for the API - we're just held up on secret support which was held up on schema being open sourced 😅 . Do you want to wait it out or do you want to land this and launch it with they said so attribution still?

henryso:
@betterengineering They still haven't fixed the SSL issue on their API, so it would still be a problem even if Schema support were in place right now.  I think it's better with the feeds anyway (one request instead of 8 and cached) and I kind of like the attribution.  I can always change things when (if?) Google decides to change how feedburner works for RSS clients.

apps/theysaidso/they_said_so.star
```diff
@@ -27,16 +27,17 @@ Author: Henry So, Jr.
...
 load("http.star", "http")
 load("render.star", "render")
 load("schema.star", "schema")
 load("cache.star", "cache")
+load("xpath.star", "xpath")
 load("encoding/base64.star", "base64")
 load("encoding/json.star", "json")
 
-URL = "http://quotes.rest/qod.json?category="
+URL = "http://feeds.feedburner.com/theysaidso/qod"
```
betterengineering, line 40: Oh, can we swap back to HTTPS now?

apps/theysaidso/they_said_so.star
```diff
@@ -72,38 +75,38 @@ def main(config):
...
-                    "quote": 'Forsooth, the server quoteth "%s".' % content.status_code,
-                    "author": "Anonymous",
+                data = {
+                    c: {
+                        "quote": 'Forsooth, the server quoteth "%s".' % content.status_code,
+                        "author": "Anonymous",
+                    }
+                    for c in CATEGORIES
                 }
+            cache.set(KEY, json.encode(data), TTL if content.status_code < 300 else ERROR_TTL)
```
betterengineering, line 100: 👌

betterengineering (approved):

PR #74, opened 2022-01-14
Initial Commit of Nightscout applet

jer-tav:
This is the Nightscout application for Tidbyt!  This originally spawned from a community request [here](https://discuss.tidbyt.com/t/need-a-developer/997).

[Nightscout](https://nightscout.github.io/) is an Open Source Continuous Glucose Monitoring application that is used by diabetics to monitor their blood glucose levels in real time.  This data is uploaded to the cloud, and can then be accessed using a simple REST API.

The app takes the data from the REST API (based on the user's ID), and then displays the trend arrow (one of double up, up, 45 degree up, flat, 45 degree down, down and double down), along with the current BG reading and a delta from the previous reading.

The arrow and current BG reading is color coded red, yellow or green based on the parameters passed to the application (see below for these thresholds).  

If the current data is more than 15 minutes out of date, the arrow and current BG reading are coloured grey.

| Argument Name | Description | Required / Optional | Default |
| --- | --- | --- | --- | 
| nightscout_id | The Nightscout ID for the user | Required | None
| normal_high | The normal high value - anything above this is considered yellow unless it is above the urgent_high | Optional | 150 |
| normal_low | The normal low value - anything below this is considered yellow unless it is below the urgent_low | Optional | 100 |
| urgent_high | The urgent high value - anything above this is considered red | Optional | 200 |
| urgent_low | The urgent low value - anything below this is considered red | Optional | 70 |


jer-tav:

recheck

apps/nightscout/nightscout.star
```diff
@@ -0,0 +1,210 @@
...
+        print("Hit - displaying cached data")
+
+        # Pull the data from the cache
+        sgv_current = int(nightscout_data_json["sgv_current"])
+        sgv_delta = int(nightscout_data_json["sgv_delta"])
+        latest_reading_dt = time.parse_time(nightscout_data_json["latest_reading_date_string"])
+        trend = nightscout_data_json["trend"]
+    else:
+        print("Miss - calling Nightscout API")
+        nightscout_url = "https://" + nightscout_id + ".herokuapp.com/api/v1/entries.json"
```
betterengineering, line 94: This url created from user input makes me pretty nervous. We really want to avoid the case where someone could make calls to another service that we don't intend or even worse - make calls to our internal systems. Can we add validation to ensure this is actually a valid ID for heroku? We could either do it through inspection of the string or we could potentially use a heroku API if they offer it?

betterengineering, line 94: This url created from user input makes me pretty nervous. We really want to avoid the case where someone could make calls to another service that we don't intend or even worse - make calls to our internal systems. Can we add validation to ensure this is actually a valid ID for heroku? We could either do it through inspection of the string or we could potentially use a heroku API if they offer it?

apps/nightscout/nightscout.star
```diff
@@ -0,0 +1,210 @@
...
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Text(
+                id = "nightscout_id",
+                name = "Nightscout ID",
+                desc = "Your Nightscout ID (i.e. XXX in https://XXX.herokuapp.com/api/v1/entries.json)",
+                icon = "user"
+            ),
+            schema.Text(
```
betterengineering, line 36: Would a `schema.Number` be useful here? Would be happy to get it added if it would be! It's non-blocking for this PR though

apps/nightscout/nightscout.go
```diff
@@ -0,0 +1,25 @@
...
+var source []byte
+
+// New creates a new instance of the Nightscout applet.
+func New() manifest.Manifest {
+	return manifest.Manifest{
+		ID:          "nightscout",
+		Name:        "Nightscout",
+		Author:      "Jeremy Tavener",
+		Summary:     "Shows Nightscout CGM Data",
+		Desc:        "Displays Continous Glucose Monitoring (CGM) data from the Nightscout Open Source project (https://nightscout.github.io/).",
```
betterengineering, line 20: One thing that makes me pretty nervous is the failure cases around Tidbyt today. If we fail to render an app, the Tidbyt will happily continue to serve an instance from it's local cache forever until it gets an updated copy. It's happened before with the clock where we continue to show a stale time which is really a poor use case. We're working to be able to set a freshness setting for each app to be able to determine how stale it can be, but we don't have it just yet.

Is this app ok to be stale? Should we put a warning in the description saying it could be stale if Tidbyt has problems? Do we need to wait until we get the ability to only serve fresh renders?

apps/nightscout/nightscout.star
```diff
@@ -0,0 +1,210 @@
...
+        latest_reading_dt = time.parse_time(nightscout_data_json["latest_reading_date_string"])
+        trend = nightscout_data_json["trend"]
+    else:
+        print("Miss - calling Nightscout API")
+        nightscout_url = "https://" + nightscout_id + ".herokuapp.com/api/v1/entries.json"
+
+        # Request latest entries from the Nightscout URL
+        resp = http.get(nightscout_url)
+
+        if resp.status_code != 200:
```
betterengineering, line 99: Could we add a check here for a 404? Or whatever the status code returned for a missing Heroku instance. Right now, we'd get alerted if someone had a misconfigured instance or misspelled ID. One thing we could also consider is:
```
return []
```
Which will remove the app from the rotation and not cause an error on our side. This is what we do for Spotify for example.

betterengineering (commented):
Thank you so much for putting this together! I know this is so important to a lot of folks and we've had numerous people asking for support for it. I want to take this one a bit slower just to make sure we get it right because this information seems pretty important to get right. I think we can support this app and we're committed to doing everything we need to do on our end to make it happen. Would love to hear your thoughts on the comments added!

jer-tav:
I get what you're saying  - someone could potentially call something
maliciously though (for example, if I put my ID as "maliciouswebsite.com#"
it would do bad things).

I can do a couple of things -
1. Validate the ID is only alphanumeric characters
2. A test call to https://XXX.herokuapp.com/api/v1/status.json first -
that's specific to the Nightscout API (
https://lk7.herokuapp.com/api/v1/status.json).
3. Blacklist any Tidbyt internal hostnames (if you can provide a list :) )

apps/nightscout/nightscout.star
jer-tav: Yes, that would be **AWESOME**!

apps/nightscout/nightscout.go
jer-tav: I could rework the UI to also display a "Last Updated" field.  Then if for some reason it's using a local cached version, the information is still accurate (albeit old).

apps/nightscout/nightscout.star
betterengineering: Replying to your comment - alpha numeric sanity check should be good enough for now. Is it an issue that anyone could punch in the right string and get this data??

jer-tav: I get what you're saying  - someone could potentially call something maliciously though (for example, if I put my ID as "maliciouswebsite.com#" it would do bad things).

I can do a couple of things -

1.  Validate the ID is only alphanumeric characters
2.  A test call to https://XXX.herokuapp.com/api/v1/status.json first - that's specific to the Nightscout API (https://lk7.herokuapp.com/api/v1/status.json).
3. Blacklist any Tidbyt internal hostnames (if you can provide a list :) )

jer-tav: Sorry, replied by email first.  I don't _think_ it's an issue, but I can check with Dave.  There's nothing stopping me just hammering in IDs to the Heroku app right now until I get a hit.  I'm fairly sure you have to actively enable this information to be accessible via API with Nightscout, so they probably cover the privacy aspect of it.

apps/nightscout/nightscout.star
jer-tav: For sure - that's a great idea.

apps/nightscout/nightscout.go
betterengineering: Ohhh, I like that a lot. Displaying last updated would make me feel a lot more comfortable

apps/nightscout/nightscout.star
jer-tav: I guess the other thing to note is there's nothing that constitutes HIPAA/GDPR data available in the APIs that Nightscout API

jer-tav: Would the preference be to `fail` or render something about an invalid ID on the Tidbyt here?

apps/nightscout/nightscout.star
```diff
@@ -147,34 +163,34 @@ def main(config):
...
...uaWlkOjNiZDEwNTMwLWIyYzAtNGM0My1hYmNlLWFlMWMxOWRjY2RjMSIgc3RFdnQ6d2hlbj0iMjAyMS0xMi0yM1QxMzoxOTowOS0wNjowMCIgc3RFdnQ6c29mdHdhcmVBZ2VudD0iQWRvYmUgUGhvdG9zaG9wIDIyLjMgKFdpbmRvd3MpIi8+IDxyZGY6bGkgc3RFdnQ6YWN0aW9uPSJzYXZlZCIgc3RFdnQ6aW5zdGFuY2VJRD0ieG1wLmlpZDphMDU2YTljNS0zYWQ5LTkwNDMtYTU1Yi0wOGQxMDgxNGRlOTgiIHN0RXZ0OndoZW49IjIwMjEtMTItMjNUMTM6MTk6MDktMDY6MDAiIHN0RXZ0OnNvZnR3YXJlQWdlbnQ9IkFkb2JlIFBob3Rvc2hvcCAyMi4zIChXaW5kb3dzKSIgc3RFdnQ6Y2hhbmdlZD0iLyIvPiA8cmRmOmxpIHN0RXZ0OmFjdGlvbj0ic2F2ZWQiIHN0RXZ0Omluc3RhbmNlSUQ9InhtcC5paWQ6YjBiZWFiY2EtNzQ1Mi03MDRkLWEzMGQtMTU1ZjdlYWRjYWFkIiBzdEV2dDp3aGVuPSIyMDIxLTEyLTIzVDEzOjI3OjE4LTA2OjAwIiBzdEV2dDpzb2Z0d2FyZUFnZW50PSJBZG9iZSBQaG90b3Nob3AgMjIuMyAoV2luZG93cykiIHN0RXZ0OmNoYW5nZWQ9Ii8iLz4gPC9yZGY6U2VxPiA8L3htcE1NOkhpc3Rvcnk+IDxwaG90b3Nob3A6RG9jdW1lbnRBbmNlc3RvcnM+IDxyZGY6QmFnPiA8cmRmOmxpPmFkb2JlOmRvY2lkOnBob3Rvc2hvcDo3MTZjZmUwYS04ODRhLThjNGQtOWNiMy0wMjE4MTQyZWYwZTk8L3JkZjpsaT4gPC9yZGY6QmFnPiA8L3Bob3Rvc2hvcDpEb2N1bWVudEFuY2VzdG9ycz4gPC9yZGY6RGVzY3JpcHRpb24+IDwvcmRmOlJERj4gPC94OnhtcG1ldGE+IDw/eHBhY2tldCBlbmQ9InIiPz6jJiF4AAAAS0lEQVQ4je2TQQoAIAgEZ6P/f9kOFVEUBnbMk+vKiKAyMyKRAYS67jRVYXh+Co2HD/gAoF0i48JYtDz/2QraeDrkUy0dmq6Bir5zAaNDECOB89QJAAAAAElFTkSuQmCC"),
```
jer-tav: I optimized all the icons to try and make them a little smaller, hence the changes here.

apps/nightscout/nightscout.star
```diff
@@ -186,21 +202,45 @@ def main(config):
...
+                            ),
+                            render.Text(
+                                content="Updated:",
+                                font="tom-thumb"
+                            ),
+                            render.Marquee(
+                                width = 32,
+                                child=render.Text(
+                                    # This time is currently in UTC - need to figure out how to display in a selected timezone.
+                                    content=latest_reading_dt.format("2006-01-02 03:04:05 PM"),
```
jer-tav, line 223: Still todo: add location to the scheme and convert UTC to the appropriate timezone

apps/nightscout/nightscout.star
rohansingh: I think `fail` is fine for "permanent" errors, e.g., invalid config that can never result in a working app.

My opinion is only remediation 1 (validating ID) is required. As soon as that is provided, the app will make a call to the `entries.json` endpoint anyway. As for blocking internal hostnames, we can manage that our end.

So basically just something like:
```starlark
if re.search("[^0-9A-Za-z\-]", nightscout_id):
  fail("not a valid ID:", nightscout_id)
```

Haven't tested that but I think it's roughly what it should look like.

apps/nightscout/nightscout.star
```diff
@@ -0,0 +1,250 @@
...
+        nightscout_url = "https://" + nightscout_id + ".herokuapp.com/api/v1/entries.json"
+
+        # Request latest entries from the Nightscout URL
+        resp = http.get(nightscout_url)
+
+        if resp.status_code == 502:
+            print("Page not found for nightscout ID " + nightscout_id + " - is this ID correct?")
+            return []
+        elif resp.status_code != 200:
+            fail("Failed to retieve the Nightscout details with status " + str(resp.status_code))
```
betterengineering, line 116: Given this is infra set up by the user - I think it's best if we never `fail()` in this app and instead always return a message prompting the user that something is wrong with their setup. These `fails()` show up in our error rate that will cause alarms to fire if it's over a certain percentage

betterengineering (commented):
Thanks for making all of the changes! This should be good to go once the fails() are removed.

apps/nightscout/nightscout.star
rohansingh: @betterengineering Do you agree with what I said about `fail` above? Or is it a bad idea?

betterengineering:
Hey @jer-tav! I can fix this up and push to your branch if that's ok? Might be easier then the back and fourth

jer-tav:
@betterengineering - hey Mark, if you've got the time to do it, that would be great!  The last missing feature is to grab the location of the user via the schema and convert `latest_reading_dt` before displaying it, plus removing the fails I think.

Did you add number support to the schema as well?  

I'm not likely to get a chance to look at it for another week or so with my current work commitments.

betterengineering:
Updated! We don't yet have schema support for numbers, will get that added in the coming weeks and will get this updated.

PR #77, opened 2022-01-15
Add SBB app

LukiLeu:
This app generates a real-time timetable for any train station in the Swiss Public Transportation Network. It shows the departure time and destination of the next five trains that are leaving the chosen station. If a train is delayed, this delay is also shown.

The station can be set with the _station_ parameter. Additionally, with the parameter _skiptime_, it is possible to show departures in the future. As an example: If this parameter is set to _10_ Minutes, the departures between _now_ and _now+10 minutes_ are skipped and the next five departures departing after _now+10 minutes_ are shown.

sbb

### DIFF
--- a/apps/sbbtimetable/sbb_timetable.star
+++ b/apps/sbbtimetable/sbb_timetable.star
@@ -11,195 +11,195 @@ SBB_URL = "https://fahrplan.search.ch/api/stationboard.json?show_delays=1&transp
 FONT_TO_USE = "CG-pixel-3x5-mono"
 NO_FRAMES_TOGGLE = 20
 COLOR_CATEGORY = {
-	"IC": "#700",
-	"EC": "#700",
-	"IR": "#700",
-	"RE": "#700",
-	"RJX": "#700",
-	"TGV": "#700",
-	"NJ": "#700",
-	"S" : "#007",
-	"SN" : "#000",
-	
+    "IC": "#700",
+    "EC": "#700",
+    "IR": "#700",
+    "RE": "#700",
+    "RJX": "#700",
+    "TGV": "#700",
+    "NJ": "#700",
+    "S" : "#007",
+    "SN" : "#000",
+    
 }
 COLOR_DELAY = "#F00"
 
 def main(config):
-	# Get the config values
-	station = config.get("station", "Bern")
-	skiptime = config.get("skiptime", 0)
-	
-	# Check if we need to convert the skiptime
-	if type(skiptime) == "string":
-		skiptime = int(skiptime)
-	
-	# Check that the skip time is valid
-	if skiptime < 0:
-		skiptime = 0
+    # Get the config values
+    station = config.get("station", "Bern")
+    skiptime = config.get("skiptime", 0)
+    
+    # Check if we need to convert the skiptime
+    if type(skiptime) == "string":
+        skiptime = int(skiptime)
+    
+    # Check that the skip time is valid
+    if skiptime < 0:
+        skiptime = 0
 
-	# Check if we have the requested data in the cache
-	resp_cached = cache.get("sbb_%s" % station)
-	if resp_cached != None:
-		# Get the cached response
-		print("Hit! Displaying cached data.")
-		resp = json.decode(resp_cached)
-	else:
-		# Get a new reponse
-		print("Miss! Calling API.")
-		resp = http.get("%s%s" % (SBB_URL, station))
-		if resp.status_code != 200:
-			fail("Request failed with status %d", resp.status_code)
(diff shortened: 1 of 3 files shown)


LukiLeu:

LukiLeu:
Why is the buildifier failing here? I don't see the reason.

henryso:
You can click on the Details links to see the errors.

The first thing you should do is merge your repository's main branch with the tidbyt repo's main branch.  This should fix the build-and-test failure.

The second thing you should do is run `make lint` which should fix the lint errors, or you can alternately change your file manually.

LukiLeu:
> You can click on the Details links to see the errors.
> 
> The second thing you should do is run `make lint` which should fix the lint errors, or you can alternately change your file manually.

I did run `make lint` on my ubuntu machine and it did not report any error. If I look into the details here, it just reports `Error: Process completed with exit code 4.` but I can't find any additional description of where the exact error is.

LukiLeu:
Ah I see `make lint` does change my file without reporting what might be wrong.

henryso:
Yeah, it's a little misleading because lint is typically a "check my code" thing rather that a "check and fix" thing.

apps/sbbtimetable/sbb_timetable.star
```diff
@@ -0,0 +1,203 @@
+load("render.star", "render")
+load("http.star", "http")
+load("encoding/base64.star", "base64")
+load("encoding/json.star", "json")
+load("cache.star", "cache")
+load("time.star", "time")
+load("schema.star", "schema")
+
+# Define some constants
+SBB_URL = "https://fahrplan.search.ch/api/stationboard.json?show_delays=1&transportation_types=train&stop="
```
betterengineering, line 10: Notes: 
- API Docs: https://fahrplan.search.ch/api/help
- Certificate is from R3 and will expire in April. We've had some expire like, two days after we land a new app and cause all sorts of problems 🤦 
- Looks like we'll be rate limited at  10080 departure/arrival tables
- Seems like it's fine to submit queries as long as we are under the req/s limit and I don't see anything that says we need attribution??

apps/sbbtimetable/sbb_timetable.star
```diff
@@ -0,0 +1,203 @@
...
+        version = "1",
+        fields = [
+            schema.Text(
+                id = "station",
+                name = "Station",
+                desc = "Station from which the timetable shall be shown.",
+                icon = "train",
+            ),
+            schema.Text(
+                id = "skiptime",
```
betterengineering, line 197: Fancy! Effectively, how far away do you live from the train??

apps/sbbtimetable/sbb_timetable.star
```diff
@@ -0,0 +1,203 @@
...
+        # Get the cached response
+        print("Hit! Displaying cached data.")
+        resp = json.decode(resp_cached)
+    else:
+        # Get a new reponse
+        print("Miss! Calling API.")
+        resp = http.get("%s%s" % (SBB_URL, station))
+        if resp.status_code != 200:
+            fail("Request failed with status %d", resp.status_code)
+        cache.set("sbb_%s" % station, resp.body(), ttl_seconds = 120)
```
betterengineering, line 51: Question: Any idea how many stations there are? We'll hit this API potentially `(24 hours * 60 minutes) / 2 minutes = 720` times the number of stations if this app is popular. At the `10080` limit we can only support 14 stations before we'd need an API key. Seems fine for now, but we'll want to keep an eye on it.

betterengineering (approved):
This looks good to me! I'm going to get this landed so we can start trying it out internally. If it still looks good, I'll launch it to the broader user base. Feel free to reply to my comments and I'll circle back tomorrow!

apps/sbbtimetable/sbb_timetable.star
LukiLeu: I can't find anything that says we need attribution: [https://fahrplan.search.ch/api/terms.en.html](https://fahrplan.search.ch/api/terms.en.html)

apps/sbbtimetable/sbb_timetable.star
LukiLeu: Actually, I have a train station right next to my home. But on weekdays I use a train station to commute that is a 30 minutes car drive away.

apps/sbbtimetable/sbb_timetable.star
LukiLeu: I think at the moment there are somewhat over 1000 stations (only train). If, later also bus and ship is added, this number will go up to ~29'000 stations.
It probably depends how many tidbyt users you have in Switzerland :-). But if you hit the rate limit, it should also be possible to increase the ttl value to five minutes.

PR #78, opened 2022-01-15
big-clock: Fix issue with API provider.

betterengineering:
This commit resolves an issue where the clock fails to render if the API
provider is down. With this change, we will now simply display the
nighttime color of the clock if the API provider is down.


betterengineering:
I suppose this doesn't fix the problem at all given the fail is not coming from within the app 😢 . I don't feel comfortable dropping SSL support given this has location data. Going to find another workaround.

PR #80, opened 2022-01-15
bigclock: Calculate sunrise/sunset in Starlark

rohansingh:
Port of implementation from [go-sunrise][1]. Uses pure math so there's
no need to call out to an external API.

[1]: https://github.com/nathan-osman/go-sunrise


rohansingh:
Results appear to be within a few seconds of the API. Variance is probably due to rounding.

PR #82, opened 2022-01-16
Add Pokedex app

mackorone:
This PR introduces the Pokedex app to the community repo! The app was created with the `make app` command and the code was copied from [`mackorone/tidbyt-pokedex`](https://github.com/mackorone/tidbyt-pokedex). The app gets data from [PokeAPI](https://pokeapi.co/), which it then displays as a static image.

pokedex

### DIFF
--- a/apps/pokedex/pokedex.star
+++ b/apps/pokedex/pokedex.star
@@ -10,17 +10,14 @@ load("render.star", "render")
 load("schema.star", "schema")
 load("time.star", "time")
 
-
 NUM_POKEMON = 386
 POKEAPI_URL = "https://pokeapi.co/api/v2/pokemon/{}"
-
 
 def get_schema():
     return schema.Schema(
         version = "1",
         fields = [],
     )
-
 
 def main():
     id_ = int(NUM_POKEMON * random()) + 1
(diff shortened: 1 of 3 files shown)


apps/pokedex/pokedex.star
```diff
@@ -0,0 +1,54 @@
...
+Author: Mack Ward
+"""
+
+load("http.star", "http")
+load("render.star", "render")
+load("schema.star", "schema")
+load("time.star", "time")
+
+NUM_POKEMON = 386
+POKEAPI_URL = "https://pokeapi.co/api/v2/pokemon/{}"
```
betterengineering, line 14: A few things we've learned to check, just wanted to document that we have checked them:
- The SSL cert is provided by cloudflare so we shouldn't have to worry about it expiring
- Looks like there are very limited rules and don't require attribution: https://pokeapi.co/docs/v2
- This API appears to be statically hosted by cloudflare, so I don't suspect it will go down very frequently 😄

apps/pokedex/pokedex.star
```diff
@@ -0,0 +1,54 @@
...
+                        render.Text("# " + str(id_)),
+                        render.Text(height),
+                        render.Text(weight),
+                    ],
+                ),
+            ],
+        ),
+    )
+
+def random():
```
betterengineering, line 52: @rohansingh we really need to add a `random` module 😅

apps/pokedex/pokedex.star
```diff
@@ -0,0 +1,54 @@
...
+def get_schema():
+    return schema.Schema(
+        version = "1",
+        fields = [],
+    )
+
+def main():
+    id_ = int(NUM_POKEMON * random()) + 1
+    pokemon_url = POKEAPI_URL.format(id_)
+    pokemon = http.get(pokemon_url).json()
```
betterengineering, line 25: According to their API, we should cache this data: https://pokeapi.co/docs/v2. Seems like we could cache for a few days on a given pokemon. We also should check the response for an error. Happy to land this in the short term and will follow up before launch.

betterengineering (approved):

apps/pokedex/pokedex.star
mackorone: @betterengineering, just FYI, I played around with caching for a bit but decided to skip it to avoid the extra complexity. Ideally, both the original request and the follow-up request (for the sprite) should be cached with the same TTL - they should use the same cache key - but I couldn't figure out how to cache a JSON dictionary containing all of the data. Serializing is easy, but I couldn't find a method for JSON deserialization.

henryso: You can create a new dict with both (base64-encode the sprite) and then use json.encode to encode it into a string that you can put it in the cache.  Then json.decode the cached value and base64-decode the sprite from it.

PR #84, opened 2022-01-16
Travel duration using google traffic information

LukiLeu:
This app shows the duration to get from an origin to a destination by using traffic information from google. To use this app, a google maps distance matrix api key needs to be provided.

google_traffic


henryso:
I think there is a problem with this PR.  It also pulls in your SBB app.  You should probably re-create this PR from a branch that comes off the branch that matches tydbit's main branch so that it doesn't pull in both apps in the one PR.

LukiLeu:
You are right, there went something wrong. I opened up a new PR #86.

PR #86, opened 2022-01-17
Travel duration using google traffic information

LukiLeu:
This app shows the duration to get from an origin to a destination by using traffic information from google. To use this app, a google maps distance matrix api key needs to be provided.

google_traffic


rohansingh:
We have an official Tidbyt API key that we could use for this instead of having to have everyone inject their own. Need a way to securely provide it though.

betterengineering:
@LukiLeu this is so rad! As @rohansingh mentions, we need to get secret support added before we can support this app. It will look something like:
```starlark
schema.Secret(
    id = "api_key",
    encrypted = "bm90IHNvIHNlY3JldAo="
)
```

But the value will come from a `pixlet encrpyt <TOKEN>` command and actually be encrypted. At runtime, we'll decode the encrypted value and place it in your config. We can add our token and commit it back to the repo encrypted so no one has to enter it. 

There are a few things that might be good to add. The one that comes to mind is to be able to set the travel mode: https://developers.google.com/maps/documentation/distance-matrix/overview#transit_mode

That way a cyclist can use this or if someone was looking to take the bus.

LukiLeu:
Looking forward to the `schema.Secret`. This will then be more user-friendly than when you need to add your own key.

In the meantime, I will add more settings in the next few days to make it more generic. (At the moment it has only the minimum settings as I needed them.)

LukiLeu:
It would actually be nice, if it was possible to show some fields in the app only if a certain condition is valid. That way it would be possible to add for example the transit mode parameter ([https://developers.google.com/maps/documentation/distance-matrix/overview#transit_mode](Transit Mode), which is then only shown if the type of transportation is public transportation. Also some checkboxes would be nice for this particular case.

rohansingh:
Just added the Tidbyt Inc's API key. If it gets too expensive we'll have to reconsider but works for now.

PR #87, opened 2022-01-17
Add ESPN news app

rs7q5:
This app displays the top three  "Top headlines" listed on ESPN or a particular sport that has this box (as far as I know I grabbed them all). It is currently set to cache every 4 hours. The default is to pull them from the front page of ESPN. 

espn_news


rs7q5:

rs7q5:
@rohansingh I was using this file as a template for some other stats and was wondering should `line 74` of the .star file be 

`text[i] = x+" "*(max_len-len(val))`?

If so what is the best way to go about changing it? My fork is both behind and ahead on commits, so I wasn't sure the best way to make this simple fix.

rs7q5:
This typo has been fixed as of 20220124

apps/espnnews/espn_news.star
```diff
@@ -0,0 +1,139 @@
...
+    title_cached = cache.get("title_rate1"), cache.get("title_rate2"), cache.get("title_rate3")
+
+    if all(title_cached) != False:  #if any are None then all(title_cached)==False
+        print("Hit! Displaying cached data.")
+        title = title_cached
+    else:
+        print("Miss! Calling ESPN data.")
+        rep = http.get(ESPN_API_URL)
+        if rep.status_code != 200:
+            fail("ESPN request failed with status %d", rep.status_code)
```
betterengineering, line 62: We may have to reconsider this `fail()`, but it's fine for now. We're trying to figure out better guidance here - but I think we may land on apps should never `fail()` simply because it will be unsustainable to ensure hundreds of APIs are online and working for all apps.

apps/espnnews/espn_news.star
```diff
@@ -0,0 +1,139 @@
...
+}
+
+def main(config):
+    sport = config.get("sport") or "All"
+    sport_txt, sport_ext = ESPN_SPORTS_LIST.get(sport)
+
+    #create full URL
+    ESPN_API_URL = ESPN_URL + sport_ext
+    if sport != "All":
+        ESPN_API_URL += "/?xhr=1"
```
betterengineering, line 46: If I understand correctly, this is used to populate their homepage - which means they may change it to better support their page without considering anyone is using it. Should be ok for now but we may have to revisit if this becomes a problem down the road.

betterengineering (approved):

PR #90, opened 2022-01-17
Added Golf Handicap App

ipv6freely:
This app allows a user to display their golf handicap information, including low and high handicaps as well as soft and hard caps. The user simply inputs their GHIN/WHS number and their last name. Data is pulled from the GHIN database
golf_handicap
.

### DIFF
--- a/apps/golfhandicap/golf_handicap.star
+++ b/apps/golfhandicap/golf_handicap.star
@@ -1,0 +1,105 @@
+"""
+Applet: Golf Handicap
+Summary: Displays your golf handicap
+Description: Displays your golf handicap using data from GHIN. Includes low/high and cap information.
+Author: Chris Jones (IPv6Freely)
+"""
+
+load("render.star", "render")
+load("encoding/base64.star", "base64")
+load("http.star", "http")
+load("schema.star", "schema")
+
+USGA_LOGO = base64.decode("""
+iVBORw0KGgoAAAANSUhEUgAAABsAAAAICAYAAAAIloRgAAAACXBIWXMAAAsTAAALEwEAmpwYAAAB
+xElEQVQokV3STUiUYRAH8N/upmbQh0luZqWCiAWp0AexdQg8GCVReavrWtA1sqRjt7x1CeKl7kEU
+RUTQB3RYEgQ1gjoUqyTGG0gtVOLHth3eZ2lpLjPPMM//P/+ZSU119M5KbLS/OHMfpjv7ZlHBHrzG
+gfD+hJHB1oMLuI7jaMUPvMO5uBB9zebyVzEWcHfEheg3rEN7SG7wz6q5eTTjFn7iyHhrTyOmsSkQ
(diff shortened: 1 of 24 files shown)


rohansingh:
If you run `make lint` in the root of the repository, it should fix the lint errors.

apps/golfhandicap/golf_handicap.star
```diff
@@ -0,0 +1,102 @@
...
+def main(config):
+    ghin_number = config.get("ghin_number", GHIN_NUMBER_DEFAULT)
+    last_name = config.get("last_name", LAST_NAME_DEFAULT)
+
+    tb8 = config.get("font", "tb-8")
+    tomthumb = config.get("font", "tom-thumb")
+    hcp_font = config.get("font", "6x13")
+
+    ghin_url = GHIN_URL.format(ghin_number, last_name)
+    data = http.get(ghin_url).json()
```
betterengineering, line 39: It looks like if I type in a number or last name incorrectly the API is pretty handy in telling us what's wrong. It might be worth adding something to tell the user? A quick and dirty example:
```starlark
    resp = http.get(ghin_url)
    if resp.status_code == 400:
        data = resp.json()["errors"]
        ghin_error = data.get("ghin_number")
        if ghin_error != None:
            return render.Root(
                child = render.Text(ghin_error)
            )
        # something similar for "last_name"
    elif resp.status_code != 200:
        return render.Root(
            child = render.Text("Error with the API")
        )
```

That way the user knows what they have entered incorrectly.

betterengineering (approved):
This looks great! My mother is a hobbyist golfer and will be very excited about this app 😄 . We should get the API status code checking in place but then it should be good to go.

apps/golfhandicap/golf_handicap.star
ipv6freely: @betterengineering sounds good to me. I wasn't sure exactly what the best way to handle errors in pixlet was since you're not doing logging to stdout or even print() statements. So you just output the issue directly to the Tidbyt display, rather than spitting out an error in the mobile app? Or does the way Tidbyt grabs info make that generally impossible?

betterengineering: @ipv6freely Great question - starlark doesn't have exception handling, so we have no way to recover if there is a `fail()` statement. One thing we've done in our internal apps is structure schema in a way that's impossible to have incorrect input. We still need to do work on our end to get the schema cleaned up so that all fields we support are available.

In the short term, the most straight forward way to do this is to display the issue back to the user using the display. Something like:
```starlark
def main(config):
    if issue:
        return display_failure("Issue with app")

def display_failure(msg):
    return render.Root(
        child = render.Marquee(
            width = 64,
            child = render.Text(msg),
        ),
    )

``

ipv6freely: Perfect. I'll get something made up and push a new commit. I also need to change the FontAwesome icon to a free one.

ipv6freely:
Added error handling, and changed the FontAwesome icon to a free icon.

Here's an example render from the error handling code:
golf_handicap

betterengineering (approved):
This looks great! Thanks again for putting this together!

PR #93, opened 2022-01-17
pokedex: Add caching and error handling.

betterengineering:
# Overview
The pokeapi requests that we cache so we don't incur high fees on their part: https://pokeapi.co/docs/v2. This change adds caching and error checking when calling to their API. We will now cache pokemon and their assets for 7 days.

# Changes
- pokedex: Add caching and error handling.
    - This commit adds caching and error handling around fetching pokemon. We will now cache a pokemon for 7 days to go easy on their API.

### DIFF
--- a/apps/pokedex/pokedex.star
+++ b/apps/pokedex/pokedex.star
@@ -5,13 +5,13 @@ Description: Display a random Pokemon along with its height and weight.
 Author: Mack Ward
 """
 
+load("cache.star", "cache")
+load("encoding/base64.star", "base64")
+load("encoding/json.star", "json")
 load("http.star", "http")
 load("render.star", "render")
 load("schema.star", "schema")
 load("time.star", "time")
-load("encoding/base64.star", "base64")
-load("cache.star", "cache")
-load("encoding/json.star", "json")
 
 NUM_POKEMON = 386
 POKEAPI_URL = "https://pokeapi.co/api/v2/pokemon/{}"
@@ -30,7 +30,7 @@ def main():
     height = str(pokemon["height"] / 10) + "m"
     weight = str(pokemon["weight"] / 10) + "kg"
     sprite_url = pokemon["sprites"]["versions"]["generation-vii"]["icons"]["front_default"]
-    sprite = get_sprite(sprite_url)
+    sprite = get_cachable_data(sprite_url)
     return render.Root(
         child = render.Stack(
             children = [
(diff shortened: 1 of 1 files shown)


apps/pokedex/pokedex.star
```diff
@@ -52,3 +55,34 @@ def main():
 def random():
     """Return a pseudo-random number in [0, 1)"""
     return time.now().nanosecond / (1000 * 1000 * 1000)
+
+def get_pokemon(id):
+    key = "pokemon_{}".format(id)
+    url = POKEAPI_URL.format(id)
+
+    data = cache.get(key)
```
mackorone, line 63: For my own understanding: why not just use the URL as the cache key?

apps/pokedex/pokedex.star
```diff
@@ -9,9 +9,13 @@ load("http.star", "http")
 load("render.star", "render")
 load("schema.star", "schema")
 load("time.star", "time")
+load("encoding/base64.star", "base64")
```
mackorone, line 12: Nit: should these be sorted?

apps/pokedex/pokedex.star
```diff
@@ -52,3 +55,34 @@ def main():
...
+    if data != None:
+        return base64.decode(data)
+
+    res = http.get(url = url)
+    if res.status_code != 200:
+        fail("request to %s failed with status code: %d - %s" % (url, res.status_code, res.body()))
+
+    cache.set(key, base64.encode(res.body()), ttl_seconds = CACHE_TTL_SECONDS)
+
+    return res.body()
```
mackorone, line 88: This method looks very similar to `get_pokemon`, sans a few minor changes. It'd be nice if this code existing as a library so each app didn't need to implement similar logic, especially logic that's somewhat error prone:
- Need to use the same `key` for `get` and `set`
- Need to return the same value passed to `set`
- Need to be mindful of the encoding scheme

mackorone (approved):
Thanks for doing this! Looks good, just a few inline comments/suggestions.

apps/pokedex/pokedex.star
betterengineering: We totally could! I started with the URL as the cache key and thought the URL might need to be encoded to avoid any weird character issues so I used the ID. I was going to do the same below in the `get_sprites` method but then realized if we use the ID as the key, we could have an issue where the sprite changed in the API for a given pokemon. So I encoded the URL as a base64 encoded string.

apps/pokedex/pokedex.star
betterengineering: If the linter passes, then it's good to go! But maybe the linter should require these to be sorted 😅 . Will fix.

apps/pokedex/pokedex.star
betterengineering: Totally. We need to support a starlark library in pixlet so we can have a common set of modules. a `get_cachable_url` method definitely feels useful. Something like:
```
get_cachable_data(url, ttl_seconds)
```

I can follow up on this PR to at least provide something that can be copy/pastad and dry out this code.

PR #94, opened 2022-01-17
Make lint more user-friendly

rohansingh:
* Add `format` as a synonym for `lint`.
* Don't require golangci-lint to be installed.
* Check for and install buildifier when running lint.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,10 @@ install-buildifier:
 	go install github.com/bazelbuild/buildtools/buildifier@latest
 
 lint:
-	golangci-lint run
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,9 +9,11 @@ install-buildifier:
 	go install github.com/bazelbuild/buildtools/buildifier@latest
 
 lint:
-	golangci-lint run
+	@ buildifier --version >/dev/null 2>&1 || $(MAKE) install-buildifier
```
betterengineering, line 12: Fancy!

betterengineering (approved):
❤️

PR #95, opened 2022-01-17
add noaa buoy app

tavdog:
A tidbyt app used for displaying current ocean swell based on a user specified noaa buoy id which can be procured from the noaa site : https://www.ndbc.noaa.gov/obs.shtml


tavdog:

apps/noaabuoy/noaa_buoy.star
```diff
@@ -0,0 +1,147 @@
...
+"""
+
+load("render.star", "render")
+load("schema.star", "schema")
+load("http.star", "http")
+load("encoding/json.star", "json")
+load("cache.star", "cache")
+
+def fetch_data(buoy_id):
+    url = "http://wildc.net/wind/noaa_buoy_api.pl?buoy_id=%s" % buoy_id
```
betterengineering, line 15: A few questions about this URL:
- Do you own this domain?
- How is this perl script being hosted?
- Can we make it HTTPS? It looks like there is a certificate for the domain

I think we can support this so long as it is your domain and it's not on a server in your closet 😄 . If this app gets popular, we may also want to figure out a way to get this app onto a more standardized API.

apps/noaabuoy/noaa_buoy.star
```diff
@@ -0,0 +1,147 @@
...
+    #buoy1_json = '{"height": "25.0", "period": "25", "direction": "SSE 161"}'   # test swell
+    #buoy1_json = '{"error" : "bad parse"}'   # test error
+    if buoy1_json == None:
+        buoy1_json = fetch_data(buoy1_id)
+        if buoy1_json != None:
+            cache.set(cache_key, str(buoy1_json), ttl_seconds = 3600)  # swell doesnt change very quickly, 1 hour should be good.
+    else:
+        buoy1_json = json.decode(buoy1_json)
+
+    print(str(buoy1_json))
```
betterengineering, line 47: Would you mind removing this print?
```suggestion
```

apps/noaabuoy/noaa_buoy.star
```diff
@@ -0,0 +1,147 @@
...
+
+def main(config):
+    # color based on swell size
+    color_small = "#00AAFF"  #blue
+    color_medium = "#AAEEDD"  #??
+    color_big = "#00FF00"  #green
+    color_huge = "#FF0000"  # red
+    swell_color = color_medium
+
+    buoy1_id = config.get("buoy1_id", 51211)
```
betterengineering, line 31: `buoy1_id` will come as a string - so you will need to type case it. I can also add support for `config.int()` but it doesn't exist today

apps/noaabuoy/noaa_buoy.star
```diff
@@ -0,0 +1,147 @@
...
+def get_schema():
+    unit_options = [
+        schema.Option(display = "feet", value = "feet"),
+        schema.Option(display = "meters", value = "meters"),
+    ]
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Text(
+                id = "buoy_1_id",
```
betterengineering, line 127: Is there any way to get a list of Buoys based on location? We have some additional schema types that may be useful. It will likely be pretty hard for a user to find the right ID.

apps/noaabuoy/noaa_buoy.star
```diff
@@ -0,0 +1,147 @@
...
+load("schema.star", "schema")
+load("http.star", "http")
+load("encoding/json.star", "json")
+load("cache.star", "cache")
+
+def fetch_data(buoy_id):
+    url = "http://wildc.net/wind/noaa_buoy_api.pl?buoy_id=%s" % buoy_id
+    resp = http.get(url)
+    if resp.status_code != 200:
+        fail("request failed with status %d", resp.status_code)
```
betterengineering, line 18: If I'm correct that this is self hosted, we should reply with a message to the user letting them know the server is down rather then failing the script

betterengineering (commented):
So sorry for the delay! I wasn't sure exactly what we could support in terms of hosted content which was causing me to drag my feet. This app is rad and I know my mom is asking for a tides app and this would be up her alley.

Can you also provide a screenshot using:
```
pixlet render apps/noaabuoy/noaa_buoy.star --gif --magnify 10
```

apps/noaabuoy/noaa_buoy.star
tavdog: I do own this domain and it's hosted on dreamhost.com.  I'm all for making it more standardized.  This was just a quick scrape job to get the data out of noaa's database.

apps/noaabuoy/noaa_buoy.star
tavdog: There is a search function on noaa's website but I haven't got around doing anything about it.  Since this app is actually about swell/wave data and not tides (maybe a name change should happen to make that clear), most users are surfers or other mariner type people who will be familiar with specific buoy ids and know how to get them.

tavdog:
Not sure where I should put the preview image.  Attaching it here
noaa_buoy
.

betterengineering:
Getting this merged! Thanks again @tavdog and so sorry for the delay

PR #97, opened 2022-01-19
nightscout: Add preview data.

betterengineering:
This commit refactors nightscout slightly to be able to support sample
data. This is important so we can render previews in the app store.


betterengineering:
cc @jer-tav

PR #98, opened 2022-01-21
Add MBTA transit app

marcusb:
mbta

### DIFF
--- a/apps/apps.go
+++ b/apps/apps.go
@@ -14,6 +14,7 @@ import (
 	"tidbyt.dev/community/apps/fuzzyclock"
 	"tidbyt.dev/community/apps/golfhandicap"
 	"tidbyt.dev/community/apps/manifest"
+	"tidbyt.dev/community/apps/mbta"
 	"tidbyt.dev/community/apps/nightscout"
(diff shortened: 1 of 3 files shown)


marcusb:

marcusb:
Hi! Couple of notes:

* The MBTA API has a quite restrictive rate limit unless using an API key, so this will not work very well for a shared installation. On the other hand I wouldn't want to put an API key in the source file. Is there a way around this? I could add a config property, but users would have to fetch their own API key which is annoying.
* There are about 8000 stops that populate a dropdown in the config. Will the UI scale to such long lists?
* I included the list of stops in the source file. One could fetch the list with an API call from `get_schema`, but I wasn't sure if HTTP requests in that context are a good idea?

drudge:
Hi @marcusb,

This looks awesome! I know that API keys and other secrets are something the Tidbyt team are working to support. I believe this comment from Mark would apply here as well:

https://github.com/tidbyt/community/pull/86#issuecomment-1014890060

rohansingh:
@marcusb Thanks a ton for the PR! We finally have a way to inject an API key securely. Going to merge this and inject our Tidbyt API key.

PR #99, opened 2022-01-21
Replace config.get with config.bool

rohansingh:
In the latest Pixlet, `config.get` always returns a string. We instead
need to use `config.bool` to get boolean values.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,3 +11,36 @@ require (
 	github.com/manifoldco/promptui v0.9.0
 	tidbyt.dev/pixlet v0.12.0
 )
+
+require (
+	github.com/PuerkitoBio/goquery v1.5.1 // indirect
+	github.com/andybalholm/cascadia v1.1.0 // indirect
+	github.com/antchfx/xmlquery v1.3.9 // indirect
+	github.com/antchfx/xpath v1.2.0 // indirect
+	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e // indirect
+	github.com/davecgh/go-spew v1.1.1 // indirect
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -9,38 +9,5 @@ require (
...
-	github.com/spf13/pflag v1.0.5 // indirect
-	github.com/zachomedia/go-bdf v0.0.0-20210522061406-1a147053be95 // indirect
-	go.starlark.net v0.0.0-20211203141949-70c0e40ae128 // indirect
-	golang.org/x/crypto v0.0.0-20210817164053-32db794688a5 // indirect
-	golang.org/x/image v0.0.0-20211028202545-6944b10bf410 // indirect
-	golang.org/x/net v0.0.0-20210825183410-e898025ed96a // indirect
-	golang.org/x/sys v0.0.0-20211205182925-97ca703d548d // indirect
-	golang.org/x/text v0.3.7 // indirect
-	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+	tidbyt.dev/pixlet v0.12.0
```
betterengineering, line 12: Huh. Why were all those direct deps there?

betterengineering (approved):
Short of looking through every toggle - this looks good! 😅

go.mod
rohansingh: I looked through every toggle. There's actually only 3 apps that use them.

rohansingh:
I looked through every toggle. There's actually only 3 apps that use them.

go.mod
rohansingh: I dunno why they disappeared, but that broke the build. I put them back.

PR #100, opened 2022-01-22
day_night_map: Used truecolor map image and added centering feature.

henryso:
I used a higher quality encoding of the map as the color reduction artifacts were bothering me.
I added the ability to center the map horizontally on the longitude of the location.


betterengineering (approved):
I love that I can center the map! LGTM

PR #102, opened 2022-01-22
ShowThis app - display information fetched from a customisable URL

janpi:
This app displays information it retrieves from a custom URL that can be defined in the app settings. It can fetch and display data from your web services, low-code platforms etc., without having to implement a custom Tidbyt app. For information on how to use this app, visit https://github.com/janpi/tidbyt-showthis

### DIFF
--- a/apps/showthis/showthis.star
+++ b/apps/showthis/showthis.star
@@ -51,7 +51,6 @@ def main(config):
 				display_vals = rep.json()
 
 				if display_vals != None:
-
 					cache_ttl_sec = CACHE_TTL_MINUTES * 60
 					cache.set("showthis_url " + url, json.encode(display_vals), ttl_seconds=cache_ttl_sec)
 
@@ -63,8 +62,6 @@ def main(config):
 					display_vals["icon"] = DEFAULT_ICON
 
 			else:
-				fail("Service request failed with status %d", rep.status_code)
-
 				display_vals["text_large"] = "Error"
 				display_vals["text_small"] = "code " + rep.status_code
 				display_vals["text_left"] = "ShowThis"
--- a/apps/showthis/showthis.star
+++ b/apps/showthis/showthis.star
@@ -17,21 +17,20 @@ DEFAULT_ICON = "iVBORw0KGgoAAAANSUhEUgAAAA0AAAANCAYAAABy6+R8AAAF4mlUWHRYTUw6Y29t
 
 def main(config):
 
+	url = config.get("url")
+	# url = "test" # query a test url
+
 	if cache.get("text_large") != None:
 		print("Hit! Displaying cached data.")
 
-		text_large = cache.get("text_large")
-		text_small = cache.get("text_small")
-		text_left = cache.get("text_left")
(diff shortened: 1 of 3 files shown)


janpi:

janpi:
recheck

janpi:
Hi guys, thanks for the cool Tidbyt! This app displays information it retrieves from a custom URL that can be defined in the app settings. It can fetch and display data from your web services, low-code platforms etc., without having to implement a custom Tidbyt app. For information on how to use this app, visit https://github.com/janpi/tidbyt-showthis.

henryso:
The cache is shared by all users of the app. The way you have it set up, the first user will get their data into the cache and the users that follow, regardless of url, until the entries time out, will get other people's data.

janpi:
Thanks! I guess I will use the configured URL then as key for the cache entries.

janpi:
@henryso I fixed the cache issue, pls confirm, thanks!

henryso:
Looks better. I guess the last suggestion from the peanut gallery is that you store a json encoded dictionary per URL. This way you won't run into a race condition right when the cache entries expire and you get some but not all values you need from the cache.

janpi:
Thanks for pointing this out @henryso - using dict in cache now. Pls confirm.

janpi:
BTW, I actually wanted to use a hash of the URL as cache key, rather than the actual URL, but I could not get any hash function to work. Is there something like hash.md5()? Thanks!

henryso:
Looks ok to me, not that this means anything it the long run. Good luck on the approval process.

rohansingh:
@janpi Thank you for this pull request! It looks really cool and I love the idea as well.

Apologies for the delay in reviewing. We're working on some security features that we need to have in place before accepting an app that regularly makes HTTP requests to user-configurable URL's. Once we have that in place, we can accept this app.

In the meantime, there is one code issue I noticed that needs to be fixed (see inline comment)

apps/showthis/showthis.star
```diff
@@ -0,0 +1,142 @@
...
+
+				else:
+					display_vals["text_large"] = "Error"
+					display_vals["text_small"] = "Invalid obj"
+					display_vals["text_left"] = "ShowThis"
+					display_vals["text_right"] = "Err"
+					display_vals["icon"] = DEFAULT_ICON
+
+			else:
+				fail("Service request failed with status %d", rep.status_code)
```
rohansingh, line 66: Let's drop this `fail`. It will kill the execution of the app immediately, so the error code will never be display to the user below.

janpi: Thanks @rohansingh! Statement was meant for testing purposes, I forgot to remove it.

janpi:
Thanks for the update @rohansingh, glad you like the idea! Looking forward to your security updates :)

betterengineering:
Hey @janpi! Thanks again for this change. We're still not in a great position to take this change at the moment and I don't want to leave you hanging indefinitely. I'm going to close this for now and we'll circle back in the future when we are better positioned to take arbitrary URLs from user input

PR #103, opened 2022-01-23
Update date_time_clock.star

AmillionAirs:
Hey gang, 

This shouldnt have been MONDAY in call caps otherwise now date takes it as a string rather than formatting. The clock just says its monday every of the weeks as is right now haha


betterengineering (approved):
oops! Thanks for the fix! Sorry we missed that in review. The time formatting here is based on Golang's, which can be a bit too clever: https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/time/format.go;l=53

PR #104, opened 2022-01-23
Add Phase of Moon app

alan-fleming:
Displays an image and a text description of the current phase of the moon. Accurate in both hemispheres.

phase_of_moon


alan-fleming:

henryso:
recheck

apps/phaseofmoon/phase_of_moon.star
```diff
@@ -0,0 +1,116 @@
...
+                    render.Image(src=base64.decode(phaseImage)),
+                    render.Padding(
+                        pad=(1,0,0,0),
+                        child =  render.WrappedText(moonPhase),
+                    )
+                ]
+            )
+        )
+    )
+
```
betterengineering, line 116: You will need this schema field to ensure we populate the `location` config object in `location = config.get("location")`:

```suggestion

def get_schema():
    return schema.Schema(
        version = "1",
        fields = [
            schema.Location(
                id = "location",
                name = "Location",
                desc = "Location for which to display the moon phase for.",
                icon = "place",
            ),
        ],
    )
```

betterengineering (approved):
Thanks for putting this together @alan-fleming, this looks great! I love that it's purely in starlark and takes the hemispheres into account ❤️ . You will need the schema added but otherwise this is good to go.

betterengineering:
You can run the following to fix the linter issues locally, then you can commit the changes:
```
make format
```

henryso:
I love this kind of app...

One thing, you don't need to project the time into any particular time zone to get `unix` out of it as UNIX time is based on the number of seconds from 1 Jan 1970 00:00:00 UTC, which is the same regardless of timezone.

apps/phaseofmoon/phase_of_moon.star
alan-fleming: How did I forget that? I even had the module added! On its way.

alan-fleming:
> I love this kind of app...
> 
> One thing, you don't need to project the time into any particular time zone to get `unix` out of it as UNIX time is based on the number of seconds from 1 Jan 1970 00:00:00 UTC, which is the same regardless of timezone.

Thanks - v new at this, so I'll amend here and that will come up in a future push

alan-fleming:
Does anyone have any suggestions as to why the build is failing? I'm outside my expertise here.

henryso:
@alan-fleming There's a blank line at then end of the star file.  Delete that line.

alan-fleming:
@henryso I'm absolutely certain the previous build failed because the linter complained there _wasn't_ a blank line at the end!
Oh well... learning is fun...

rs7q5:
I think when I ran make lint on my system before committing it adds the line. I noticed last night when I ran `make lint` it adds it, so it may not like two empty lines?

PR #106, opened 2022-01-26
Add goodservice app

blahblahblah-:
Powered by data from [goodservice.io](https://www.goodservice.io), displays real-time projections of departure times for NYC Subway.

goodservice


blahblahblah-:

rohansingh:
Hi @blahblahblah-. Thanks for opening this pull request!

This app seems to have a lot of overlap with the built-in NYC Subway app. I'm wondering if we should just improve that in some way rather than having two separate apps. What do you think?

blahblahblah-:
@rohansingh Yeah, I'm open to that idea. I don't know what source you were using for that, but I've been using my own API from goodservice, so the data may not be the same.

blahblahblah-:
Hey @rohansingh, just want to bump this back up to your attention and see what the next steps are.

rohansingh:
@blahblahblah- Let me get back to you on this tomorrow with info on how we can proceed.

rohansingh:
Hi @blahblahblah-. Thank you again for the contribution and the patience on this.

We discussed internally and here's the hold-up– we believe there's significant likelihood of confusion between this app and the built-in NYC Subway app. Because of how popular the subway feature is and the fact that we advertise it as a main use-case of the Tidbyt, we're afraid that we'd be on the hook for supporting and maintaining this app in a way that we aren't with most community apps.

What I'd like to do is hold off on merging this until we add a tagging feature to the Tidbyt phone app to clearly distinguish between first-party Tidbyt apps and community apps. Currently there is a bit of text that says "App X by <person>", but I don't think that's clear enough. Instead I'm working on adding something much clearer.

At that point I'd be comfortable merging this. Hope that makes sense and would love to hear your thoughts.

blahblahblah-:
That sounds good, thanks for getting back to me. Just wanted to reiterate that I'm also open to help make improvements of the built-in app.

betterengineering:
@blahblahblah- I think we're finally ready for this! We have a lot of work in flight to make our app store more sophisticated. In addition, we've seen very few folks confuse official apps for third party one so far.

betterengineering:
Going to land this and get it fixed up in trunk given this has been open for so long

PR #107, opened 2022-01-26
Update date_time_clock.star

AmillionAirs:
@Inxi teaching me a great coding lesson on why using the exact date and time during testing is a bad idea for error checking haha. I also added upper() since i like the look better with all upper case text.


betterengineering (approved):
Thanks @AmillionAirs! Also, thanks Henry for the assist!

PR #108, opened 2022-01-27
Added dummy data

ipv6freely:
Added dummy data so the preview is rendered in such a way that a user can see what the app will actually look like. Only happens if both Schema fields are empty. 
golf_handicap


ipv6freely:
recheck

betterengineering (approved):
Looks great! Thanks for adding the previews!

Issue #110, opened 2022-01-29
Is it possible to replicate the location behavior of the NYC Subway app?


samandmoore:
Hi there! 

I'm really excited about this new community apps approach. Thanks for doing this. 

## TL;DR
I'm wondering if it's possible to achieve the same thing that the existing NYC Subway departures app does with the location schema? I need a way to accept a user's location and then use that to generate a list of bus steps nearby so that the user can choose one for the app.

## Context
I've built a [NYC MTA Bus departures app](https://github.com/samandmoore/pixlet-mta-bus-time) and it's working great for me in my home setup today (currently I'm rendering it minutely in AWS Lambda 🤪).

I'd like to contribute it back to the community. I modeled it off of the built-in NYC MTA Subway departures app, but obviously since that's closed source, it's not quite the same. 

In the app today, I've just hard-coded it to use a stop that I care about. In order to make the stop dynamic, I'd need to be able to access the user's location and then query the MTA API to provide stops nearby. This appears to be exactly what the existing Subway departures app does, but I can't quite figure out how to replicate that. It seems like it's not possible.

samandmoore:
this was resolved with the addition of the `LocationBased` schema type. thanks!

PR #112, opened 2022-01-29
Add NYC Bus app

samandmoore:
This is based on the built-in NYC Subway app. 

I live somewhere in NYC where it's common to take the bus to get where I gotta go, so having this is really useful. 

nyc_bus

nyc_bus


apps/nycbus/nyc_bus.star
```diff
@@ -0,0 +1,159 @@
...
+
+load("cache.star", "cache")
+load("encoding/base64.star", "base64")
+load("encoding/json.star", "json")
+load("http.star", "http")
+load("render.star", "render")
+load("schema.star", "schema")
+load("time.star", "time")
+
+API_KEY = "<INSERT_API_KEY_HERE>"
```
samandmoore, line 16: I would like to use the secret/encrypt schema stuff once that's ready.

apps/nycbus/nyc_bus.star
```diff
@@ -0,0 +1,159 @@
...
+            schema.Dropdown(
+                id = "stop_code",
+                name = "Stop Code",
+                desc = "The stop code to use.",
+                icon = "bus",
+                default = stop_options[1].value,
+                options = stop_options,
+            ),
+        ],
+    )
```
samandmoore, line 50: I filed an issue #110 because I could use some help figuring out how to replicate the behavior of the NYC Subway app. At the moment this is just hardcoding 2 options and pointlessly asking for your location. I'd like to use your location to populate a list of stops to pick from. There's an MTA Bus Times API for that, I just don't know if it's possible to use the location to power the options in the schema if the user needs to provide the location?

apps/nycbus/nyc_bus.star
```diff
@@ -0,0 +1,159 @@
...
+
+def build_row(journey):
+    return render.Row(
+        expanded=True,
+        main_align="space_evenly",
+        children=[
+            render.Circle(
+                color="#%s" % journey["line_color"],
+                diameter=14,
+                child=render.Text(journey["line_name"], color="#fff", font="CG-pixel-3x5-mono"),
```
samandmoore, line 83: this is the smallest font I could find. It works great for bus lines with 2-3 characters (M1, Q69, etc) but it looks less awesome for 4 character bus lines (Q100).

samandmoore (commented):
Thanks for making it possible to contribute apps! 

I've just got one real question here about building a stop list picker based on the user's location.

samandmoore:

samandmoore:
I can fix up the linter violations this week

apps/nycbus/nyc_bus.star
betterengineering: Here is the secret docs: https://github.com/tidbyt/pixlet/blob/main/docs/modules.md#pixlet-module-secret

apps/nycbus/nyc_bus.star
betterengineering: You're looking for `LocationBased`: https://github.com/tidbyt/pixlet/blob/main/docs/schema/schema.md#locationbased

apps/nycbus/nyc_bus.star
betterengineering: It's the smallest font we can find as well 😅 . I'm not sure it gets any smaller - it's a pretty impressively small-yet-readable font.

betterengineering (commented):
Sorry for the delay @samandmoore! I figured it was best to get you everything you needed before review. Those doc links are all hot off the press. This app is rad and I know a lot of folks have been asking for bus support. Let me know if you need any additional help!

betterengineering:
Oh! And `make lint` will auto-format for you!

samandmoore:
Awesome! Happy to be able to contribute something useful. 

I will review your feedback and fix everything up at some point this week.

apps/nycbus/nyc_bus.star
samandmoore: @betterengineering hey! the secret docs say to use `app name` with `pixlet encrypt`. In this case, is that "NYC Bus" or should I use the app id "nyc-bus"? I'm just worried that since the app name is two words that might be an issue?

samandmoore: It seems like it does the right thing if is just quote it `pixlet "NYC Bus" secret-stuff-here`

samandmoore:
okay! i'm feeling good about this now. it's ready for another review.

i don't think it's possible to test the schema stuff for real, right? i reworked the local dev experience to rely on the dynamic get_stops schema stuff so we can be confident it works correctly in some sense. hopefully it looks good in the app UI 😅

samandmoore:
It looks like one of the tests is failing:

```
=== RUN   TestManifestsValidate
    apps_test.go:30: 
        	Error Trace:	apps_test.go:30
        	Error:      	Received unexpected error:
        	            	app descriptions should start with an uppercased character
        	Test:       	TestManifestsValidate
--- FAIL: TestManifestsValidate (0.00s)
```

I think it must be a funky issue with that validation because the description starts with an uppercase letter. It does have a hyphen in it, which is maybe the issue.

samandmoore:
ah yea, the issue is here https://github.com/tidbyt/community/blob/4d472aac620d0e1c39b45a273c83d416c4bb0504/apps/manifest/validate.go#L92

`strings.Title` must be returning something other than "Real-time"

i can just change the description if that's easier than fixing that validation to allow hyphenated words

apps/nycbus/nyc_bus.star
noahcolvin: @betterengineering can you clarify the question @samandmoore asked about the secret encryption?

The [doc](https://github.com/tidbyt/pixlet/blob/main/docs/authoring_apps.md#secrets) says to use `my-app-name` when the example is more like the lower kebab-case `ID` in the manifest and not the `Name`. `make app` doesn't even let you create an app name like that, it tells you to use title case.

`pixlet encrypt` will let me use anything there but since `decrypt` doesn't work locally I can't know if it's correct until it's in the app. It would be helpful if the documentation specified exactly what that should be if it maps directly to something in the manifest or if it's some other obvious value.

Thanks!

samandmoore: Got it. Let me re-encrypt the value. 

I used `make app` and didn't have any issues with "NYC Bus" as the app name. I chose that name based on the names of the built in apps. In fact, I also chose the description the same way. So, if the app name and description aren't valid then I guess the built-in apps aren't either? 

I'm totally down to change things if you'd prefer. I mostly just wanted to make this app feel like it fits right in with the "NYC Subway" and "NYC Ferry" apps y'all provide.

samandmoore: Okay! all done. 

I reencrypted the value and also just changed the description to remove the "-" so that the validation logic doesn't fail on the hyphenated starting word's conversion to title case.

samandmoore:
All fixed up! thanks for the feedback.

apps/nycbus/nyc_bus.star
noahcolvin: How do you know if it's right? You can type `pixlet encrypt "any-old Thing" secret` and it'll give you a value that you can paste into your app. If only the server can decrypt it and that third parameter is important in the process, then I feel the documentation is a little unclear. I have no way of knowing what other apps used for `pixlet encrypt` so I can't use them as an example.

I know things are new and tools are moving fast, and that's awesome; the last few pixlet releases added a ton of useful things. But I still have a good amount of uncertainty that what I do locally will work on the server and in the mobile app.

samandmoore: @noahcolvin i actually don't know if it's right 😅 

but after re-reading the docs, I'm thinking that the id makes more sense than the app name. I'm still hoping that @betterengineering can clarify for us

samandmoore: @noahcolvin it appears, from reading the source code, that the "app name" should be the [filename minus the ".star" suffix](https://github.com/tidbyt/pixlet/blob/eb3b928560c732708cb35de63d687dfaf702cd26/runtime/secret.go#L97) ([and here](https://github.com/tidbyt/pixlet/blob/eb3b928560c732708cb35de63d687dfaf702cd26/runtime/secret.go#L54)) but that seems like it wouldn't work properly if the filename is `nyc_bus.star` and the app id is `nyc-bus`. 

Sooo, i'm pretty sure i need to reencrypt this value using `nyc_bus`

noahcolvin: @samandmoore yeah I saw that last night too but wasn't 💯 sure if that was the case or not. I'd prefer @betterengineering (or anyone who knows) give a definitive answer and a line added to the docs or the example updated to be more clear.

It really killed my momentum last night when I set down to make something and the first thing I tried to do was encrypt an API key and the example didn't really make sense.

drudge:
I believe that pixlet encrypt should pass your app's ID, as defined in the go source file and should not include the `.star` extension.

For my PagerDuty app, with an `ID` of `pagerduty`, I encrypted the secret with:

```console
pixlet encrypt pagerduty my-client-secret
```

samandmoore:
Thanks @drudge! Although, that doesn't clarify things for me because in the case of the pagerduty app, it could be working incidentally because the id and the filename without .star are both the same value.

apps/nycbus/nyc_bus.star
```diff
@@ -0,0 +1,178 @@
...
+    data = res.json()["data"]["stops"]
+    stops = [
+        schema.Option(display = "%s - %s" % (stop["name"], stop["direction"]), value = stop["code"])
+        for stop in data
+    ]
+
+    return stops
+
+def main(config):
+    api_key = secret.decrypt(ENCRYPTED_API_KEY) or config.get("api_key")
```
samandmoore, line 70: okay, so now the tests are failing here. i think that's because the decryption doesn't work in CI?

betterengineering:
Hey @samandmoore! So sorry for the delay. I just got a chance to take a look and it appears the issue is a bit deeper. We need a way to render the app if there are no defaults passed. Right now, the app tries to make a request with a key that has a `None` value. But ideally we would simply serve up some sample data if there are no config values set. I'm going to merge this as is and follow up with a fix - I don't want to have you wait any longer or have to put in any more fixes

PR #113, opened 2022-01-30
Add Sunrise / Sunset times app

alan-fleming:
sunrise_sunset

Based on the sample code, but added graphics and support for 12 and 24 hour clocks.


betterengineering (approved):
This looks amazing! Love the art. There is a failing test for one of your other apps - but it's not related to this change. I'm going to land this and then follow up with fixing the test and formatting in main.

PR #114, opened 2022-01-30
Updates to the Sbb app

LukiLeu:
I added the following things to the SBB app:
- An error screen that is similar as in the golf app. The error screen is shown if there was an error with the API as well as there was an incorrect station entered.
- It is now checked if the color of a train exists in the dictionary, prior to using the color from the dictionary. This is done, in case there are still not all train types in that dictionary.
- The station name is now encoded. This should now also allow entering Stations with special chars like space or ä/ö/ü (which we have a lot).
- Already added the new typeahead schema field which is currently in a PR. Will check back if everything is correct, once the documentation to that field is provided.


rohansingh:
We need to merge the latest Pixlet changes into this repo for `Typeahead` to work.

PR #115, opened 2022-01-31
Add National today app

rs7q5:
This app displays today's holidays from National Today (from their What is today page). 

nationalToday


apps/nationaltoday/nationaltoday.star
```diff
@@ -0,0 +1,130 @@
...
+                date_split[0] = date_split[0][:4]
+            else:
+                date_split[0] = date_split[0][:3]
+
+            holiday_txt.insert(0, " ".join(date_split))
+        else:
+            holiday_txt = ["Error", "Could not get holidays!!!!"]
+
+        #cache the data
+        cache.set("holiday_rate", json.encode(holiday_txt), ttl_seconds = 86400)  #grabs it once a day
```
rohansingh, line 68: Please only cache for 30 minutes instead of an entire day. Otherwise we could, for example, fetch the page at 11 PM on Monday and keep serving Monday's response through 11 PM on Tuesday.
```suggestion
        cache.set("holiday_rate", json.encode(holiday_txt), ttl_seconds = 1800)  # cache for 30 min
```

PR #117, opened 2022-01-31
Add Jokes JokeAPI app (Attempt #2)

rs7q5:
This app pulls a joke from [JokeAPI](https://jokeapi.dev/). The default is safe mode, to TRY and avoide any questionable jokes. 

Pulls twice a day, but this could be switched to once a day. 

jokeAPI


rohansingh:
Hey @rs7q5, thanks for the PR, and for including the safe mode.

To make sure we abide by [the community standards][1] and don't make anyone feel uncomfortable, I'd really like to err on the side of being safe and inclusive here. Would you consider just removing all the toggles and having safe mode always enabled? Having safe mode locked in would make this a lot quicker and easier to accept.

[1]: https://github.com/tidbyt/community/blob/main/docs/CODE_OF_CONDUCT.md#our-standards

rs7q5:
Not a problem, I probably should have just only used safe mode anyways. It is set when people set their jokes, so I'm sure there are some that squeak by, but there's not much we can do about that.

rohansingh:
Thanks!!

PR #123, opened 2022-02-02
Don't cache error responses

rohansingh:
If servers respond with errors, don't cache them. Otherwise we never try
to fetch again until the cache expires, which for these apps takes quite
a while.


rohansingh:
(/cc @rs7q5)

PR #125, opened 2022-02-02
Steam App for Tidbyt

jer-tav:
This app displays what a user is currently playing on Steam, along with the art for the game.  If the user is currently not playing a game, the list of games from the previous two weeks of activity is displayed.

steam
steam

The users' profile must be public in order to display the game activity.

### DIFF
--- a/apps/steam/steam.star
+++ b/apps/steam/steam.star
@@ -17,7 +17,7 @@ API_RECENTLY_PLAYED_GAMES = API_BASE_URL + "IPlayerService/GetRecentlyPlayedGame
 API_OWNED_GAMES = API_BASE_URL + "IPlayerService/GetOwnedGames/v0001"
 
 def main(config):
-    steam_id = config.get("steam_id")
+    steam_id = config.get("steam_id", None)
     api_key = secret.decrypt("""
         AV6+xWcE+DYxl3gCCIpeRaoTWtnpTqIEqS8zWqnvoStfp/IrLru54NPMHgekh8jwUshSLUFp4Y9GFy7C/
         zLyBm1DzJODRDfTo3NROgQt7jJI6nzK/8I6MPMv33VZ4/ZjcSH3BkQ0THHB0XcGIv++ntIj3/tcDsvQtr
@@ -115,6 +115,19 @@ def main(config):
                 ),
             ],
         ),
+    )
+
+def get_schema():
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Text(
+                id = "steam_id",
+                name = "Steam User ID",
+                desc = "Your Steam User ID (i.e. 123456789 in https://steamcommunity.com/profiles/123456789)",
+                icon = "user",
+            ),
+        ],
     )
 
 def display_failure(msg):
--- a/apps/steam/steam.star
+++ b/apps/steam/steam.star
@@ -55,13 +55,12 @@ def main(config):
         if resp.status_code != 200:
             return display_failure("Failed to get the current game icon with code {}".format(resp.status_code))
 
-        if resp.json()["response"]["game_count"] != 1:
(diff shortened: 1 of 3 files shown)


apps/steam/steam.star
```diff
@@ -0,0 +1,160 @@
...
+API_PLAYER_SUMMARIES = API_BASE_URL + "ISteamUser/GetPlayerSummaries/v0002"
+API_RECENTLY_PLAYED_GAMES = API_BASE_URL + "IPlayerService/GetRecentlyPlayedGames/v0001"
+API_OWNED_GAMES = API_BASE_URL + "IPlayerService/GetOwnedGames/v0001"
+
+def main(config):
+    steam_id = config.get("steam_id", None)
+    api_key = secret.decrypt("""
+        AV6+xWcE+DYxl3gCCIpeRaoTWtnpTqIEqS8zWqnvoStfp/IrLru54NPMHgekh8jwUshSLUFp4Y9GFy7C/
+        zLyBm1DzJODRDfTo3NROgQt7jJI6nzK/8I6MPMv33VZ4/ZjcSH3BkQ0THHB0XcGIv++ntIj3/tcDsvQtr
+        NkoszNXck6JEZaDpo=
```
rohansingh, line 25: Let's use this API key owned by tidbyt.com. That way we can contact them for higher rate limits and also OAuth support in the future.

```suggestion
        AV6+xWcE9acn7G/Amrp+rzPaSppKSmwWdVBJklcT6q63X9wG8UFn0ma1U9oCtmwjzVyZFF1JzzpLre2K
        mVuRNN+Lk7UICOvuChgDC6pstPv9ecAHFGu9h9C7kc3ntBcbUARJIVWXZVH2ji+PtIis/jekWarIgISt
        qukjL49KF/GT3BtMwJ4=
```

apps/steam/steam.star
```diff
@@ -0,0 +1,160 @@
...
+
+        resp = http.get(API_OWNED_GAMES, params = {"key": api_key, "input_json": str(json_blob)})
+
+        if resp.status_code != 200:
+            return display_failure("Failed to get the current game icon with code {}".format(resp.status_code))
+
+        if resp.json()["response"]["game_count"] == 1:
+            game_icon_hash = resp.json()["response"]["games"][0]["img_icon_url"]
+            game_icon_url = "http://media.steampowered.com/steamcommunity/public/images/apps/" + str(current_game_id) + "/" + game_icon_hash + ".jpg"
+            main_icon = http.get(game_icon_url).body()
```
rohansingh, line 61: Can you please cache this icon? Here's the documentation for the cache module:
https://github.com/tidbyt/pixlet/blob/main/docs/modules.md#pixlet-module-cache

apps/steam/steam.star
jer-tav: Absolutely - much better than using mine!

apps/steam/steam.star
jer-tav: I would love to, but I'm slightly at a loss for how to do this with the jpeg it downloads.

I have this for the key: `key = steam_id + "_" + current_game_id` - so it should be unique for the player and the current game.

I'm doing this for pulling the icon and trying to put it in the cache:
`main_icon = http.get(game_icon_url).body()
print("Setting Cache")
cache.set(key, base64.encode(main_icon), ttl_seconds = CACHE_TTL_SECONDS)`

But when I look at the `main_icon`, it looks to be a binary object.

henryso: `base64.encode(main_icon)` is encoding the binary `main_icon` into a string for the cache.  You'll need to `base64.decode(cache.get(key))` to decode it when you get it out of the cache.

jer-tav: Never mind - this was an issue with using `--watch` with `pixlet serve`.  Code works as expected.

rohansingh (approved):

PR #130, opened 2022-02-04
clockbyhenry: Remove print statement

rohansingh:
This is just printing a lot to Tidbyt server logs.


matslina (approved):

henryso:
Sorry about that.  I should have removed it.

rohansingh:
No worries. Long term, we need to figure out some better way to handle `print()` from apps than just spitting it to server logs.

PR #131, opened 2022-02-04
binaryclock: Fix defaults

rohansingh:
The default values in schema need to be the actual option values and not
the option text.


rohansingh:
(/cc @LukiLeu )

PR #136, opened 2022-02-04
[Date Progress] Hide progress labels

drudge:
While testing the configuration, I noticed that the "Show percentages" option was not hiding the percentage labels.

Before:
date_progress

After:
date_progress


drudge:
Tagging @possan for visibility.

betterengineering (approved):
Thanks @drudge

PR #138, opened 2022-02-05
Add Analog Time app (Attempt two)

rs7q5:
Apparently my app name can't have a number in it...

Anyways because we needed more clocks here's another analog clock created using `render.Plot`

There is a toggle to display the date in the top left corner (default is to disable)

analog_clock

### DIFF
--- a/apps/analogtime/analog_time.star
+++ b/apps/analogtime/analog_time.star
@@ -14,7 +14,6 @@ load("schema.star", "schema")
 load("math.star", "math")
 load("time.star", "time")
 
-
 def main(config):
     timezone = config.get("location")
     if timezone != None:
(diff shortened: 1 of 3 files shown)


apps/analogtime/analog_time.star
```diff
@@ -0,0 +1,121 @@
...
+#analog_time.star
+#Created 20220204 RIS
+#Last Modified 20220204 RIS
+
+load("render.star", "render")
+load("schema.star", "schema")
+load("math.star", "math")
+load("time.star", "time")
+
+base_URL = "https://nationaltoday.com/what-is-today/"
```
rohansingh, line 17: Looks good to me other than I don't think this should be in here

```suggestion
```

rs7q5: Correct, I guess I forgot to remove it. not sure why that broke lint though.

betterengineering (approved):
LGTM!

PR #140, opened 2022-02-05
Adding Near Earth Objects App

noahcolvin:
This app hits the Near Earth Object API [NeoWs](https://www.neowsapp.com/) to display the name of the next nearest object to Earth as well as the relative time, distance, and speed of when it will miss.

Also allows the user to configure kilometers vs miles.

neo_byt

### DIFF
--- a/apps/apps.go
+++ b/apps/apps.go
@@ -23,6 +23,7 @@ import (
 	"tidbyt.dev/community/apps/manifest"
 	"tidbyt.dev/community/apps/mbta"
 	"tidbyt.dev/community/apps/nationaltoday"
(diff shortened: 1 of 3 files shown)


noahcolvin:

noahcolvin:
recheck

rohansingh:
@noahcolvin Looks great, thanks for the PR!

rohansingh:
Hi @noahcolvin. Unfortunately I had to revert this because it was failing tests.

The flavor of Starlark we use on our server does not allow `while` loops. Is it possible to rewrite `find_soonest_starting` to use a `for` loop that ranges over a fixed sequence?

PR #146, opened 2022-02-08
Steam

jer-tav:
Added better description for the app for getting 17 digit Steam ID, and demo data for preview.


jer-tav:
Thanks Rohan, I totally messed that up 🤦‍♂️

PR #148, opened 2022-02-10
Adding Near Earth Objects App

noahcolvin:
New PR from merged/reverted https://github.com/tidbyt/community/pull/140. Changed `while` to `for` loop and also increased cache time (these objects are usually known about for a while).

This app hits the Near Earth Object API [NeoWs](https://www.neowsapp.com/) to display the name of the next nearest object to Earth as well as the relative time, distance, and speed of when it will miss.

Also allows the user to configure kilometers vs miles.

neo_byt

### DIFF
--- a/apps/apps.go
+++ b/apps/apps.go
@@ -23,6 +23,7 @@ import (
 	"tidbyt.dev/community/apps/manifest"
 	"tidbyt.dev/community/apps/mbta"
 	"tidbyt.dev/community/apps/nationaltoday"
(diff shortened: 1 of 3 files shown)


noahcolvin:
I'll get the linter setup locally in the next few days and fix that. You folks are making changes quickly 😃

rohansingh:
I think I should be able to run the formatter on a PR. Let me try.

betterengineering (approved):
ty!!

PR #156, opened 2022-02-15
Fixed cache issues in ESPN news app

rs7q5:
Fix isn't pretty, but it works. 

Preview should now work correctly in Tidbyt App and each sport is cached separately as well.


betterengineering (approved):
Thanks for the quick fix!

PR #160, opened 2022-02-18
Fishbyt: increased Marquee offset and adjusted delay based on fact length

vlauffer:
Fixed a couple of readability issues for my fishys!


betterengineering (approved):
LGTM

PR #163, opened 2022-02-19
fixed stupid typo between schema and config causing it to not work.

tavdog:


betterengineering (approved):
Oops - totally missed this @tavdog! My appologies

PR #166, opened 2022-02-19
Google Traffic App: Updates

LukiLeu:
Added the following changes:
- An alias for the destination and the origin can be provided. In that case, the alias is shown instead of the address
- The Distance is shown in a fourth row instead of the same row as the time
- The units of the time are shown in their short representation (eg. m instead of min)
- The Distance can now be shown in miles and kilometers


betterengineering (approved):
LGTM! I'd normally be weary when changing the defaults for schema fields - but your default for showing distance feels harmless

PR #171, opened 2022-02-23
ESPN News: Added two feature requests

rs7q5:
Added Features:

- There is a speed option to slow down the scroll text (slowing down may not allow text to finish, but there's nothing that can be done with that at the moment)
- There is an option to scroll the text vertically now as well.

### DIFF
--- a/apps/espnnews/espn_news.star
+++ b/apps/espnnews/espn_news.star
@@ -6,10 +6,11 @@ Author: rs7q5 (RIS)
 """
 #espn_news.star
 #Created 20211231 RIS
-#Last Modified 20220215 RIS
+#Last Modified 20220223 RIS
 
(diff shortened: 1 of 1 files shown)


apps/espnnews/espn_news.star
```diff
@@ -77,8 +78,47 @@ def main(config):
...
+                cross_align = "start",
+                expanded = True,
+                children = title_format,
+            ),
+            offset_start = 0,
+            offset_end = 0,
+        )
     return render.Root(
-        delay = 30,  #speed up scroll text
+        delay = int(config.str("speed", "30")),  #speed up scroll text
```
betterengineering, line 121: I suppose we need a `config.int()` 😅

betterengineering (approved):

PR #172, opened 2022-02-23
pagerduty: Show on-call information when not on-call

drudge:
This PR tweaks the overall design of the PagerDuty app.

### Changes

* Adds a new view of the status bar that scrolls through the on-call status by team if the user is off-duty. 
* Changes the way I build the PagerDuty logo, fixes the margins, and allows for hiding it completely.
* Renamed triggered alert count label from "TRIG" to "NEW"
* Adds the ability to hide the on-call bar when you are not on-call

### Before this PR:
pagerduty

### After this PR:
pagerduty

### Notes

* I'm using the `humanize` module in Pixlet so this may require a new version to be published before it can be deployed
* I've included a devcontainer that allows developing Community apps entirely from Github Codespaces.


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.222.0/containers/go/.devcontainer/base.Dockerfile
```
betterengineering, line 1: This is rad! Can we add it to a separate PR?

apps/pagerduty/pagerduty.star
```diff
@@ -6,8 +6,8 @@ Author: drudge
 """
 
 load("cache.star", "cache")
-load("encoding/base64.star", "base64")
 load("encoding/json.star", "json")
+load("humanize.star", "humanize")
```
betterengineering, line 10: Getting a release of pixlet cut so your tests pass! One sec

.devcontainer/Dockerfile
drudge: Sure thing, I've reverted my merge and I'll open a separate PR.

drudge: https://github.com/tidbyt/community/pull/174

betterengineering:
Going to land this, it should pass on the main branch

PR #174, opened 2022-02-23
Add devcontainer configuration for Github Codespaces

drudge:
This PR adds a [devcontainer](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/configuring-codespaces-for-your-project) configuration for [Github Codespaces](https://github.com/features/codespaces) to the Community repo.

It configures a Go dev environment, clones and builds the latest version of Pixlet and makes it available in the user's `$PATH`. The Bazel and Go VSCode extensions are installed automatically as well.

This allows anyone to click on Codespaces in the community repo and be creating apps right in the browser with no set up required. When they go to commit it tells them they don’t have permissions to push to the repo and asks to make a fork. If the user chooses to fork, Github makes a new fork and then offers to create a PR automatically.

Running `pixlet serve` in a terminal opens a port which can be opened in the embedded browser for awesome split-screen development:
Screen-Shot-2022-02-19-19-37-57 93

Note: I've noticed that public sharing doesn’t work because Github serves over HTTP, so we’d have to use wss for the web socket connection.


betterengineering (approved):
This is super rad! Can't wait to give this a whirl.

PR #178, opened 2022-02-24
nycbus: Fix issue with previews.

betterengineering:
This commit adds preview data so that we can render the app without
having to make an authenticated request.


betterengineering:
cc @samandmoore

samandmoore:
Awesome! Thanks for doing this.

PR #181, opened 2022-02-24
Add GA Pilot Buddy app

icdevin:
This app uses a user-chosen location and a third-party API for nearby aerodromes and their current flight rules and METAR (weather) information. For General Aviation pilots, having this information available at a glance is invaluable as it's the first step in making a go-no go flight decision. I did not include full METAR information as it would probably be a bit too long-form for horizontal marqueeing but the chosen fields are the most relevant.

The app only has one other option other than location, which is "Show All Aerodromes". By default, non-public aerodromes are filtered out; however, if this is enabled, non-public ones such as military or private will also be shown.

Just FYI, flight rules color coding goes like this:
Color-coding-chart

So in this example, all 3 local airports are VFR (just how I like it!)

ga_pilot_buddy

Here's somewhere on the edge of some worse weather:
ga_pilot_buddy

### DIFF
--- a/apps/gapilotbuddy/ga_pilot_buddy.star
+++ b/apps/gapilotbuddy/ga_pilot_buddy.star
@@ -45,7 +45,7 @@ FLIGHT_RULES_COLOR_MAP = {
 
 def get_avwx_headers(config):
     return {
-        "Authorization": "Token {}".format(secret.decrypt(AVWX_TOKEN) or config.get("avwx_token"))
+        "Authorization": "Token {}".format(secret.decrypt(AVWX_TOKEN) or config.get("avwx_token")),
     }
 
 def get_nearby_aerodromes(location, config):
@@ -61,8 +61,8 @@ def get_nearby_aerodromes(location, config):
 
         # Although we only show three, this grabs extras in case some get filtered
         # out such as military or private aerodromes
-        params = { "n": "10" }
-        resp = http.get(url, params=params, headers=get_avwx_headers(config))
+        params = {"n": "10"}
+        resp = http.get(url, params = params, headers = get_avwx_headers(config))
         if resp.status_code != 200:
             print(resp)
             return None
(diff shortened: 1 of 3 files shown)


icdevin:

SomethingNew71:
Very cool!

apps/gapilotbuddy/ga_pilot_buddy.star
```diff
@@ -0,0 +1,201 @@
...
+    "LIFR": "#D300D3",
+}
+
+def get_avwx_headers(config):
+    return {
+        "Authorization": "Token {}".format(secret.decrypt(AVWX_TOKEN) or config.get("avwx_token")),
+    }
+
+def get_nearby_aerodromes(location, config):
+    # Truncates the precise lat/lng for data privacy since this will be sent out
```
betterengineering, line 52: ❤️

apps/gapilotbuddy/ga_pilot_buddy.star
```diff
@@ -0,0 +1,201 @@
...
+    # to a third party API
+    lat = humanize.float("#.#", float(location["lat"]))
+    lng = humanize.float("#.#", float(location["lng"]))
+    str_geo = "{},{}".format(lat, lng)
+
+    aerodromes = cache.get(str_geo)
+    if aerodromes == None:
+        url = "https://avwx.rest/api/station/near/{}".format(str_geo)
```
betterengineering, line 60: Great work making sure your calls are limited! It looks like you'll get 4K calls per day assuming you're on the free plan. I see you're also aggressively caching and truncating the lat/lng. One decimal place is [11.1km/6.9mi](https://gis.stackexchange.com/questions/8650/measuring-accuracy-of-latitude-and-longitude) so there will certainly be only a few calls per week for a major metropolitan area like New York or Los Angeles. If we ever get so many users where this app is doing 4k calls per day in the current form we'd also be in a position to just pay for it 😄

betterengineering (approved):

apps/gapilotbuddy/ga_pilot_buddy.star
icdevin: Thanks! It would've ended up being pretty noisy otherwise for sure. One digit is better for caching but often there's many airports in a tight space so it may need to be increased later if it starts looking wrong. In any case, luckily airports don't open or close often at all so it can be heavily cached. 😎 

That's very generous of you. For now I've got overages turned on until it goes over the limit, but I'll let you know if things get crazy. At some point it'll probably be cheaper to pay for the tier anyway because their first paid tier has a "Multiple Reports" call so I can group three requests into one. Food for thought lol.

PR #182, opened 2022-02-24
Add Vertical Message App

rs7q5:
This was a quick feature request on the community and I think this actually belongs more so as a feature addition to the messages app, but since I don't have the original source code, I created a simple app to display a long vertical message.

The user can also select a few colors and change the scroll speed.
Feel free to reject the PR or use the code to modify the existing messages app though.

vertical_scroll


betterengineering (approved):
Looks great!

Issue #41, opened 2022-01-10
Global color personalization variable


joeyhoer:
To provide some level of personalization, a global configuration setting for "preferred color" could be used by app developers to customize the display of their apps. (e.g. setting the display to "yellow" would tint all apps supporting the global variable yellow).

Along with this, color helper functions (e.g. invert, hue-rotate, saturation-rotate, lightness-rotate; see [Sass color functions as possible reference](https://sass-lang.com/documentation/modules/color)) would also be useful in making the app's display colors relative to a single variable.

PR #194, opened 2022-02-28
Updating Date Progress app to use color functions and added color schema

joeyhoer:
@possan I've updated your app to utilize color functions that can dynamically adjust the color rather than hard coding the colors (as I suggested in https://github.com/tidbyt/community/issues/41, it would be sweet if there was a global color library that could be used for this purpose).

Anyway, this PR gives users the ability to choose colors for each of the progress bars for a bit more personalization. I haven't been able to test it with the schema, but I'm _pretty_ sure I got it all right.

### DIFF
--- a/apps/dateprogress/date_progress.star
+++ b/apps/dateprogress/date_progress.star
@@ -20,26 +20,26 @@ P_SHOW_VALUES = "show_values"
 P_SHOW_DAY = "show_day"
 P_SHOW_MONTH = "show_month"
 P_SHOW_YEAR = "show_year"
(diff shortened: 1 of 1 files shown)


rohansingh:
Hi @joeyhoer. Thanks for the contribution!

Looks like there are two things that need to be fixed to pass the tests:
- [ ] Run `make format` to fix formatting.
- [ ] Fix the bug here: https://github.com/tidbyt/community/runs/5353499434?check_suite_focus=true#step:5:56

joeyhoer:
@betterengineering @rohansingh Just a heads up, this has been updated!

PR #197, opened 2022-02-28
NOAA Buoy Huge update.  displays most data,  user can select which data to show

tavdog:
fetch buoy data directly from NOAA via rss link
displays swell, wind, temperatures, tide, vis . . . 
user selects which data to display in the app
Screen Shot 2022-02-27 at 6 34 10 PM


apps/noaabuoy/noaa_buoy.star
```diff
@@ -1,7 +1,7 @@
 """
 Applet: NOAA Buoy
-Summary: Display buoy wave data
-Description: Display swell data for user specified buoy. Find buoy_id's here : https://www.ndbc.noaa.gov/obs.shtml Buoy must have height,period,direction to display correctly
+Summary: Display buoy weather data
```
betterengineering, line 3: One thing to note - the Summary and Description that get displayed in the app are actually in the `.go` file in this directory. We need a better way to sync the two or eliminate one of them to make this more clear, our bad!

betterengineering (approved):

apps/noaabuoy/noaa_buoy.star
tavdog: Yes I noticed that this morning too.  I'll make sure to update the go file too. Thanks.

PR #210, opened 2022-03-02
Creating a Warframe Cycles app

grantmatheny:
This very simple app displays the time remaining in the current cycles for the open world areas in the free-to-play game [Warframe](https://www.warframe.com/). I use my Tidbyt next to my computer, and figuring out which time in which cycle at a glance is really nice. Of course, [Warframe Hub](https://hub.warframestat.us/#/) does a nice job of this, but if I don't want to dedicate a whole browser tab to it, Tidbyt is great. I'm still relatively new to Python, and very new to Starlark, having just gotten my Tidbyt last week, so if the code looks amateurish, it is! Any suggestions are welcome.

warframe_cycles

### DIFF
--- a/apps/apps.go
+++ b/apps/apps.go
@@ -7,7 +7,6 @@ package apps
 import (
 	"errors"
 
-	"tidbyt.dev/community/apps/MNLightRail"
 	"tidbyt.dev/community/apps/analogclock"
 	"tidbyt.dev/community/apps/analogtime"
 	"tidbyt.dev/community/apps/arcadeclassics"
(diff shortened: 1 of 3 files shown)


grantmatheny:

grantmatheny:
Re-linted this, and while I can't see what changed (maybe the encoding on a particular character?) git definitely seems to think something was fixed.

grantmatheny:
It looks like the linter failed due to a non-linting problem with its executable, unless I'm missing something. Can this be re-run?

rohansingh:
@grantmatheny You're totally correct. I'll rebase this on `main` where the problem is fixed and the test should pass.

PR #211, opened 2022-03-02
Add option to pokedex to show imperial units.

mtimkovich:
Also adds a space between number and units.

ptal @mackorone


betterengineering:
Going to merge this - if @mackorone is strongly opposed we can revert it 😄

PR #213, opened 2022-03-02
Re-encrypt secrets based on app package name

rohansingh:
We're standardizing on encrypting secrets using the app package name
rather than the filename. So whereas before you would do this:

```
pixlet encrypt nyc_bus ...
```

Now we do this:

```
pixlet encrypt nycbus ...
```

Hopefully this should cause less confusion down the road.


rohansingh:
Tested against tidbyt/tidbyt#3731 and it passes.

PR #228, opened 2022-03-04
google-traffic: Clean up google traffic.

betterengineering:
This is such a great app and it sucks that we had to remove it. This
commit cleans up this app so it no longer shows up in our tests or in
our repo.


betterengineering:
cc @LukiLeu

PR #230, opened 2022-03-04
update tempest weather to use sea level pressure

codylerum:
This matches what the tempest website uses to display pressure.

See: https://discuss.tidbyt.com/t/tempest-weather-showing-wrong-pressure-value/1360


codylerum:

PR #235, opened 2022-03-06
Add PowerBall

AmillionAirs:
Tidbyt team welcome to another adventure of Amillion Air tries to learn how to do coding things. I have bought a second tidbyt for my parents and i know they love to play the Powerball so i figured id make this simple app for them to keep them up to date!

App shows

Current Prize Pool
Last numbers drawn
Last draw date and if there were any winners
next draw date

image

### DIFF
--- a/apps/powerball/powerball.go
+++ b/apps/powerball/powerball.go
@@ -1,0 +1,25 @@
+// Package powerball provides details for the PowerBall applet.
+package powerball
+
+import (
+	_ "embed"
+
+	"tidbyt.dev/community/apps/manifest"
+)
+
+//go:embed powerball.star
+var source []byte
+
+// New creates a new instance of the PowerBall applet.
+func New() manifest.Manifest {
+	return manifest.Manifest{
+		ID:          "powerball",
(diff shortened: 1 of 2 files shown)


rohansingh:
I haven't been able to find any docs on the Powerball API endpoint. Do you know what the rate limit is?

Might be best to add caching so that we don't end up making too many requests.

AmillionAirs:
Hey Rohan,

You're correct the API doesn't seem to have documentation. All i did to find it was go the Powerball website and the just inspect elements to see where they were pulling from. Powerball only draws like once every few days so i would imagine setting the cache to update twice a day should be fine right? We can always adjust later.

This is my first time with caching so if anything looks off please let me know!!

rohansingh:
Looks good but I would reduce the cache TTL to something like 600 seconds (10 minutes). If the cache interval is something like 12 hours, then we could potentially cache a value at 11:59 PM and then it wouldn't update until halfway through the next day.

AmillionAirs:
Yes sir, you're the boss.

PR #238, opened 2022-03-07
Add Strava applet

rgkimball:
Hello!

The applet in this PR implements a simple display for summary stats from [Strava](https://strava.com) using their API. You log in using OAuth and can display data from your runs, rides or swims either all-time or your YTD stats only. Currently the application is under my user, but happy to transfer this if the Tidbyt team feels that would be easier to maintain! Only the client id and secret would need to be replaced.

IMG_8355

Here are a few more images from the Pixlet environment showing other options and the relating config:

- Runs: https://i.imgur.com/h5dX9Ox.png
- Swims: https://i.imgur.com/LSlfvgI.png (apologies to swimmers for the lack of personal data here!)
- YTD Rides: https://i.imgur.com/3wAlDkP.png

Happy to hear/incorporate any feedback you may have.

### DIFF
--- a/apps/strava/strava.star
+++ b/apps/strava/strava.star
@@ -83,7 +83,7 @@ Gr1eimwzTjdSOy+wFaLiTvmqj9hwAAAABJRU5ErkJggg==
 
 def main(config):
     token = config.get("auth")
-    client_secret = OAUTH2_CLIENT_SECRET or config.get('dev_api_secret')
+    client_secret = OAUTH2_CLIENT_SECRET or config.get("dev_api_secret")
     access_token = cache.get(GLOBAL_CACHE_PREFIX + "access_token")
     refresh_token = cache.get(GLOBAL_CACHE_PREFIX + "refresh_token")
 
@@ -96,7 +96,7 @@ def main(config):
         access_token = get_access_token(token, client_secret)
 
     headers = {
-        "Authorization": "Bearer %s" % access_token
+        "Authorization": "Bearer %s" % access_token,
     }
 
     timezone = config.get("timezone") or "America/New_York"
@@ -176,11 +176,11 @@ def main(config):
     print(stats)
 
     display_header = []
-    if show_logo == 'true':
+    if show_logo == "true":
         display_header.append(render.Image(src = STRAVA_ICON))
     if period == "ytd":
         display_header.append(
-            render.Text(" %d" % year, font = "tb-8")
+            render.Text(" %d" % year, font = "tb-8"),
         )
 
     SPORT_ICON = {
@@ -193,11 +193,11 @@ def main(config):
     # more interesting statistic than speed so we"ll vary the third item:
     if sport == "ride":
         third_stat = [
-             render.Image(src = ELEV_ICON),
-             render.Text(
-                 " %s %s" % (humanize.comma(float(stats.get("elevation_gain", 0))), elevu),
-             ),
-         ]
+            render.Image(src = ELEV_ICON),
+            render.Text(
+                " %s %s" % (humanize.comma(float(stats.get("elevation_gain", 0))), elevu),
+            ),
+        ]
     else:
         if float(stats.get("distance", 0)) > 0:
             split = float(stats.get("moving_time", 0)) / float(stats.get("distance", 0))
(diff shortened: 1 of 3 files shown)


rgkimball:

rohansingh:
@rgkimball If you don't mind, I'm just going to push a commit to fix the lint issues and get this merged.

rgkimball:
> @rgkimball If you don't mind, I'm just going to push a commit to fix the lint issues and get this merged.

No objections, push away!

apps/strava/strava.star
```diff
@@ -0,0 +1,378 @@
...
+             (res.status_code, res.body()))
+
+    token_params = res.json()
+    refresh_token = token_params["refresh_token"]
+    access_token = token_params["access_token"]
+    athlete = int(float(token_params['athlete']['id']))
+
+    cache.set(GLOBAL_CACHE_PREFIX + 'athlete_id', str(athlete), ttl_seconds=CACHE_TTL)
+    cache.set(GLOBAL_CACHE_PREFIX + "access_token", access_token, ttl_seconds = int(token_params["expires_in"] - 30))
+    cache.set(GLOBAL_CACHE_PREFIX + "refresh_token", refresh_token, ttl_seconds = int(token_params["expires_in"] - 30))
```
zephyern, line 300: It's my understanding that the cache is global across all users. Are you sure you can store these tokens in the global cache? Wouldn't that mean that this app would only work for the first user who ever authenticated?

rgkimball: Is that true? I can include the athlete ID in the key name if that's the case but would appreciate if someone could confirm that the cache is not local to the device.

zephyern: This was my understanding after reading the docs here (https://github.com/tidbyt/pixlet/blob/main/docs/authoring_apps.md):

> Make sure to create cache keys that are unique for the type of information you are caching. Each app has its own cache, but that cache is shared among every user. Two installations of the same app by two different users will share the same cache.

rgkimball: Ok, is there an "instance ID" or something that I can include so people can have multiple Strava views on one device?

zephyern: I just asked the same question in my pull request. :) I was checking out your code to see how you were doing OAuth, as I'm not quite sure how to test it.

It's really clever how you put the token exchange in the render function. I could try that to test that part of my code.

Can I ask, how did you test the initial OAuth request? Did you just have it redirect the response to a server you own? Thanks!

rgkimball: Thanks! I was able to test the oauth exchange using pixlet serve, you can test your config that way as well. 

The one thing I couldn't get working locally was the secret encryption, I assume only apps on device have access to the private key?

zephyern: Oh, thanks for the advice! Let me try `pixlet serve` again.

Re: Secrets
Yep! (Same doc as above.)

> When you run pixlet locally, secret.decrypt will always return None. When your app runs in the Tidbyt cloud, secret.decrypt will return the string that you passed to pixlet encrypt.

rgkimball: Got it, thanks!!

rohansingh:
@rgkimball @zephyern After reading your discussion, I ended up making several edits here to get the auth working for multiple users.

First off, just to address this:

> It's my understanding that the cache is global across all users. Are you sure you can store these user-specific tokens in the global cache? Wouldn't that mean that this app would only work for the first user who ever authenticated?

This is totally correct. The cache is scoped per app, but it's global to all users of that app. Here's how we handle this for OAuth apps and store user-specific stuff:

1. Return the refresh token from `oauth_handler`. This token is permanently saved for that particular user.
2. Cache other stuff using the refresh token as part of the cache key.

Here's an example of how it's done in the Netatmo app:
https://github.com/tidbyt/community/blob/0c65a014ea92a86c920bf2f1a4dfcf60571be293/apps/netatmo/netatmo.star#L132

I modified the code for this app to do the same thing. The refresh token is returned from `oauth_handler`, and stuff like the access token and athlete ID are cached using the refresh token as the cache key.

---

The big reason that I _had_ to make this change is because `oauth_handler` needs to return the refresh token if it's available. This is because both the authorization code and the access token will expire after a while. So @zephyern:

> It's really clever how you put the token exchange in the render function. I could try that to test that part of my code.

That actually won't work. It will actually seem like it works locally for a bit, but then the token exchange will start failing after a while because the authorization code is only valid for a few minutes.

---

Finally, I changed the flow to return the preview stats if there's nobody logged in. We need to do this to have a real preview show up in the Tidbyt phone app, so that user's know what the app looks like.

rohansingh:
Rebasing to fix tests.

PR #243, opened 2022-03-08
Countdown Clock app

cubsaaron:
This countdown clock has a few feature enhancements over the default one. It has a more fine grained time (displays days, hours, and minutes left), can display longer event text, and has a slick fade effect.


cubsaaron:

PR #247, opened 2022-03-10
Add PurpleAir application

posburn:
Add the PurpleAir app that calculates and displays the air quality near you using data from a PurpleAir sensor. Select a sensor close to you based on your location or specify the sensor id directly. Uses the US EPA calculation method to determine the AQI. US EPA is the only method supported at this time.


posburn:

apps/purpleair/purpleair.star
```diff
@@ -0,0 +1,617 @@
...
+    location_data = json.decode(location)
+    if location_data == None:
+        return sensors
+
+    # Get latitude and longitude of location to use in the api call
+    desc = location_data.get("description", None)
+    latitude = float(location_data.get("lat", 0))
+    longitude = float(location_data.get("lng", 0))
+
+    # Get the sensor list constrained to a small area around this location
```
fxb, line 167: The best practice is to truncate these coordinates to protect the users privacy.

Here are some examples how other apps do it:

https://github.com/tidbyt/community/blob/main/apps/weathermap/weather_map.star#L284
https://github.com/tidbyt/community/blob/main/apps/gapilotbuddy/ga_pilot_buddy.star#L52

posburn: Thanks @fxb, I didn't know about that. I'll make that change.

posburn: @fxb After giving this more thought, I don't think truncating the coordinates is the right way to go. 

First, I'm never sending those exact coordinates to the service. Instead, I use them to create a bounding box of a certain size. I then send the bounding box to the service.

Second, the main use of the app is to find a sensor that is very close to you as air quality can vary greatly depending upon where you live, even just a mile or two a way.

I could increase the size of the bounding box. Right now on average I think it is about 2 square miles depending upon what the longitude value is. But I don't want the configuration list to get too big as then it becomes less useful.

So my thought right now is to not make any changes, or increase the size of the bounding box to say something like 0.1 (~ 8 or 9 square miles). Thoughts?

fxb: Technically the bounding box can easily be reversed, just by calculating its center.

0.01 degrees equals about 0.7 miles, so the delta you are using for the bounding box (+- 0.025 degrees) equals about 1.7 miles.

It probably wouldn't hurt too much to truncate the original coordinates to 0.01 degree accuracy, and your bounding box would shift by at most 0.7 miles. You could then then slightly increase the bounding box area, by 0.01 degrees (0.025 -> 0.035), which would cover the original area you wanted to (and a bit more in the direction the bounding box is now shifted)

posburn: Ok I'll try that

posburn: That seems to work well @fxb, thanks. Please take a look.

fxb: Looks good to me! :)

rohansingh:
Thanks for the contribution!

PR #254, opened 2022-03-11
Bug Fixes

rjison:
Display the next future date, skip past items. Remove cache for now -- I seemed to have gotten cache value from another users setting.


rs7q5:
In your path at the end you should have this command and that should fix it. Provided you have go installed and everything it is very likely a path issue. 

`export PATH=$PATH:$HOME/go/bin`

rjison:
#

PR #255, opened 2022-03-12
Bug Fixes

rjison:
Display the next future date, skip past items. Remove cache for now.

### DIFF
--- a/apps/spotthestation/spotthestation.go
+++ b/apps/spotthestation/spotthestation.go
@@ -17,7 +17,7 @@ func New() manifest.Manifest {
 		Name:        "SpotTheStation",
 		Author:      "Robert Ison",
 		Summary:     "Next ISS visit overhead",
-		Desc:        "Enter your spotthestation.nasa.gov location's RSS Feed URL",
+		Desc:        "Enter your spotthestation.nasa.gov location's RSS Feed URL.",
 		FileName:    "spotthestation.star",
 		PackageName: "spotthestation",
 		Source:      source,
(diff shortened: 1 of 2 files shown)


tavdog:
you should not remove data caching.

rjison:
> you should not remove data caching.

I noticed that my Tidbyt was displaying information based on someone else's settings. I don't know how that would be possible, but the only thing I could think of is the cache was server side and the XML I cached, was then overwritten by someone else.

That's why I removed it. Do you think I'm wrong here?

tavdog:
> 
The data caching is needed to keep the cloud costs low.  Without caching, every install of your app will pull the xml data every time it updates ( around every 5 minutes or so ).  This is at a time when they are trying to dial back on their cloud costs too.  

I made an edit to your code to use a location unique cache key.  It started working for most users after that.  I'm not sure if your stale data was caused by the non-unique caching problem or something else.  The current cache key is based on the full xml url which I assume is ok unless there is a character limit for the cache key.  In that case it would probably be best build the cache key based on just the last part of the url (eg.  "United_States_Florida_Orland.xml" )

rjison:
Does anyone have some advice for getting this last test to pass?

brandontod97:
Apparently they seem to be aware of other apps causing your PR to fail (which from what I saw seems to be your issue). So hopefully they'll resolve it soon. I'm not sure how others are seemingly avoiding the issue altogether though.

rjison:
@rohansingh Any advice to get this to pass the tests?

rohansingh:
Could you please rebase on `main`?

PR #256, opened 2022-03-12
Todoist Integration Tidbyt App

zephyern:
I've created an integration for the Todoist app.

I use todoist all the time and have been using this locally so that I can wake up each day and see exactly how many tasks I have to do that day. (See the README for more details about how the app works and what it displays.)

image

One question: What's the best way to cache the todoist data per user? Is it safe to use the auth-token as part of the cache key? Is there another way to identify a unique user or device?

Thanks for all of your help! Everything else worked like a charm. :)


zephyern:
Moving this PR on a different local branch. Will resubmit.

PR #261, opened 2022-03-13
Add Centipede as an animation to the Arcade Classics app

sotteson1:
This adds a third animation to the existing Arcade Classics app (which I also wrote). It randomly chooses a different color scheme and the number/location of mushrooms. The centipede will then roam around the mushrooms, eating any mushrooms it visits for the second time.
centipede


btjones:
As an owner of both a Tidbyt and an original Centipede machine, this is awesome.

PR #271, opened 2022-03-15
drop decimal of temperature as it is always 0

codylerum:
The API returns air_temperature always as a whole number so don't force it to be a float with a 0 decimal part


codylerum:
Thanks. Looks like it showing up correct now.

PR #274, opened 2022-03-18
strava: Fix scopes to work with Tidbyt app

rohansingh:
According to the RFC, OAuth 2.0 scopes should be space-delimited,
but Strava expects them to be comma-delimited.

This actually worked in the dev environment because Pixlet dev tools
mistakenly send scopes as comma-delimited. However, the Tidbyt
smartphone app sends scopes as space-delimited.

Work around this for Strava by having a single "scope" with a comma in
it.


rgkimball:
Nice catch, thanks!

PR #277, opened 2022-03-19
Abstract Clock

AmillionAirs:
I know how much Henry loves his clocks so I though...... Why not add another one? This one utilizes the 64 pixel width of the Tidbyt to display the time horizontally. I also added in a few colors for the hands so you can mess with those, and also the ability to add a hand for the month, and display the day down on the bottom. Attached is a screenshot of the basic look. I hope you guys can test the location based schema cause it wont work on my home PC.

image


rohansingh:
Still wrapping my head around how to read this but looks great 😂

PR #278, opened 2022-03-19
Add color toggle for warframe cycles app

grantmatheny:
I originally thought about just making the app in color by default, but this seemed like an ideal reason for me to learn how to use the config schemas, so I made the color a user-facing toggle (especially useful in the case of color-blind users).


rohansingh:
Looks great. Apologies for the delayed merge.

PR #281, opened 2022-03-21
New App: Happy Hour

N-Brooks:
Hello again! I'm back with a new app. This one is called Happy Hour. It uses the [CocktailDB's](https://www.thecocktaildb.com/) free random drink generator API. It displays a drink image, name, and the list of ingredients, and the cache refreshes at the top of the hour so everyone enjoys the same cocktail.

Screwdriver

In this image everything fits on the page, but if the ingredients are too long, they scroll vertically. If the drink name is too long, it scrolls horizontally.


rohansingh:
Thanks for the contribution! Looking forward to using this :)

PR #282, opened 2022-03-22
Fix world clock labels

ejb:
This is meant to address the label bug in the world clock app, which is live in the app right now and is [affecting end users](https://discord.com/channels/928484660785336380/931545988756283412/955854041269809202).

I'm unable to test this locally as Pixlet doesn't allow me to preview the location schema output. So fingers crossed!

### DIFF
--- a/apps/worldclock/world_clock.star
+++ b/apps/worldclock/world_clock.star
@@ -43,58 +43,65 @@ def main(config):
 
         now = time.now().in_location(timezone)
 
+        location_name = render.Box(
+            height = 7,
+            width = 43,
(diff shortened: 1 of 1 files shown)


ejb:
While I'm at it, I pushed up a design tweak: clocks are now on the right, and labels on the left, which seems more readable to me.

ejb:
Ok, one more thing: the location names are now coloured by daylight/nighttime. It can be disabled in the app's config.

world_clock

rohansingh:
Thanks for the bug fix!

PR #285, opened 2022-03-23
Strava: bug fix for new screen, correct date range for previous activities

rgkimball:
Apologies - this should have been in https://github.com/tidbyt/community/pull/283. The timestamp range on previous activities is wrong in the last PR, this corrects it.


rohansingh:
Thanks for the bug fix! Apologies for the delayed merge.

PR #290, opened 2022-03-24
New Applet: US Yield Curve

rgkimball:
Hi Tidbyt team,

This PR adds a new applet to the library that displays the daily yield curve, similar to how one might monitor it in a Bloomberg terminal. The data is retrieved without authentication from treasury.gov via XML and no configuration is required. The white line shown represents the latest yields, while the fading blue lines represent historical curves going back to the beginning of the year - I find this display offers a very intuitive read on the direction of the curve's movement and helps me visually assess whether it is steepening/flattening.

Finally, I include the current value of the 10-year note and the spread between the 10-year and the 2-year; two closely monitored data points by investors since they represent the base cost of long term debt financing (e.g. mortgages) and short term risk of a recession (particularly high when 10y-2y is negative).

image

Happy to address any questions or concerns about the app, thanks for reviewing!

### DIFF
--- a/apps/apps.go
+++ b/apps/apps.go
@@ -72,6 +72,7 @@ import (
 	"tidbyt.dev/community/apps/transsee"
 	"tidbyt.dev/community/apps/twitterfollows"
 	"tidbyt.dev/community/apps/unsplash"
+	"tidbyt.dev/community/apps/usyieldcurve"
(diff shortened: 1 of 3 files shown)


rohansingh:
Rebased on `main`.

rohansingh:
Looks amazing! Thanks for the contribution!

Issue #299, opened 2022-03-28
Strava Applet Not Displaying Correctly in App Preview or Tidbyt Display


OrianTheHunter:
The Strava applet is not displaying correctly on either my Android app or physical Tidbyt.  I am able to select the Strava applet and connect my account but no preview is displayed within the application and once I save and add to Tidbyt it does not display on the actual Tidbyt device as expected in the applet rotation order (it simply skips over it).  Is anyone else experiencing issues with the Strava applet integration?  Or could this be Strava account security settings or something impacting the Tidbyt connection?
![Screenshot_20220328-140356](https://user-images.githubusercontent.com/102548060/160460999-cac0a04b-e9f9-40e5-9ef2-257912653a58.png)
![Screenshot_20220328-140417](https://user-images.githubusercontent.com/102548060/160461020-6e6712b5-2c1d-4467-bdee-cb153f69ee61.png)

rgkimball:
Hey @OrianTheHunter, developer of the Strava applet here - thanks for the report.

Mine had been working earlier today presumably using cached data, but I just removed and re-added my own screens and had the same experience as you - black screen, skipped in the rotation. The demo data flashes briefly once I log in but as soon as it tries to grab my data it seems to crash.

I don't have the official API secrets but tried using my own ID/key locally and was able to pull data in again:

![image](https://user-images.githubusercontent.com/7003930/160476839-42a83901-3b02-4578-904b-b1f65546058d.png)
I suspect this means we've hit the [rate limits](https://developers.strava.com/docs/rate-limits/) - the default setting is 100 requests per 15 minutes and capped at 1,000 requests per day - it's very possible that the community has hit the level of installs that breaches these limits. I was able to reproduce the error locally by disabling the cache and refreshing until I hit the limit using my own API keys. If this is also what's causing the issue on device, the app should begin working again tomorrow and Tidbyt will need to reach out to Strava to resolve - I'm sure apps like VeloViewer have much higher limits than what's available by default!

In the interim, I'll be submitting a PR to replace the `fail` statements in the code with an error message so users aren't left in the dark when this type of thing happens You would see a screen like this instead:

![gif](https://i.imgur.com/5uX25ij.gif)

Sorry for the interruption, hope we'll get everyone back online shortly!

OrianTheHunter:
Hi @rgkimball, 

[...]

PR #300, opened 2022-03-28
Strava: Replace app fail with error message

rgkimball:
Hey @rohansingh and team,

This PR is something of a proposal; per #299 the API rate limits are causing apps to show a black screen on device and in the phone app. While this is technically a valid failure, it isn't very informative for the user who may feel they've set something up incorrectly. I assume you guys have logging for `fail` calls on your side, but it's hard for me infer the cause of problems like this when there isn't an error I can easily introspect (maybe a developer logs section of the app would help here?).

When receiving a non-200 status code, this PR displays the code and error message instead of causing an app failure. The hope is that users who run into API-related failures would include this information in future bug reports and help us sort out issues more quickly. Happy to discuss/revise as you guys see fit.

proposed screen

### DIFF
--- a/apps/strava/strava.star
+++ b/apps/strava/strava.star
@@ -149,7 +149,8 @@ def progress_chart(config, refresh_token, sport, units):
                 print("Getting %s month activities. %s" % (query, url))
                 response = http.get(url, headers = headers)
                 if response.status_code != 200:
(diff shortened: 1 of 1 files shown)


betterengineering (approved):
Looks good! One other trick you can do is return an empty list to tell the Tidbyt not to render it at all:

```
return []
```

PR #314, opened 2022-04-04
sync: Sync app list.

betterengineering:
Some of these apps were not committed properly. I've generated the proper config using:
```
 go run tools/main.go sync
 ```


saltedlolly:
And so it begins! 😀

PR #317, opened 2022-04-04
Optionally show just Sunrise or Sunset or both in Sunrisesunset

maddox:
This change adds the ability to optionally show just the sunset, or just the sunrise, or both.

When both are selected, the widget renders as it did before. When only 1 is selected, the data is rendered along with a title.

/cc @alan-fleming @betterengineering 

## Both 
both

## Sunrise 
sunrise

## Sunset 
sunset


maddox:

maddox:
🤔 I can't quite make out what the linter does not like. Is there a linter I should run against changes locally?

rs7q5:
did you run make lint on your computer?

maddox:
Seems to be a format issue on `sunrisesunset.go`? I'm unclear about this, as this file was not changed.

```
apps/sunrisesunset/sunrisesunset.go:1:3: syntax error near //
apps/sunrisesunset/sunrisesunset.go # reformat
```

rs7q5:
but did you run make lint on your computer? That formats the lint file. I'm guessing it's mad about some spacing somewhere.

If you run it on your computer it should tell you where the error is.  I think you just need to run `make lint` and if there are no formatting errors and what not you'll see it return `buildifier -r ./` and then you just need to recommit. 

Because you are right as you said you didn't change that file, but you have to run `make lint` to get the lint request to pass.

maddox:
Ah! When running it, it didn't return anything. So I thought it wasn't detecting any issues. I see now that it re-formatted for me. Thanks!

rs7q5:
No problem, yeah `make lint` just formats the `.star` file and stuff (not the lint file like in my previous comment), glad it worked!

apps/sunrisesunset/sunrise_sunset.star
```diff
@@ -126,6 +175,20 @@ def get_schema():
                 desc = "Location for which to display the sun rise and set times.",
                 icon = "place",
             ),
+            schema.Toggle(
```
betterengineering, line 178: One thought I have: If we made this a dropdown, we could eliminate the case where both sunrise and sunset would be toggled off. It could default to showing both and have options to select either or.

betterengineering (commented):
This is a nice change! Will want @alan-fleming's review though before we merge. I'd also consider removing the case of ever having both being disabled.

apps/sunrisesunset/sunrise_sunset.star
maddox: 🤦 I missed the Dropdown item in Schema. I'll do another pass 👍

maddox: OK, should be resolved with 8f6972beea61184dd90422d9f1eec6aa171069db

alan-fleming (approved):
Absolutely happy with this. Glad to see it's not just me who has not yet mastered the inscrutable ways of the Linter!

PR #319, opened 2022-04-04
New app: SF Muni predictions

juniormonkey:
This app renders predicted next arrival times for San Francisco Muni transit, for a single transit stop.

For data the app uses the NextBus API, https://retro.umoiq.com/xmlFeedDocs/NextBusXMLFeed.pdf.

To fit on the screen, the app renders a single next arrival time for the three next lines at the given stop.

A few things (line colors, the odd K/T interlining, and hacky destination abbreviating) are hard-coded in a way that is specific to SF Muni, but with a few changes this app could be extended to support any transit system that is available via NextBus.

sf_next_muni

This is my first Tidbyt app - comments and feedback are very welcome!


juniormonkey:

betterengineering (approved):
Looks great! Thanks for the app!

PR #323, opened 2022-04-06
New App: Who's That Pokemon?

N-Brooks:
whosthatpokemongif

This app uses the PokeAPI and the Imgix API to get a silhouette of a Pokemon and then switch to the revealed sprite halfway through the Tidbyt cycle. The images and the Pokemon JSON all cache for 60 days, because Pokemon almost never change so there's no need to pull them often. This also caches images in the exact same way the other Pokemon app does -- this means that we can potentially share caching and lighten the API calls even further.


rs7q5:
This takes me back, super excited for this one!

PR #327, opened 2022-04-07
Yet Another Transit App

gdcolella:
I've written another app using the TransSee API to render the next few trains (or other means of transit) coming to one or two nearby stations.

It has a few differences from the default app and the existing TransSee:

- Allows you to specify a minimum time-to-arrival (to avoid showing a train you won't be able to make).
- Allows you to combine multiple lines on the same stop in one row.
- More subway-esque iconography than the TransSee app.
- Shows multiple arrivals to each station, different than the native MTA app.

Similarly to the TransSee app, it requires a TransSee premium subscription. Thanks to the developer for helping me use the API!

moretransit


gdcolella:

gdcolella:
Thanks!! Yeah, the TransSee app separates lines by direction. So the configuration shown here is only showing Manhattan-bound trains from these stations (since those are the ones I take to commute). One of my goals for this app was actually *not* showing trains going both directions on the same page since I just want to see when I need to leave my house!

If you're interested in both directions using this app I'd recommend running two instances of it with a changed "title" (Manhattan in the screenshot) and the alternate routes configured. This way the TidByt would show a Downtown and Uptown page separately. I think that makes more sense since the trains are going totally different places - I never actually want to know the next train going _anywhere_ from a station vs knowing when the next train is to get somewhere specific.

apps/moretransit/moretransit.star
```diff
@@ -0,0 +1,213 @@
...
+
+    for combined_station in stations:
+        station_name = None
+        station_arrivals = []
+        for service, route, station_code in combined_station:
+            premiumKey = config.str("premium")
+            data = None
+
+            if premiumKey == None:
+                data = sample_response_data[route + ":" + station_code]
```
betterengineering, line 70: Will it be obvious to users that they need to provide a premium key to get real info?

betterengineering (approved):

PR #331, opened 2022-04-07
New app: USGS Earthquakes

csilv:
This app displays up to three earthquakes that have occurred during the past 24 hours, based on a configurable location. Supports settings for what radius to check and what minimum magnitude to consider. Data is provided by the USGS GeoJSON Summary Feed [https://earthquake.usgs.gov/earthquakes/feed/v1.0/geojson.php](https://earthquake.usgs.gov/earthquakes/feed/v1.0/geojson.php). Fetch results are cached and updated every five minutes.

This is my first Tidbyt app, any comments or suggestions for improvement would be greatly appreciated.

screenshot_scaled


csilv:
> Looks great, and nicely done! I was only curious about the radius, looks like 1000 is the largest you could pick, is that an api limitation? Few earthquakes within 1000 miles of Florida, I was thinking it would be cool to display only real big ones, but anywhere in the world. Would that be possible in a future release?

Thanks for raising this issue. I'll admit, my California bias might be showing in that we are constantly having (small) earthquakes, so a relatively small radius works best for us (I personally choose 100km with a minimum of 1mag). The two options I have would be to either change the radius field to text entry for the user to enter a numeric number, or just add some additional options... so, perhaps i could add 2500 and 5000, and perhaps end it with an "Unlimited" option to get world-wide earthquakes. Let me know if you have a preference either way. I lean towards adding this options for simplicity.

rjison:
Chris,

Thanks for considering it. 

If someone picks some silly combination like >1.0 and unlimited, can you limit the results?  If so, I think have drop downs makes the most sense. If you let people enter text you’ll have to validate. 

Personally, I’m interested in big ones anywhere, just for curiosity. If I move to a hotspot in the future the local ones would interest me. 

I’ll have to consider writing a hurricane app now come to think of it :)

csilv:
> Chris, Thanks for considering it. If someone picks some silly combination like >1.0 and unlimited, can you limit the results?

Yes, the API lets the caller set a limit on how many results come back and currently that limit is up to three. (I chose this value because that's the most that can be seen before the tidbyt will advance to the next step with the default duration.) So yeah, I'll go ahead and add some additional options including one for worldwide.

> I’ll have to consider writing a hurricane app now come to think of it :)

Sounds like a great idea, go for it ;-)

csilv:
@rjison Updated to add additional radius options, 2500, 5000, and unlimited. Thanks for the suggestions!

apps/usgsearthquakes/usgs_earthquakes.star
```diff
@@ -0,0 +1,299 @@
...
+""")
+
+def main(config):
+    # Get latitude and longitude from location.
+    location = config.get("location", DEFAULT_LOCATION)
+    loc = json.decode(location)
+    tz = loc.get("timezone")
+
+    # Truncate to protect the user's privacy.
+    lat = humanize.float("#.##", float(loc.get("lat")))
```
betterengineering, line 76: Thank you! 🙌

betterengineering (approved):
This looks great! Thanks for the app!

PR #343, opened 2022-04-11
New App: Traffic

rgkimball:
This PR revives the functionality of a previously high-usage app - @LukiLeu's live traffic. As you'll recall, the cost of serving Google's Maps API proved to be unsustainable and it was subsequently removed from the app library after receiving hundreds of installs. This alternative app provides a familiar display with two new sources of data, and users will need to bring their own API keys in order to keep license usage low.

- **MapQuest**: provides real time directions for driving, cycling and walking, influenced by their traffic data
- **OpenRouteService**: provides "ideal" travel times that do not reflect traffic for a broader array of travel modes: e-bikes, wheelchairs, freight and hiking.

**Also, since I'm unable to test the location selection tool from my local Pixlet emulator, I would very much appreciate some extra scrutiny from the Tidbyt team to ensure everything works as expected!**

Finally, given the sensitivity around this app and the history around [its predecessor](https://github.com/tidbyt/community/pull/228), I am happy to discuss or revise as you/the community see fit.

## Key Features

image

Like the previous app, we use `schema.Location` to set an Origin and Destination with optional labels. If the user doesn't provide labels, we will reverse-query the coordinates to get the nearest address:

ezgif-2-db12d9624b

I found better performance with ORS than MapQuest for reverse address search from GPS coordinates, so although both are available and implemented, I default to ORS if the user has provided a key for both regardless of the transportation mode selected. Additionally, when the results come back I compare the elements of the label (street address, city, state, country) and only display values that differ. Notice the example above displays NY and CT but not USA.

The API employed by each transportation mode is noted in the selection so users know which key(s) they need to use the app:

image

One key enhancement over the previous app is relative-duration color-coding. MapQuest's API provides both a "real time" and an "ideal" travel time when querying for directions, allowing us to calculate how much longer the current travel prediction is than it would be in a vacuum. We use this information to display times in green/yellow/red based on how much longer your trip may take - this proves to be a great way to monitor peak commute times and helps Tidbyt owners decide to leave earlier/later to save time! (see above vs. below for heavy vs. moderate traffic)

We also are afforded a number of route configurations with MapQuest: we can tell the API that we prefer non-toll roads, or to avoid limited access roads like highways. These will all serve to make our trip duration more accurate if users know their commute well:

image

Since we don't receive traffic information from ORS, we'll simply display times for those transportation methods in white:

image

It's worth noting that this app is written to support the extension back to Google Maps if user demand justifies its inclusion, though with 15,000 monthly requests on the free MapQuest plan and having observed reasonable accuracy, I don't personally see the need to migrate!

The one major downside to MapQuest/ORS - no public transportation support! Though with the numerous train schedule apps on the platform already, I don't see this as a blocker to adoption.

## API Key Acquisition

I included the registration URL in the description of the schema, though I'd love to hear if there is a better way to offer instructions to app users. In both cases it is easy (and free!) to register for API keys:

### [MapQuest Profile, Manage Keys](https://developer.mapquest.com/user/me/apps)
image
### [OpenRouteService Developer Dashboard](https://openrouteservice.org/dev/#/home)
image

In my testing I never came close to my personal API limits:

#### MapQuest API Limits
15,000 direction requests per month (~500/day, or one request every 2.8 minutes). We cache directions for 5 minutes, so users can easily have two screens running 24 hours per day and won't see an error until the 26th day of the month on the free plan. This is something of a balancing act since we want the information to be timely, though I'd imagine most Tidbyt users wouldn't need this screen to be on 24 hours/day! Heavy users can always[ pay MapQuest for more credits](https://developer.mapquest.com/plans) or register multiple API keys if they need more data.
#### ORS API Limits
2,000 direction requests daily (we cache for 5 minutes, so feasibly a Tidbyt user could use this screen to show 6 different ORS routes), and 1,000 reverse geocode requests (we cache coordinate pairs for 24 hours - directions will run out long before these do, which is another good reason to prioritize ORS for reverse search when the user has supplied both keys).

Thanks for reading, happy to discuss and answer questions!

### DIFF
--- a/apps/apps.go
+++ b/apps/apps.go
@@ -74,6 +74,7 @@ import (
 	"tidbyt.dev/community/apps/tempest"
 	"tidbyt.dev/community/apps/theysaidso"
 	"tidbyt.dev/community/apps/todoist"
+	"tidbyt.dev/community/apps/traffic"
 	"tidbyt.dev/community/apps/transsee"
 	"tidbyt.dev/community/apps/twitterfollows"
 	"tidbyt.dev/community/apps/unsplash"
@@ -154,6 +155,7 @@ func GetManifests() []manifest.Manifest {
 		tempest.New(),
 		theysaidso.New(),
 		todoist.New(),
+		traffic.New(),
 		transsee.New(),
 		twitterfollows.New(),
 		unsplash.New(),
--- a/apps/traffic/traffic.go
+++ b/apps/traffic/traffic.go
@@ -1,0 +1,25 @@
+// Package traffic provides details for the Traffic applet.
+package traffic
+
+import (
+	_ "embed"
+
+	"tidbyt.dev/community/apps/manifest"
+)
+
+//go:embed traffic.star
+var source []byte
+
+// New creates a new instance of the Traffic applet.
+func New() manifest.Manifest {
+	return manifest.Manifest{
+		ID:          "traffic",
+		Name:        "Traffic",
+		Author:      "Rob Kimball",
+		Summary:     "Time to your destination",
+		Desc:        "Shows the duration to get from an origin to a destination by using traffic information from MapQuest.",
+		FileName:    "traffic.star",
+		PackageName: "traffic",
+		Source:  source,
+	}
+}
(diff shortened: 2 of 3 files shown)


maddox:
This is great! Was a Waze option considered?

I know that Home Assistant has a great component for drive times using Waze.

rgkimball:
> This is great! Was a Waze option considered?

I didn't look at Waze specifically, but it certainly could be implemented! The code is written to be modular so it is amenable to new services.

betterengineering (approved):
This looks great! Thanks @rgkimball!

PR #344, opened 2022-04-12
mbta: remove timezone parameter

marcusb:


marcusb:
App doesn't work in prod. Simplifying the code a bit, for lack of better ideas.

PR #345, opened 2022-04-12
New App: Climate Clock

rgkimball:
This is a relatively simple, yet impactful app. We display two screens here relating to climate change:

- **The Deadline**: a countdown to the estimated point where a 1.5°C increase in global temperatures will be irreversible
- **The Lifeline**: the percent of the world's energy generated by renewable sources (wind, solar, hydro etc.)

NYC locals may recognize the first as the countdown that has displayed in Union Square above the Best Buy since 2020. Last year, similar clocks were also installed in Glasgow and Seoul. The origin story of this clock is related to Greta Thurnberg's UN speech and is intended to spread awareness about the urgency of climate change. Tidbyt owners who decide to display one or both of these clocks will have a chance to start a conversation with the people around them 

Much more information can be found about the clocks, underlying data and their proposals here: https://climateclock.world

## Screens

Although ClimateClock.world does offer a free API, the numbers can be easily calculated and their anchor points change infrequently. This app was therefore built with no external dependencies, and we can update the underlying data if our clocks become out of sync with the official ones.

The Lifeline

The Deadline

I don't consider myself to be much of an artist, so any enhancements to the graphics from the community are welcome!

There are also other numbers that could be implemented if people found them to be interesting and wanted to contribute graphics/PRs!

- AUM of the Green Climate Fund
- Land protected by indigenous people
- Climate change news feed (would require us to turn on the API, not terribly complicated)
- Other climate data not sourced from ClimateClock

### DIFF
--- a/apps/climateclock/climate_clock.star
+++ b/apps/climateclock/climate_clock.star
@@ -24,47 +24,47 @@ ALL_DATA = {
             "display": {
                 "deadline": {
                     "color_primary": "#eb1c23",
-                    "color_secondary": "#eb1c23"
+                    "color_secondary": "#eb1c23",
                 },
                 "lifeline": {
                     "color_primary": "#4aa1cc",
-                    "color_secondary": "#4aa1cc"
+                    "color_secondary": "#4aa1cc",
                 },
                 "neutral": {
                     "color_primary": "#ffffff",
-                    "color_secondary": "#ffffff"
+                    "color_secondary": "#ffffff",
                 },
                 "newsfeed": {
-                    "separator": " | "
+                    "separator": " | ",
                 },
                 "timer": {
                     "unit_labels": {
                         "day": [
                             "DAY",
-                            "D"
+                            "D",
                         ],
                         "days": [
                             "DAYS",
-                            "D"
+                            "D",
                         ],
                         "year": [
                             "YEAR",
                             "YR",
-                            "Y"
+                            "Y",
                         ],
                         "years": [
                             "YEARS",
                             "YRS",
-                            "Y"
-                        ]
-                    }
-                }
+                            "Y",
+                        ],
+                    },
+                },
             },
             "modules": [
                 "carbon_deadline_1",
                 "renewables_1",
-                "newsfeed_1"
-            ]
+                "newsfeed_1",
+            ],
         },
         "modules": {
             "carbon_deadline_1": {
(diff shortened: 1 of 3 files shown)


rgkimball:
I'm not sure the linter has it right here. If I make the suggested changes I'll be breaking the values:

```
                 "lang": "en",
-                "rate": 2.0428359571070087e-8,
-                "resolution": 1e-9,
+                "rate": 2.0428359571070087e - 8,
+                "resolution": 1e - 9,
                 "timestamp": "2020-01-01T00:00:00+00:00",
```

apps/climateclock/climate_clock.star
```diff
@@ -0,0 +1,415 @@
...
+                "flavor": "lifeline",
+                "growth": "linear",
+                "initial": 11.4,
+                "labels": [
+                    "WORLD'S ENERGY FROM RENEWABLES",
+                    "GLOBAL RENEWABLE ENERGY",
+                    "RENEWABLES",
+                ],
+                "lang": "en",
+                "rate": 2.0428359571070087e-8,
```
betterengineering, line 215:
```suggestion
                "rate": "2.0428359571070087e-8",
```

apps/climateclock/climate_clock.star
```diff
@@ -0,0 +1,415 @@
...
+
+def renewables(DATA):
+    fps = 20
+
+    data = DATA["data"]["modules"]["renewables_1"]
+    initial = data["initial"]  # 11.4
+    units = data["unit_labels"][0]
+    rate = data["rate"]
+    start = time.parse_time(data["timestamp"])
+    resolution = int(str(data["rate"])[-1]) + 1
```
betterengineering, line 276:
```suggestion
    resolution = int(data["rate"][-1]) + 1
```

apps/climateclock/climate_clock.star
```diff
@@ -0,0 +1,415 @@
...
+                "flavor": "lifeline",
+                "growth": "linear",
+                "initial": 9.52,
+                "labels": [
+                    "GREEN CLIMATE FUND",
+                    "CLIMATE FUND",
+                    "GCF",
+                ],
+                "lang": "en",
+                "rate": 0,
```
betterengineering, line 94:
```suggestion
                "rate": "0",
```

apps/climateclock/climate_clock.star
```diff
@@ -0,0 +1,415 @@
...
+                "flavor": "lifeline",
+                "growth": "linear",
+                "initial": 43.5,
+                "labels": [
+                    "LAND PROTECTED BY INDIGENOUS PEOPLE",
+                    "INDIGENOUS PROTECTED LAND",
+                    "INDIGENOUS PROTECTED",
+                ],
+                "lang": "en",
+                "rate": 0,
```
betterengineering, line 114:
```suggestion
                "rate": "0",
```

betterengineering (approved):
The linter is wrong, but we don't have a great way to tell it it's wrong 😅 . What about making it a string? It sucks since that means we'd have to make sure it's a string when updated. But the benefit is the linter won't accidentally change this line in the future.

betterengineering:
I looked into simply ignoring these lines, hoping we could just add a `# ignore` line and it doesn't appear supported at all 😢 :https://github.com/bazelbuild/buildtools/issues/890. So sorry about this, I feel like this should be something we could configure.

rgkimball:
Had a feeling you might say that, fixed in the last couple commits per your comments. Should be good to go now!

betterengineering (approved):

PR #353, opened 2022-04-17
Strava: Add Last Activity Screen

rgkimball:
Hello!

By community request, this PR adds a new screen to the Strava app (thanks to @rjison for the great suggestion and feedback!). Currently we cover athlete summary statistics (all-time or YTD), and a plot of monthly activities compared with the previous month. This adds a third screen showing details of your most recent activity:

last activity demo

Given the issues with the API recently I feel it's important to note that the new screen uses the same data request as the monthly summary, so Tidbyt users who use both screens will still only be using one API call per 6 hours to do so. This applies across sports too: if a user shows three "Last activity" screens for running, cycling, and swimming plus a monthly summary, we use the data cached from the first historical activity request to supply data for all four.

Since the relevant information can vary from activity to activity, the screen makes some assumptions about what users would find most relevant and hides everything else:

- Average heart rate is only shown if you've decided to make this public in your activity privacy settings. Otherwise, we show another proxy for effort: average watts for cycling (unless it's zero), or calories burned in all other cases.
- If you achieved any personal records you'll see a gold trophy with the number of PRs, otherwise this is hidden
- If you had other achievements (2nd or 3rd place efforts), those will be enumerated with the hollow trophy. If you had no achievements, bike rides will show the elevation if it is above zero.
- If you've received any Kudos on your activity a count will be displayed, otherwise it will be hidden (shown below)

Low-data activity

- If the activity was virtual or stationary (i.e. there is no map), we display a third statistic in the header: either the total moving time for runs and swims, or the mile/km pace for rides; time is displayed by default for the latter, so whichever of the three is missing from the "with map" version
No-map activity

This screen only queries activities since the beginning of the previous month, so the display is intentionally hidden (`return []`) if there have been no activities. Should be good motivation for people to get outside if it's been a while!

Finally, this makes some small revisions to the existing screens:

- Per discussion in #299, provides better support for indoor cyclists by displaying mile or km pace instead of elevation when data is unavailable (thanks @OrianTheHunter!). On the monthly summary we display an activity count now instead (same as current behavior for runs/swims)
- Updates all `render.Plot` functions to switch arguments from `xlim` and `ylim` to `x_lim` and `y_lim` per latest version of Pixlet - I assume the rendering engine expects the same change, let me know if I'm mistaken!
- Fixes round function to not floor all calculations; we can use `round(x, 0)` if removing significant digits is desirable, otherwise we should just use `int()`
- Animates the monthly summary chart upon first load
- Animate the athlete summary stats on first load


rgkimball:
Noting that this also fixes the "xlim"/"ylim"->"x_lim"/"y_lim" problem that is currently causing the monthly summary screen to error out.

betterengineering (approved):
Thanks again @rgkimball! This looks amazing. One fear I have is it may be too expensive to render server side, but I think it's worth a go and we can follow up in Discord if there are performance issues with this app.

PR #354, opened 2022-04-18
NYC Bus: Formatting suggestions

rgkimball:
Hey all! As someone who lives much closer to a bus stop than a subway, I really love @samandmoore's NYC Bus app. One of the stops I've configured has buses from an express SBS line, so the name of the bus overflows within the box:

ezgif-5-4af7bc9820

Technically I still know which bus it is, but figured this could be cleaned up a bit! With Sam's blessing, this enhancement will split any bus codes exceeding 4 characters into two parts and alternate between them:

ezgif-5-47764d3f71

You may also notice that I took a moment to realign the text and fonts to match the official Tidbyt Subway & Ferry apps and make the whole experience feel a bit more cohesive. If you compare side-by-side (and when they are alternating on the device), there are a few minor differences that this PR resolves: border position (-1px) and color, marquee font, marquee speed, ETA font and ETA color:

## Original
image

## Revised
image

Hope this contribution is welcome, though I understand if Sam would prefer to keep things the way they are. Love the app either way - thanks for considering!

### DIFF
--- a/apps/nycbus/nyc_bus.star
+++ b/apps/nycbus/nyc_bus.star
@@ -136,7 +136,7 @@ def build_row(journey):
         anim = []
         for part in parts:
             anim.extend(
-                [render.Text(part, color = "#000", font = "CG-pixel-4x5-mono")] * 20
+                [render.Text(part, color = "#000", font = "CG-pixel-4x5-mono")] * 20,
             )
 
         line_name = render.Animation(children = anim)
--- a/apps/nycbus/nyc_bus.star
+++ b/apps/nycbus/nyc_bus.star
@@ -5,6 +5,7 @@ Description: Real time bus departures for your preferred stop.
 Author: samandmoore
 """
 
+load("re.star", "re")
 load("cache.star", "cache")
 load("encoding/base64.star", "base64")
 load("encoding/json.star", "json")
@@ -98,15 +99,16 @@ def main(config):
         )
 
     return render.Root(
+        delay = 75,
         child = render.Column(
             expanded = True,
-            main_align = "space_evenly",
+            main_align = "start",
             children = [
                 build_row(journeys[0]),
                 render.Box(
                     width = 64,
                     height = 1,
-                    color = "#aaa",
+                    color = "#666",
                 ),
                 build_row(journeys[1]),
             ],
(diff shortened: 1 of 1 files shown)


samandmoore:
Wow! I love these changes @rgkimball.

I had wanted to match the styling and layout of the subway and ferry apps but ran out of patience 😅

As for the text splitting approach for the bus lines, that seems fine to me. Getting 4 character bus line names to fit was hard enough initially. This seems like a smart approach for these longer ones. I forgot that sbs lines show up that way.

samandmoore:
@rgkimball I do wonder though, is the SBS text significant? I wonder if we should just trim it? I forget how it works but I didn't think there were SBS and non SBS variants of the same line, are there?

apps/nycbus/nyc_bus.star
```diff
@@ -98,40 +99,80 @@ def main(config):
...
+        _, borough, first, sep, second = match[0]
+
+        # Lines like "M55/56" should translate to -> M55, M56
+        if sep in ("\\", "/"):
+            parts = [borough + first, borough + second]
+        elif sep == "-":
+            # Lines like "Bx41-SBS"
+            parts = journey["line_name"].split("-")
+        else:
+            # Lines like "M44 Ltd"
```
samandmoore, line 132: I guess this example is relevant to my top level question about whether we need the sbs or ltd.

Also I just searched and confirmed there are sbs and non variants for the same lines, like the bx12

samandmoore:
LGTM!

rgkimball:
> @rgkimball I do wonder though, is the SBS text significant? I wonder if we should just trim it? I forget how it works but I didn't think there were SBS and non SBS variants of the same line, are there?

Glad you appreciate the changes! The M86 doesn't have a non-SBS line but I do often take the M15 up 1st Av which has both local and express routes that stop in the same place along the way: both stop at 14th, 25th, 28th and so on, while the local also stops at 17th, 20th, 23rd and 27th.

- https://bustime.mta.info/m/?q=M15
- https://bustime.mta.info/m/?q=M15-SBS

apps/nycbus/nyc_bus.star
rgkimball: Admittedly I'm not sure if you would see "Ltd" come through your feed - I included it for completeness based on the table I was referencing here: https://metro.fandom.com/wiki/List_of_bus_routes_in_Manhattan

Same goes for M55/56 - this might just come through as whichever of the two arrives. If "-SBS" is the only possible suffix I'm open to simplifying this bit if you'd prefer!

samandmoore: I suspect that the ltd won't be in the bus code we're using here but I'm fine to leave this as is.

betterengineering (approved):

PR #359, opened 2022-04-20
New App: Pollen Count

N-Brooks:
This app uses the Tomorrow.io API to display a pollen count app for a general area, with the value from 1 - 5 of the total pollen and the type of pollen in the air. Because this is a free API, the calls are limited, but I'm hoping that won't be an issue (at least for a while). This caches your latitude and longitude, rounded to the nearest 0.5 (so people in the same ~50km area can share the same data), for 12 hours. So it will hit a rate limit if 250 different ~50km areas use the app.

pollencountpreview

The API Key used here is just a free key anyone can sign up for, so nothing to protect here.


apps/pollencount/pollen_count.star
```diff
@@ -0,0 +1,318 @@
...
+    "locality": "",
+}
+
+COLORS = {
+    "yellow": "#D19C21",
+    "red": "#B31F0E",
+    "green": "#338722",
+}
+
+API_URL_BASE = "https://api.tomorrow.io/v4/timelines?&fields=treeIndex,weedIndex,grassIndex&apikey=d4N7IbI70OfDtNh56VihCnwi3Crg2th6&timesteps=1d&location="
```
betterengineering, line 27: Is this a sensitive API key? If so, please generate a new API key and add this as a secret 🙀  https://github.com/tidbyt/pixlet/blob/main/docs/modules.md#pixlet-module-secret

apps/pollencount/pollen_count.star
```diff
@@ -0,0 +1,318 @@
...
+
+def main(config):
+    print("Initializing Pollen Count...")
+
+    #Get lat and long from schema.
+    location = config.get("location")
+    loc = json.decode(location) if location else DEFAULT_LOC
+
+    #Round to 1 decimal place (1.1km)
+    lat = roundToHalf(loc.get("lat"))
```
betterengineering, line 37: Thank you!! 🙌

betterengineering (commented):
Thanks for the app @N-Brooks! I just want to confirm the token is ok to be in plain text before merging.

apps/pollencount/pollen_count.star
N-Brooks: It's a free API that anyone can sign up for and has no potential I can see for misuse (the worst people could do with it is make other requests instead of making their own free account and use up the rate limit lol), but for absolute safety's sake I can do that.

N-Brooks: I've updated with a new, encrypted API key. Let me know :)

betterengineering (approved):
Thanks for the secret changes!

PR #364, opened 2022-04-23
New App: Wanted Poster

rjison:
My daughter thought it would be funny to make a Wanted Poster for our Tidbyt.  So why not, you can upload a picture, set the part you want displayed, and it'll appear in an old western style wanted poster.. in 64x32 pixels, so identification and apprehension may be difficult.


maddox:
Example screenshot would be great

rjison:
> Example screenshot would be great

Done

PR #368, opened 2022-04-24
US Yield Curve: Add zoom feature

rgkimball:
Hi!

Following some community feedback here, this simple PR adds a zoomed-in view of the US yield curve for short, medium and long term bonds, following industry conventions for ranges. Each display shows the spread across visible maturities only, as well as the current yield of the longest maturity bond in view.

This requires no additional data, and the default display is unchanged; the "All" view will still show the 10 year yield and 10Y-2Y spread since this is the most popular indicator.

image

Thanks to golashes from Discord for the suggestion!

### DIFF
--- a/apps/usyieldcurve/us_yield_curve.star
+++ b/apps/usyieldcurve/us_yield_curve.star
@@ -1,6 +1,6 @@
 """
 Applet: US Yield Curve
-Summary: Plots treasury rates
+Summary: Plots the US yield curve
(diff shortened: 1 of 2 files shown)


rgkimball:
Noting that this also fixes the "ylim"->"y_lim" problem that is currently causing the app to error out.

betterengineering (approved):

PR #371, opened 2022-04-26
Analog Time: Add Custom color options

rs7q5:
User can now customize the colors of the analog time clock

### DIFF
--- a/apps/analogtime/analog_time.star
+++ b/apps/analogtime/analog_time.star
@@ -1,19 +1,20 @@
 """
 Applet: Analog Time
 Summary: Show time analog style
-Description: Shows the time on an analog style clock.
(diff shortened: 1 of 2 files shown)


apps/analogtime/analog_time.star
```diff
@@ -102,6 +120,41 @@ def get_schema():
...
+            ),
+            schema.Text(
+                id = "hour_color",
+                name = "Hour hand",
+                desc = "Default color is #a00.",
+                icon = "brush",
+                default = "#a00",
+            ),
+            schema.Text(
+                id = "minute_color",
```
betterengineering, line 152: We really need a color picker 🤦

betterengineering (approved):

apps/analogtime/analog_time.star
rs7q5: Glad I could be of assistance.

PR #382, opened 2022-05-02
bugfix: Fixup natdex.

betterengineering:
This commit adds the app to the list of apps and fixes the summary so
that the tests will pass.


betterengineering:
cc @laurenkopac - I had to shorten your summary so the tests would pass 😭 . Feel free to follow up with an additional PR if you want to see changes before we launch it!

PR #386, opened 2022-05-03
dailykanji: Fix API key.

betterengineering:
This commit sets our own API key for the daily kanji app.


betterengineering:
cc @rjison

PR #388, opened 2022-05-03
Create an app that tells you where it's 5 o'clock

grantmatheny:
Based on an idea in the discord, I whipped up this app last night.


grantmatheny:
realized I can't start an app name with "5" closing and will re-open with new app.

PR #389, opened 2022-05-03
Create five o'clock somewhere app

grantmatheny:
Based on a conversation in Discord. Earlier PR rescinded due to app name starting with a 5. This one should be otherwise the same.


betterengineering (approved):
🎉

PR #390, opened 2022-05-03
Add cache to r/place app

funkfinger:
Adding cache to r/place app - I don't know how the backend works and how or if I can test this using the render / push commands, but It seems to work the first time around, and I assume it will be cached in future fetches. Also fixed the image to not be a gif as it's excluded by `.gitignore`


rs7q5:
You can test using `pixlet serve` and maybe set the cache expire time super short to see that it does what you expect.

betterengineering (approved):
LGTM!

PR #393, opened 2022-05-04
fixing image hold - lasts longer than app switch

funkfinger:
changing the image hold time from 1500 to 1 to allow for correct app switching - the it appears as if the `animation.AnimatedPositioned` `hold`  (and likely but not tested `duration`) property overrides the app switch delay


funkfinger:
THANKS!

PR #397, opened 2022-05-04
Vertical Message: Add line spacing option

rs7q5:
User can now adjust line spacing of vertical message

### DIFF
--- a/apps/verticalmessage/vertical_message.star
+++ b/apps/verticalmessage/vertical_message.star
@@ -7,7 +7,7 @@ Author: rs7q5
 
 #vertical_message.star
 #Created 20220221 RIS
-#Last Modified 20220426 RIS
+#Last Modified 20220504 RIS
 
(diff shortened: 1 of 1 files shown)


apps/verticalmessage/vertical_message.star
```diff
@@ -81,6 +89,20 @@ def get_schema():
                 default = "tb-8",
                 options = fonts,
             ),
+            schema.Text(
+                id = "linespacing",
+                name = "Line Spacing",
```
betterengineering, line 94: We really, really need a number schema field 😭

betterengineering (approved):

apps/verticalmessage/vertical_message.star
rs7q5: I would appreciate it yes :)

You'll then be getting a ton of PRs though

PR #403, opened 2022-05-04
goodservice: Fix direction issue.

betterengineering:
This commit fixes a config option for direction. cc @blahblahblah-


blahblahblah-:
Thanks for the fix!

PR #406, opened 2022-05-04
ifparank: Set Tidbyt api key.

betterengineering:
This commit updates the API key to use one that Tidbyt has access to.


betterengineering:
cc @cubsaaron

PR #415, opened 2022-05-06
Decouple price/percent truncation

cgrin:
Percent truncation is currently applied based on the price truncation toggle. This fixes to use the percent truncation toggle.


cgrin:

betterengineering:
Thanks for the fix @cgrin! cc @ethanfuerst

PR #425, opened 2022-05-09
New app: FullyBinaryTime

dinosaursrarr:
Create fully binary time app

Clock app showing the time in fully binary format, where we recursively divide the day into twos.

See here for context:  http://anthony.liekens.net/index.php/Misc/TrueBinaryTime


betterengineering (approved):
🤓

PR #429, opened 2022-05-10
New App: Tube Status

dinosaursrarr:
Shows the current status of transport lines in London


dinosaursrarr:
Is there a way to get the marquee to go back to the beginning without adding black space? I'd love it were a continuous scroll, with "Bakerloo" coming straight after "Tram" with no gap.

rs7q5:
Then I suspect the last "frame" that is scrolling is just over a full screen possibly. I'm not exactly 100% sure.

dinosaursrarr:
Actually, the black space is a moot issue. It only shows up with `pixlet serve` and the gif. The `delay` is set so that Tram only appears a second or two before the end of the 15 second app cycle.

PR #440, opened 2022-05-13
New App: FitbitWeight

rjison:
FitbitWeight - Gets data from fitbit using OAuth2.0. Dislpays weight, but also, optionally, bmi or fat %.


betterengineering (approved):
This looks great! Just an FYI, we will want to replace the oauth client secret and ID with our own before we can launch this app

rjison:
@betterengineering Thank you! Looking forward to seeing this app on my Tidbyt!

PR #442, opened 2022-05-14
New App: Game of Life

dinosaursrarr:
A simple implementation as we have a limited screen size. Wraps around the edges of the screen to keep things more interesting.

https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life


dinosaursrarr:
> @dinosaursrarr any chance you can attach a gif or screenshot of it in the PR?

There already is one, it just doesn't show the diff by default.

betterengineering (approved):

PR #449, opened 2022-05-21
Created Burger of the Day applet

kstardesigns:
burgerotd

"Burger of the Day" applet, based on the burger board behind the counter at Bob's Burgers! It shows a set burger each day. It can also be configured to show a random burger every cycle, or a custom burger if a user has an idea for their own punny burger.


kstardesigns:

austinrfnd:
This is going to be AMAZING!

betterengineering (approved):
One thing to note, this needs to be in `apps/apps.go` to be able to be picked up. With that, a few things will fail tests. Happy to merge as is and I can follow up with a few fixes

PR #453, opened 2022-05-26
New App: TV Quotes

rs7q5:
Display a random TV quote from a random show or a random quote from a selected show.

quotes are cached for 5 minutes

edit: quotes come from the following API: https://github.com/alakhpc/tv-quotes-api

edit2: the preview below has the show in `tb-8` font but it's actually in `CG-pixel-3x5-mono` Also looks like the quote here is too long (10 times the height of the screen, so it will get cutoff, but there's nothing I can do about that)
tv_quotes


mrmcpowned:
> edit2: the preview below has the show in tb-8 font but it's actually in CG-pixel-3x5-mono Also looks like the quote here is too long (10 times the height of the screen, so it will get cutoff, but there's nothing I can do about that)

You might be able to solve this by requesting multiple quotes at once, and filtering out ones past a certain character limit, and then locally picking a random one from the filtered list. You could also include this as a user togglable setting so they can choose if they want to omit longer quotes or not.

rs7q5:
> You might be able to solve this by requesting multiple quotes at once, and filtering out ones past a certain character limit, and then locally picking a random one from the filtered list. You could also include this as a user togglable setting so they can choose if they want to omit longer quotes or not.

The api only allows a max of 10 quotes at a time so picking from a filtered list wouldn't work. Since the app switches over within a certain time frame you probably wouldn't notice it since it'd never get there most of the time. 

But I did think about that though!

rs7q5:
Was able to get an endpoint added to return quotes from a single character, which should in theory produce shorter quotes overall.

PR #457, opened 2022-06-01
Indego Stations Upload

raypatt:


apps/indegostations/indego_stations.star
```diff
@@ -0,0 +1,121 @@
...
+
+def main(config):
+    rep = http.get(url)
```
betterengineering, line 22: Might be worth adding a cache?

betterengineering (approved):

PR #459, opened 2022-06-04
New app: Tartan

dinosaursrarr:
Weaves and displays a random tartan

Probably a bit cleverer than it needs to be, as it weaves pixels in detail, and then loses all the detail when resizing to fit on a tidbyt screen.

Has a library with over 3,000 patterns to choose from. Default mode is to display a random one each time it runs.

Nice way for Tidbyt users to connect with their Scottish heritage, or just enjoy some pleasing low-resolution patterns.

It looks like a massive diff, but the vast majority of it is just a dictionary of available input patterns and their names.

At 15 seconds a pop, there is nearly 15 hours of solid entertainment in this app. That's a lot of replay value.

60MB archive containing all the patterns available on dropbox. Long text will scroll on tidbyt, but not in these stills.
https://www.dropbox.com/s/e6fuk7oocmzzd70/tiny_tartans.zip?dl=0


betterengineering (approved):
TIL! This is very fun!

PR #468, opened 2022-06-10
indegostations: Temporarily disable app.

betterengineering:
This app is failing HTTP reqeusts when getting stations. The issue is, because the request is in get_schema, our server fails to load the app all together. The Tidbyt team needs to figure out why this request is failing on our end since it does work with pixlet and we need to figure out a better way to ensure our server can startup when an app fails to create schema. I'm disabling the app for now so that we can get the rest of the apps out 😿


betterengineering:
cc @raypatt

PR #479, opened 2022-06-22
Added Shopify App

kcharwood:
I created an app to display daily metrics from a Shopify store.

- Revenue
- Orders
- Units


kcharwood:

kcharwood:
Looks like I need to review the lint errors

kcharwood:
Not sure how to decipher the `make lint` error...

```
make lint                                                                                                                                
buildifier -r ./
apps/shopify/shopify.star:453:26: syntax error near should_continue
apps/shopify/shopify.star # reformat
make: *** [lint] Error 1
```

rs7q5:
It doesn't like line 453. looking at the source code. It's because you have while loop(s), you can't use those as infinite loops are bad. 

> apps/shopify/shopify.star:453:26: syntax error near should_continue

kcharwood:
Ok I'll rewrite that section.

kcharwood:
Although I'm paginating an unknown data set, so while loop does make since..

kcharwood:
Ok I removed the while loop, and went with a for loop with a known number of pages.

kcharwood:
@betterengineering @rohansingh @matslina anything else you need from me here?

apps/shopify/shopify.star
```diff
@@ -0,0 +1,554 @@
...
+            if ex_sku in sku:
+                # print("{} is excluded".format(sku))
+                return True
+
+    # print("{} is not excluded".format(sku))
+
+    return False
+
+def cache_orders(orders, start_time, store_name, api_token):
+    cache_key = "orders_{}_{}_{}".format(start_time, store_name, api_token[-6:] if len(api_token) >= 1 else "")
```
matslina, line 458: It would be good to use the full `api_token` here. Limiting ourselves to the last 6 characters makes it easier for an attacker to guess a token suffix and access the cached order data. The ideal might be to compute a hash of the full `api_token` and `store_name` and use that as cache key.

apps/shopify/shopify.star
```diff
@@ -0,0 +1,554 @@
...
+                desc = "The API Token for your Shopify Private App",
+                icon = "key",
+            ),
+            schema.Text(
+                id = "display_name",
+                name = "Company Name",
+                desc = "The company name to display",
+                icon = "signHanging",
+            ),
+            schema.Dropdown(
```
matslina, line 253: A simpler way of allowing a custom timezone to be selected is to use `schema.Location`. The location will default to the users device location, but allows the user to select some other location if so desired. The location object itself includes a timezone field.

Another approach is to drop timezone entirely, and use the magic `$tz` variable that's passed by the Tidbyt runtime environment. It's automatically set to the users device location's timezone. You can find examples by searching for `$tz` in this repo.

apps/shopify/shopify.star
```diff
@@ -0,0 +1,554 @@
...
...TCi7P0OkbiHSlVLKIkcZAGXtTUDSKZ7DH2Ulnzfi8R3YiMfe7oJ26RFAAJAeJ5RyEAFONmR0lJwu5dTpiznh0BKh5McdpD5UtQsBKf7EqjTzO/2dJe2To8uWbeuzi2rQAwSyQQABQDYoog1bE9gUq7w6IYxqkgmeSKjKWFmsKLC1x/KkHIkzvPS0jYf291DCaKydv/sPedIEYkHAEgIIACzBDCF2IWCmIn6n5PRcTspSyeuyF3AmwhswkdAu3rFWj3Pv8sVv+EewiV8d7Sk+MfllPOVb6wNIyy8BBAD55Q/peSZQF/NPluKiOTzMO48nE35aK7oZkwnz7JQciT/X4avfkqRf8w/fi4U91IJZ+zmCjWYdQQABgCPcBCWtImAGBEJPnMtDwfOkotk8UjCb/5NMsUo+5GSRAFGnkPIAj/Ts5/kgL3b10csPVDV1ZVECmgIBRxNAAOBo90H5XBMwMxJeOu/A1dqnZ/OmRXN4rwIzKLiG9ysoyLVstJ8+AfZJP3fyr3Gy/QNca79QxoHQzbveYD9x/48PCIBAKgIIAFJRwTkQGIHApldmFvSfKv6oFGqmUDSTXx3M5G7meh41mDxCNVzKFoGByXryTd4v4hAP5x/kwOxg70VFB7GVbrYAox2vEEAA4BVPw86cE/hh04IPGAXqGn7kvJbzEFzDQcG1nBP+b5GTYKzoqZf3gvgjz8l4jZke4qf517SmQyf2lb6O3PpjZYp6IPAeAQQA77HAEQhknQAPTcv6fZWXcRKZq3ik4EoWcJUQ6ipB+kru1K70fHBgJtjR4k+cQ/+wVuIwj9j/kVdlHO7nf6v9jccwhJ/1WxINgsAgAQQAgyhwAALWE6h/YdH7E5ou8+n4h3hbg8s5KLhMaLpcSvqAJjWDXyuUSNIlPJlNWa/dOCRyOkaSqoNHQTqU1G8TyT9zQqa3OPB5izv5/40r39G4LnxrbeC5znFIQVUQAIFxEEAAMA54qAoCVhA4l9mwq4SfkEt8cTL/TpNCTyWhpir+mxByKv9HniqEnsSJjjgFMk0QUk3k1w8Xc9hQyJ3wFH5fnhRAnLv+/3Qn3c3v0nsGz0mpOfh4l7To41GKLl5Cd5ZHMHpJ6tM8gnGWA5VOQ1CnFoo7cDqpzBn3pE7GfbJDadHR0fZ37RimH6SJAxCwJYH/Az7x75xxnfKnAAAAAElFTkSuQmCC
```
matslina, line 18: To save some space, here's the same graphic pre-scaled to 24x24 with exif data removed:
```
iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAMAAADXqc3KAAAABGdBTUEAALGPC/xhBQAAACBjSFJNAAB6JgAAgIQAAPoAAACA6AAAdTAAAOpgAAA6mAAAF3CculE8AAABI1BMVEUAAACWv0iWv0iWv0iWv0iWv0iWv0iWv0iWv0iWv0iWv0iWv0iWv0iWv0iWv0iWv0iWv0iWv0iWv0idw1TP4qzi7s3O4aqfxVjS5LGz0XvN4ajS5LDP4quix16tzW+81ovT5LK10n/q8tr2+vD+/v3D25epymmcw1PZ6L3f7Mj8/fr////9/vyZwU251YanyWXX57mvznSqy2rp8tm204Dd6sPX5rmkyGGgxVn6/PbG3Jvq8tuxz3emyWTV5bb4+/K304K+147l79Hy9+imyWPG3Jz0+Oyqy2v2+e/3+vHo8divznP///6bwlCcwlKkx2DF3Jqvz3TC2pWy0HjY57zs8979/vvJ3qGqy2zS47Dm79P5+/Tf7MfI3qCx0HibwlGXv0nUrRiRAAAAEnRSTlMAInq75vu8IKn9IVzzdv51/HlGNAFfAAAAAWJLR0QrJLnkCAAAAAFvck5UAc+id5oAAAEiSURBVCjPdVLnWgJBDMw17jiQsjZU7ChqBHsBC6BS7Nh7ef+n8JKscvh9zJ/cztwm2UkABIZp2Y5jR0wDwnC9qNKIen6Hj8VVCPHYL9+XUF1IJPX/wvcPDA4Na4Xv+DpPZmR0LDuus1EdT74nJqemZ2Zzc/N88oI+pZ/8wuISEpYLdEwZYDJfFJawwoQJEY5ZxNW19Q0SNpmwwOa4hdtqZ7dEQpkJGxyOe7h/gDlOdciEo4UKVlVN6hxpgVPlEY9PVJ2Fhk5lUWhi6/RMnbNwcSnFud0qZq6uVVv6bRKTlge2sVK+uS2JUCTzDbHk7v7h8e+FT2IJuGJi4bn28vrWev/4/ApMdEO2E77DtgMkewyq92iDYXmpzjK4XXvip/+tzw/wsDO/5t4LZQAAAABJRU5ErkJggg==
```

apps/shopify/shopify.star
```diff
@@ -0,0 +1,554 @@
...
+    return [x for xs in xss for x in xs]
+
+def make_shopify_request(store_name, api_token, endpoint, parameters):
+    url = "https://{}.myshopify.com/admin/api/2022-04/{}.json".format(store_name, endpoint)
+    headers = {"Content-Type": "application/json", "X-Shopify-Access-Token": api_token}
+    response = http.get(url = url, params = parameters, headers = headers)
+    if response.status_code == 404:
+        return ERROR_404
+    elif response.status_code == 429:
+        return ERROR_429
```
matslina, line 543: A store with large enough order volumes (and high enough cycle speed) will hit the rate-limit frequently as orders accumulate over the day. Would it be good to not only report the rate-limit, but also respect it by backing off?

Since the rate limit is per store/token, and probably not all that many stores reach that kind of a daily order volume, I suppose it's not critical? WDYT?

apps/shopify/shopify.go
```diff
@@ -0,0 +1,25 @@
...
+	"tidbyt.dev/community/apps/manifest"
+)
+
+//go:embed shopify.star
+var source []byte
+
+// New creates a new instance of the Shopify applet.
+func New() manifest.Manifest {
+	return manifest.Manifest{
+		ID:          "shopify",
```
matslina, line 16: We have an internal Shopify app in the works, with a similar feature set. It's release is pending us setting up a public Shopify app, with their full authentication flow, so in the meantime I think this would be a wonderful addition! The internal app is however already hogging the `shopify` app ID, so this will need a different ID for this to get merged.

matslina (changes requested):
Hey @kcharwood. Thank you for the PR!

Apologies for taking so long to review; we've had a lot on our plates these last two weeks. I started looking at this more closely today, and will keep at it after the long weekend. In the meantime I wanted to share some early feedback.

kcharwood:
I'm on vacation as well. I'll address comments when I return.

apps/shopify/shopify.star
kcharwood: I chose to do this because I have seen some businesses that have financial source of truth in different than timezones than the person who might be located with a Tidbyt (my company included, I work CT but the company financials are based in ET).

"Location" fills a bit weird here to me.

apps/shopify/shopify.go
kcharwood: Changed the name to Shopify Chart

apps/shopify/shopify.star
kcharwood: @matslina is there a way to sleep or wait for an amount of time? I couldn't find that in the docs anywhere.

matslina: One approach is to write a cache record to mark backoff (e.g. `cache.set("rate-limited", "%d" % backoff, ttl_seconds=backoff)`), and check that before performing any API calls (e.g. `if cache.get("rate-limited"): handle_rate_limit()`).

apps/shopify/shopify.star
matslina: Yeah, I agree that Location is a bit weird for this use case, but while listing timezones in a dropdown works OK for users in the US, it gets really messy when we take the rest of the world into consideration. =\

kcharwood: Is there plans to support a Schema for Timezones explicitly? Probably wouldn't be hard to just have the whole list.

matslina: No immediate plan. If it was added, I suspect the best UX would still be to base it on location rather than on a full list. There're so many timezone designations out there that it'd have to be searchable, and location is a pretty convenient way to implement that search. Especially for non-technical users who've never seen a timezone designation before.

apps/shopify/shopify.star
matslina: So there's no way to pause an execution by e.g. calling a `sleep()` function, and there's no way to schedule executions. When rate-limit is hit, the current execution has to take some reasonable action (perhaps displaying a "rate-limit" error message.) The app will be executed again the future, as the Tidbyt device cycles through installed apps, offering an opportunity to get fresh data from the API.

The `rate-limited` cache record serves to stop these subsequent executions from hitting the API needlessly for a period of time, allowing the rate limit to be reset. Without it, the app could get into a state where it's calling the API, getting a 429, and then just trying again on the next execution. This can stop a rate-limit from being reset.

Setting an appropriate TTL on the `rate-limited` cache record makes it so that the cache eventually evicts the record. When that happens, new executions will attempt calling the API again, and hopefully stop receiving 429s.

But yeah I'm not sure if this is going to be needed. It'd have to be a pretty large number of daily orders for the rate limit to ever get triggered by Shopify's API.

kcharwood: Ya without a true sleep method to respect the wait time sent back by Shopify, I think I would have to really over-engineer something here to support the edge case, at least given what I know now.

apps/shopify/shopify.star
kcharwood: I'll update to support the Location schema.

apps/shopify/shopify.star
kcharwood: @matslina if you are good with no action here, then I think its ready.

matslina: Sure thing. Let's merge.

Just in case you're curious, there's an example of the rate-limit backoff over here: https://github.com/tidbyt/community/blob/main/apps/formula1/formula_1.star#L85

matslina (approved):

matslina:
Thanks @kcharwood!

kcharwood:
Thanks @matslina 

How long does it take for it to become available in the Tidbyt app?

PR #482, opened 2022-06-22
New App: Hayward Omnilogic Pool Temperature

kcharwood:
hayward_omnilogic

### DIFF
--- a/apps/haywardomnilogic/README.md
+++ b/apps/haywardomnilogic/README.md
@@ -3,3 +3,34 @@
 This app reads the latest pool temperature from the Hayward Omnilogic Controller, and displays it on the Tidbyt.
 
 ![](hayward_omnilogic.png)
+
+
+## Getting a Token
+
+Hayward does not provide an Oauth2 flow, and Tidbyt does not support storing username/password credentials, so you must manually get a token to use in the app.
+
+To obtain a token, execute the following command by replacing YOUR_USERNAME and YOUR_PASSWORD with your credentials to login to the app. 
+
+```
+curl -X "POST" "https://www.haywardomnilogic.com/MobileInterface/MobileInterface.ashx" \
+     -H 'Content-Type: text/xml' \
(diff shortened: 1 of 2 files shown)


matslina (approved):
Love the pixel art! Happy to merge this as is.

Wanted to check though if it'd be possible to default to a more illustrative graphic for preview purposes? E.g., if username isn't provided, then render the full graphic but with a placeholder temperature and pool name? Would make it easier for a user to understand what the app has to offer.

matslina (changes requested):
Bad news... =(

After some internal discussion, we decided that we're going to have to reject apps that have users enter username and password. We to accept authentication tokens, as these can generally be revoked if need be, but username/password generally do not have this property and therefore shouldn't be persisted in our database.

I'm guessing there's no oauth2 authentication flow provided by the pool sensor service?

kcharwood:
@matslina There is no oauth2 service provided. There is an XML endpoint to grab a token as you can see in the code.

I guess you don't have a way for a dynamic config option to run some custom code to take some user entered parameters and then save the result as a token?

matslina:
@kcharwood There is a (currently undocumented, possibly unsupported for community apps) dynamic configuration field that generates a new schema field by calling a handler whenever some other schema variable (A) is changed by user. Unfortunately, A is also persisted as a configuration value, so since A needs to hold username&password in this use case, we're back at the original problem...

kcharwood:
I have a few other ideas for apps (SleepNumber IQ and Sutro), but both would need username/password i think...

skola28:
> I have a few other ideas for apps (SleepNumber IQ and Sutro), but both would need username/password i think...

Might be a kludge, but you can have the user manually enter (copy/paste) the token into a schema field.

kcharwood:
I took @skola28's idea and changed out to use a token generated outside the system. I updated the README to include instructions on how to generate a token.

matslina (approved):

PR #485, opened 2022-06-24
formula_1: respect API rate limit

matslina:


matslina:
@AmillionAirs PTAL. This should make the app back off completely for 25h when the API returns 403, indicating rate limit in effect.

PR #488, opened 2022-06-25
subreddit: fix parsing of location config

fytpet:
Adds JSON decoding of location config that was missing.
Should fix subreddit app that is currently not displaying on community.


betterengineering (approved):
ty!!

PR #493, opened 2022-06-29
Added the OGS Games Viewer App

Hendeca:
ogs_games_viewer

I am an avid go player (the Chinese abstract strategy game) and wanted to make an app that would artfully display a selection of my in-progress games. This app uses an OGS ([online go server](https://online-go.com)) Username to pull in-progress go games from the OGS API and display them, along with the name of the opponent and what color they're playing.

If there are improvements to the app, I'd be happy to make them! Please let me know if there's anything you need me to change. Thank you and I hope my app can be added!

EDIT: Updated the .gif preview using the new code

### DIFF
--- a/apps/apps.go
+++ b/apps/apps.go
@@ -85,6 +85,7 @@ import (
 	"tidbyt.dev/community/apps/noaabuoy"
 	"tidbyt.dev/community/apps/nyancat"
 	"tidbyt.dev/community/apps/nycbus"
+	"tidbyt.dev/community/apps/ogsgamesviewer"
 	"tidbyt.dev/community/apps/ohhighwaysigns"
 	"tidbyt.dev/community/apps/pagerduty"
 	"tidbyt.dev/community/apps/pathtrainschedule"
@@ -214,6 +215,7 @@ func GetManifests() []manifest.Manifest {
 		noaabuoy.New(),
 		nyancat.New(),
 		nycbus.New(),
+		ogsgamesviewer.New(),
 		ohhighwaysigns.New(),
 		pagerduty.New(),
 		pathtrainschedule.New(),
--- a/apps/ogsgamesviewer/ogs_games_viewer.star
+++ b/apps/ogsgamesviewer/ogs_games_viewer.star
@@ -279,7 +279,21 @@ def main(config):
             child = not_found_graphics,
         )
 
-    PLAYER_ID = get_player_id_by_username(USERNAME)
+    # If a new Username has been set in the options, reset the games cache
+    # and pull a new player_id
+    if cached_user == None:
+        cached_user = cache.set("username", USERNAME, ttl_seconds = 240)
+        PLAYER_ID = get_player_id_by_username(USERNAME)
+        cached_player_id = cache.set("player_id", str(PLAYER_ID), ttl_seconds=240 )
+        games_cache = None
+    elif USERNAME != cached_user:
+        cached_user = cache.set("username", USERNAME, ttl_seconds = 240)
+        PLAYER_ID = get_player_id_by_username(USERNAME)
+        cached_player_id = cache.set("player_id", str(PLAYER_ID), ttl_seconds=240 )
+        games_cache = None
+    else:
+        PLAYER_ID = cache.get("player_id")
+        games_cache = cache.get("games")
 
     # Get the player ID and game details from the API
     # If the API didn't return a player ID, show a message
@@ -288,16 +302,6 @@ def main(config):
         return render.Root(
             child = not_found_graphics,
         )
-
-    # If a new Username has been set in the options, reset the games cache
-    if cached_user == None:
-        cached_user = cache.set("username", USERNAME, ttl_seconds = 240)
-        games_cache = None
-    elif USERNAME != cached_user:
-        cached_user = cache.set("username", USERNAME, ttl_seconds = 240)
-        games_cache = None
-    else:
-        games_cache = cache.get("games")
 
     # If there is an existing games cache, pull info from the cache
     # otherwise, pull in new games and cache them
--- a/apps/ogsgamesviewer/ogs_games_viewer.star
+++ b/apps/ogsgamesviewer/ogs_games_viewer.star
@@ -11,22 +11,13 @@ load("http.star", "http")
 load("render.star", "render")
 load("schema.star", "schema")
 
+# Constants
 NUM_OF_GAMES = 5
 FRAMES = 50
+SIZES = [9, 13, 19]
 
-# Sort moves by axis instead of by move number
-def sort_moves(moves):
-    moves_dict = {}
-    turn_dict = {
-        "b": "w",
-        "w": "b",
-    }
-    turn = "b"
-    for move in moves:
-        move_key = "{},{}".format(int(move[0] + 1), int(move[1] + 1))
-        moves_dict[move_key] = turn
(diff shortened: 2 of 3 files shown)


Hendeca:

Hendeca:
One note: I wasn't sure how I could properly cache this app since there some complex data that is pulled in and would probably need to be stored as more than a string. If you know of a good way to cache this information, please let me know.

I also realized I left a `print()` command in the code that was not intended to be there.

rs7q5:
You can store them as a 'json' and cache `json.encode(x)` and then `json.decode` the cached data. So I. The example above x could be a dict or some other type too.

Hendeca:
@rs7q5 Ah good idea, thank you! I'm somewhat new to doing a pull request like this, what is the best way to edit my code and re-submit it? Is it best to just close this request and start another, or is there a better way?

rs7q5:
You can just recommit your changes to your branch on your repo and they will update here when you push to your main branch. So no need to close this pull request. 

In the future though for your own sanity you may want to open a PR on another branch other than main so it's easier if you have multiple PRs and more apps your fixing at a time. I learned that one the hard way.

Hendeca:
@rs7q5 Thanks, that's good to know! I must sleep, but I will try adding caching tomorrow using the method you suggested!

rs7q5:
No problem! A lot of the apps do this method anyway for some of the complex data, so hopefully you shouldn't have too much trouble!

Hendeca:
@rs7q5 I have one more question for you regarding caching: will the cache be cleared if one of the schema options is changed? I am using a text field in the schema to pull from the API, so I'd like the cache to clear if the value is changed. Do I need to do this manually, or will the Tidbyt automatically clear the cache when a new value is entered? Thanks again for being so helpful!

rs7q5:
The cache should be rechecked when you change a config setting, Now that being said if you still want it to reget the cached data if a user changes a certain config input I think you'd have to do it manually by caching the old value (which would have to be a greater amount of time for the one that you want to reset otherwise it wouldn't line up). 

Having it reset when the value is changed though would cause it to constantly to try and do `http.get` and you could run into rate limits. (make sure to  have a unique cache for each user. so the key may be `cache_key=OGS_games_%s` % username` . 


Also I'm assuming by value you mean the username correct?

Hendeca:
@rs7q5 Yes I was talking about the Username! Ok, that makes sense, though I _think_ the way I'm doing it makes sense? Basically if the username cache is empty, I set it to the new username and if the username cache is not equal to what is in the schema options, I set the cache to the new username and pull new data. If the username is the same as what is cached, then the games cache is used. I just pushed a version with these changes. I do like the idea of using a custom cache key, though I'm not sure if it's necessary here. I may be missing something though.

I added cacheing for the API call to get the player_id based on the username as well. If you see anything wrong with the method of cacheing let me know. Thanks again for the help

rs7q5:
Did you try using `pixlet serve` and seeing if the cache stuff works correctly? That would be the best way to check. that it is working as expected. 

Although you will  need to have unique cache keys because keys are shared across devices. right now everyone would see the same data on a tidbyt (which in this case may or may not be your data since the username could be different). 

you could just add the username to each cache key you currently have and that should be sufficient as I'm assuming there has to be unique usernames.

Hendeca:
@rs7q5 Yes I did try it out and it seemed to be working well. However I didn't know that about everyone seeing the same cache keys between devices! Glad I asked because that would have caused all kinds of problems. I'll make the cache keys unique to the username which should prevent any conflicts. Thanks again, this is all so helpful haha

rs7q5:
No problem! Luckily they are easy things to fix!

Hendeca:
Ok, I think this version is actually ready to go. I got in the cacheing with username-specific keys, tested the cacheing, cleaned up the code, and fixed a bug. Things seem to be working smoothly. I think this is now ready for review.

Hendeca:
Sorry to keep triggering the false alarm. Found one more thing I need to fix tomorrow: I realized the game detail data is not cached so there are still unnecessary API calls. Will fix tomorrow and push

matslina:
Go player here. Very excited to get this app up on my Tidbyt! =)

How about displaying a historical game by default, i.e. when `not username`? This would produce a nice, explanatory preview in the list of apps, making it easier for users to understand what the app does.

Would it make sense to raise the cache TTL for the username-to-id records? I'm guessing they change very infrequently? Could cut request rate to OGS down by 25%.

Hendeca:
> 

Both great ideas! I love the idea of having some kind of historical game before a username has been entered. I could also increase the TTL for the boolean username cache too. In addition to that, I still need to add the cache for the game data itself, which will have a much shorter TTL so that you can see new moves that are played. Thanks for these recommendations, I'll get on that!

Hendeca:
@matslina I implemented the changes you suggested. I only increased username and player_id cache TTLs to 1 hour, as the username can be freely changed. I didn't want to leave the player_id cache in for too long since that player_id will no longer correspond to the correct cache key once the username has been changed. If you think I should increase it, let me know. I'm basically trying to avoid a situation where one user changes their username and another adopts the same username sometime after that. Probably a pretty unlikely edge case, but if the username was something popular (I'm thinking "Sai" for example) then there could be a conflict.

I tested the cacheing and it seems to be working correctly! At this point there is only one small thing that I'd like to fix, but I'm not sure how. The marquee scrolling doesn't seem to work very well when coupled with the Animation widget. This is mentioned in the Widget documentation (that animated children in an animation can cause problems) so I'm not sure if there's a fix for this right now. If someone knows of a fix, please tell me!

Unless anyone has other suggestions or finds other issues, I think this is ready to be added.

A potential future feature to add:
* Make the most recent move blink

rs7q5:
> I tested the cacheing and it seems to be working correctly! At this point there is only one small thing that I'd like to fix, but I'm not sure how. The marquee scrolling doesn't seem to work very well when coupled with the Animation widget. This is mentioned in the Widget documentation (that animated children in an animation can cause problems) so I'm not sure if there's a fix for this right now. If someone knows of a fix, please tell me!

Have you looked into the `animated` module?

rs7q5:
If you go to `render.Sequence` in this first link, there's a link on the word `animation` which is the second link. 
https://github.com/tidbyt/pixlet/blob/main/docs/widgets.md

https://github.com/tidbyt/pixlet/blob/main/docs/animation.md

Hendeca:
> If you go to `render.Sequence` in this first link, there's a link on the word `animation` which is the second link. 

Oh I thought there was a separate module called `animated` instead of `animation`. I am using the animation widget and have read through the page you linked, but I haven't found anything to fix the issue this far!

matslina:
@Hendeca We'd be happy to merge as is if you'd like. Can always cut new PRs with improvements later on. LMK.

Hendeca:
> `animation.star` has different capabilities than `render.animation`.
> 
> you also have to use `load("animation.star","animation")` to use it.

Ohhh, I see, I got confused there. I will look into this! Thanks for the recommendation.

> @Hendeca We'd be happy to merge as is if you'd like. Can always cut new PRs with improvements later on. LMK.

That would be great, I think it's good enough as-is! Thank you :)

I will see if I can improve the marquee animation. Most of the time it's not much of an issue, but it would be nice if each time a new board was shown the marquee scrolled. I'll work on this and add a later PR if I am able to fix it

PR #494, opened 2022-06-29
New app: Tindie sales

joeycastillo:
This app displays sales stats for sellers on Tindie. Admittedly a bit niche, but useful for folks who want to know when they need to ship out some boxes. 

To use, the app user must set their Tindie username and API key in the Tidbyt app. They can optionally set a “Show Sales Since” date to only show i.e. year-to-date sales or sales after an arbitrary cutoff date. 


joeycastillo:

PR #497, opened 2022-06-30
SF Next Muni UI improvements

juniormonkey:
1. Include the stop ID in the schema Option display, to distinguish between inbound and outbound stops with otherwise the same name; and
2. Add a new display format to show slightly less of the destination, and two instead of one next predictions:
sf_next_muni


matslina:
Hey @juniormonkey. Looks like this could use a rebase now that the other PR is merged.

betterengineering (approved):

PR #498, opened 2022-06-30
New App: Sound Transit

bigfootjon:
This app shows upcoming stop times for up to two different stations in Sound Transit's Link Light Rail network. Each "station" in the network is actually 2 different stations in the APO: one station per direction of travel. For most of link that's North / South.

So while this app can display 2 logical stations it can realistically only show one actual station, but the schema allows the user to configure 2 completely different stations if they wish (or just one station)

I modeled most of the design off other transit apps in this repo, and tried to write idiomatic code but I think it came out kinda ugly.

image


bigfootjon:

PR #499, opened 2022-06-30
USA Severe Weather Alerts

aschechter88:
USA Severe Weather Alerts. Pulled from the US National Weather Service. Uses location from schema. 5 minute caching on lat/long point. National Weather Service API has no API key requirement.

## No alerts here!
no-alerts

## Moderate and Severe/Extreme level alerts!
alerts


btjones:
@aschechter88 it looks like the hurricane tracker has an option for skipping the app when there are no hurricanes.

Appears that app calls `return []` to skip the app instead of calling `return render.Root`.

https://github.com/tidbyt/community/blob/852dac7f7136830ab07df883ce36c7a7f92f6462/apps/hurricanetracker/hurricanetracker.star#L99

PR #504, opened 2022-07-02
Severe WX Alerts USA

aschechter88:
Display severe weather alerts issued by the US National Weather Service.

## No Alerts
no-alerts

## Alerts
alerts

### DIFF
--- a/apps/severewxalertsusa/severewxalertsusa.star
+++ b/apps/severewxalertsusa/severewxalertsusa.star
@@ -1,0 +1,342 @@
+"""
+Applet: SevereWxAlertsUSA
+Summary: Display US Severe Wx Alerts
+Description: Show Severe Weather Alerts in your location issued by the US National Weather Service.
+Author: aschechter88
+"""
+
+load("render.star", "render")
+load("schema.star", "schema")
+load("http.star", "http")
+load("encoding/json.star", "json")
+load("encoding/base64.star", "base64")
+load("time.star", "time")
+load("cache.star", "cache")
+
+DEFAULT_LOCATION = """
+{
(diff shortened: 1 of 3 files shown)


apps/severewxalertsusa/severewxalertsusa.star
```diff
@@ -0,0 +1,346 @@
...
+
+## run the main applications
+def main(config):
+    jsonLocation = json.decode(config.str("location") or DEFAULT_LOCATION)  ## set the location from the schema data or use the default
+
```
matslina, line 30: Would you mind adding some code to truncate latitude and longitude? This would help protect against leaking users' exact location to the API provider. As a secondary benefit, it also improves cache rate a bit in densely populated areas.

I'm thinking 2 decimal points would be sufficient, based on [this discussion](https://gis.stackexchange.com/questions/8650/measuring-accuracy-of-latitude-and-longitude).

You can find some examples over here: https://github.com/tidbyt/community/search?q=truncate

matslina: Great, thank you!

There is one small issue though. Shorter inputs do not appear to be truncated correctly. E.g., `truncated_location_digits("0.313") == "0.313"`, while `truncated_location_digits("0.3130") == "0.3150"`.

I think you can simplify the function and get the desired effect like this:
```
def truncated_location_digits(x):
    return str(int(math.round(float(x) * 200)) / 200)
```

betterengineering: To unblock, I am going to merge and update main with the suggestions

PR #510, opened 2022-07-05
New App: Shuffle Images

rs7q5:
Randomly display an image from a list of up to 20 user-defined images.

Still trying to figure out how to get them to display linearly (i.e. order defined), but can't figure out a way to cache an index that wouldn't get constantly overwritten.
A unique index key would be fine across devices since the images aren't cached, but couldn't figure out a good way to do that at the moment.

This app also may help reduce the number of apps listed on the device if you have a lot of photos.


rs7q5:
Note I just noticed the burger of the day app was also modified here, but I didn't change anything in that file. It doesn't look like any content was changed, but I just checked on the main branch and looks like the two lines that got modified were done so when I ran `make lint`

cc @kstardesigns for reference.

kstardesigns:
> Note I just noticed the burger of the day app was also modified here, but I didn't change anything in that file. It doesn't look like any content was changed, but I just checked on the main branch and looks like the two lines that got modified were done so when I ran `make lint`
> 
> cc @kstardesigns for reference.

@betterengineering @rohansingh @matslina Please merge in this change, it should fix an issue in the burger app where certain letters like ñ aren’t appearing correctly

rs7q5:
Also fyi. This should probably go under home,  entertain or create when trying to add it to the tidbyt. 

Would be nice if when we used `make app` we could add pre selected categories so when you made it live the app would appear in these lists, granted the list could get long that way, but only worth it if it makes it easier on you guys.

rs7q5:
FYI I think I modified apps.go the wrong way when trying to resolve the conflict with the shopify app that was added as it wasn't in the apps.go and so this commit would now delete that, sorry about that. 

I can't update my branch as github currently says it's up to date, so sorry about that.

apps/apps.go
```diff
@@ -102,7 +102,7 @@ import (
 	"tidbyt.dev/community/apps/redditrplace"
 	"tidbyt.dev/community/apps/sbbtimetable"
 	"tidbyt.dev/community/apps/sfnextmuni"
-	"tidbyt.dev/community/apps/shopifychart"
+	"tidbyt.dev/community/apps/shuffleimages"
```
matslina, line 105:
```suggestion
	"tidbyt.dev/community/apps/shopifychart"
	"tidbyt.dev/community/apps/shuffleimages"
```

PR #514, opened 2022-07-08
Pokedex: Swap Random function

rs7q5:
Swapped the random function with the new `random.number` module available in pixlet. Not sure if this will help with the app constantly repeating a lot of the starter pokemon in the evening (at least in my experience).

However this should ideally be tested before actually making it live. I also only commented out the original random module stuff just in case that wasn't the issue. FWIW locally it seems to randomize, but doesn't always on the tidbyt it seems.

cc @mackorone as they are the original author and there may be something else different that I'm missing?

### DIFF
--- a/apps/pokedex/pokedex.star
+++ b/apps/pokedex/pokedex.star
@@ -11,7 +11,8 @@ load("encoding/json.star", "json")
 load("http.star", "http")
 load("render.star", "render")
 load("schema.star", "schema")
-load("time.star", "time")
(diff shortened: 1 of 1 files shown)


rs7q5:
@mackorone should I go ahead and delete the commented out old random module you had in there originally? I should be able to remove the use of the `time` module too or would you like me to go ahead and leave it commented out?

mackorone:
@rs7q5 you can delete all of the old code - it's not necessary anymore.

betterengineering (approved):

PR #518, opened 2022-07-11
New App: Fantasy Name Generator

trumpetbear:
A new app to randomly create a fantasy RPG character
screenshot
.


matslina:
A little bit worried the sheer size of the script will be a problem in production. I'll merge and test. Hopefully it works fine.

trumpetbear:
Let me know if the size is an issue. There are a bunch of background images and we could easily pare those down.

PR #521, opened 2022-07-13
Adding the "Display My IP" app

kuzmik:
A really simple app that will fetch the public IP of the network the Tidbyt is connected to and display it. Partially I did this to learn Starlark, and partially because my IP changes so often I find myself using a bash alias to print it, and thought this would be a decent use of the Tidbyt.


kuzmik:

PR #523, opened 2022-07-13
Introduce configuration option for 24h time format

hloeding:
Introduced a new configuration option, so the user may select the time format to be used:

- AM/PM
- 24h

Default is AM/PM, as before


hloeding:

PR #539, opened 2022-07-19
Fixed layout issue for Sunrise/Sunset

kcharwood:
Currently, the app has a broken layout:

# Currently:
image

My fixes here result in it now working:

# Updated:
image

@alan-fleming I'm not sure if my changes would cause any conflicts here as I don't know why you had it set up with specific padding originally.


alan-fleming:
I'd prefer not to do it this way. It doesn't answer the issue that times with 4 digits overflow on the other side of the screen, which was the real issue I was attempting to fix.  Working on this right now here by altering the image.

PR #541, opened 2022-07-19
Update sunrise_sunset.star

alan-fleming:
Reduce size of images by 1 column and remove contributed pad fudging code, squish a couple of bugs others pointed out too.


matslina:
Thanks @alan-fleming and @kcharwood!

PR #544, opened 2022-07-19
todoist: reduce cache ttl to 60s

matslina:


matslina:
@zephyern I think we can comfortably do 1 req per minute and user without exceeding the API rate limit. It's rewarding to see your task count go down as you get stuff done. =)

PR #553, opened 2022-07-20
Todoist config to allow hiding when no tasks

JukeboxRhino:
I like this app, but I would prefer an option to not have the app show when I have no tasks that I need to worry about.

I'd like to add a config toggle which would allow the app to not render when the number of tasks fetched is zero.

It defaults to the current behavior so that only users who change this config would opt-in to not rendering the app in this scenario.

screenshot

### DIFF
--- a/apps/todoist/todoist.star
+++ b/apps/todoist/todoist.star
@@ -47,6 +47,9 @@ l/d/WBrBhOVz/g1OHB6agdqc6AAAAABJRU5ErkJggg==
 
 DEFAULT_NAME = "Todoist"
 DEFAULT_FILTER = "today | overdue"
+DEFAULT_SHOW_IF_EMPTY = True
+
+NO_TASKS_CONTENT = "No Tasks :)"
 
 TODOIST_URL = "https://api.todoist.com/rest/v1/tasks"
 
@@ -76,11 +79,15 @@ def main(config):
             if num_tasks == -1:
                 content = "Error"
             elif num_tasks == 0:
-                content = "No Tasks :)"
+                content = NO_TASKS_CONTENT
             else:
                 content = humanize.plural(int(num_tasks), "Task")
 
             cache.set(cache_key, content, ttl_seconds = 60)
+
+        if (content == NO_TASKS_CONTENT and not config.bool("show")):
+            # Don't display the app in the user's rotation
+            return []
 
     else:
         # This is used to display the app preview image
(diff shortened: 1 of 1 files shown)


JukeboxRhino:
@zephyern I'd love it if you gave this PR your blessing - thanks for making this app

zephyern:
Consider it blessed! Thanks!

My only feedback would be to tweak the wording of the description. Maybe:

"Show this app when there are no tasks."

The current wording may be confusing as to which value causes the app to show and which to hide.

Thanks again!

zephyern:
BTW how did you hear of this app? Did you just happen to find it in the code, or did someone mention it somewhere?

JukeboxRhino:
> BTW how did you hear of this app? Did you just happen to find it in the code, or did someone mention it somewhere?

It was announced in the Tidbyt discord!

JukeboxRhino:
> My only feedback would be to tweak the wording of the description...
Changed, I agree it was a little wordy

zephyern:
I just checked and on the iOS app it doesn't show the description. Maybe we should change "Show" to "Show When No Tasks" or something?

Otherwise it'll just say "Show" haha

image

JukeboxRhino:
Weird that it never shows the description...but I will change it 👍

zephyern:
Awesome! Not sure if you saw, but I left a comment on the code itself with regards to caching. Otherwise looks great. Thanks again!

JukeboxRhino:
I'm normally a GitLab guy but I don't see any comments @zephyern

zephyern:
I think the comment is here:

https://github.com/tidbyt/community/pull/553#discussion_r926082061

JukeboxRhino:
There, should be fixed wrt caching now

zephyern:
It's close, but actually if the data is in the cache then it skips your logic. :) 

I think you need to move that logic down a bit so that it hits even if the data is in the cache.

Does that make sense?

JukeboxRhino:
Should be good now...thanks for catching that!

JukeboxRhino:
For some reason I can't seem to see your review unless I hover over that link to it, and then all I can see is the first sentence or so. Can't tell if I'm just stupid or don't have permission in this repo or something lol

zephyern:
Changes look great now. :) Not sure why my comments aren't coming through, but you've addressed them all. Hope you enjoy the app!

matslina:
ready to merge?

PR #558, opened 2022-07-22
New App: Twitch for Tidbyt

drudge:
This PR adds a new Twitch applet that allows a user to show follower or subscriber count for a Twitch profile.

twitch

Feel free to replace with your own client ID/secrets.


betterengineering (approved):
This looks great! Thanks @drudge! Will follow up with our client/secrets

PR #561, opened 2022-07-24
Fix London apps

dinosaursrarr:
Both of these apps disappointingly failed to work for real on my actual Tidbyt. When trying to configure them on my phone, the location look up doesn't actually provide a list of stops/stations to choose from.

See descriptions on each of the commits for details of why I think this should work. I'd appreciate a second eye. I'm not honestly 100% confident in the fixes, given that I thought they were going to work before. It's a difficult dev workflow when you have to wait for the actual release to be able to test it, and it's not obvious how the handlers wrap intermediate outputs.

### DIFF
--- a/apps/londonbusstop/london_bus_stop.star
+++ b/apps/londonbusstop/london_bus_stop.star
@@ -14,7 +14,7 @@ load("schema.star", "schema")
 load("secret.star", "secret")
 
 DEFAULT_STOP_ID = "490020255S"
-STOP_URL = "https://api.tfl.gov.uk/StopPoint/%s"
+STOP_URL = "https://api.tfl.gov.uk/StopPoint"
 ARRIVALS_URL = "https://api.tfl.gov.uk/StopPoint/%s/Arrivals"
 
 # Allows 500 queries per minute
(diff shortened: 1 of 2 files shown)


rs7q5:
Starlark does support the look up `k in d` you should be able to use either just check the key by using `if k in dict.keys()` or use `dict.get`

Also is it always just using the default location? Are you json.decoding the cache data or is the issue simply with the display of optional locations?

rs7q5:
Also is there a reason a new app is included in this commit?

dinosaursrarr:
> Also is there a reason a new app is included in this commit?

Because git is a confusing machine for creating unwanted state.

dinosaursrarr:
> Starlark does support the look up `k in d`

The interpreter didn't complain about `if k in d` but it didn't give the correct answer. Switching it to `d.get(k)` did. I suspect the interpreter was seeing `k in d` as trying to start a look (as in `for k in d`) and returning a truthy value because there was at least one element.

> Also is it always just using the default location?

Currently it's bypassing the schema altogether and using a hard coded stop/station. That's because the handler was broken so you couldn't choose anything else. You could put in a location, but then it would always give an empty results screen. There's literally nothing you could tap on with your finger.

rs7q5:
> The interpreter didn't complain about `if k in d` but it didn't give the correct answer.

weird. Isn't supposed to return a truth value though lol. like if it's in the array? or were you wanting it to return the value? Obviously in that case then `dict.get` was what you were looking for. 

> Currently it's bypassing the schema altogether and using a hard coded stop/station.

Ah so that's what you were fixing. got it.  As long as you are returning a list always you should be fine even if there are no options (which it looks like you are).

matslina:
Hello. Something like `k in d` will evaluate to `True` iff the dictionary `d` contains a key `k`, no matter what the associated value `d[k]` is. If e.g. `d[k]` is `None`, `0`, `False` or `""`, these will all be treated as `False` in an if statement. Doing `d.get(k)` means you get `d[k]` if `k in d`, and `None` otherwise. I'm guessing in these cases the key was present, but the associated value evaluated was "not truthy" in the if-statement?

Either way, this seems to work so off we go with the merge.

PR #574, opened 2022-08-02
Split scores up to 8 app instances

cmarkham20:
Added the ability for users to install the app up to 8 times and have the scores split between each instance.


matslina (changes requested):
Hey there. Sounds like a neat change, but it looks like this PR has gone a bit haywire?
image
image
=)

cmarkham20 (commented):
While yes, there are quite a few changes, the are mainly the same changes across the 6 "scores" apps, and there are 3 new Standings apps. Of these apps work well on my testing environments. I'm not 100% sure how to open separate PRs for each but will look to do so in the future.

cmarkham20:
@matslina Let me know what I can do to push this through... Again, there are a lot of apps within this PR hence amount of files.

matslina (approved):

PR #577, opened 2022-08-04
strava: Don't hammer the API when we get HTTP 429

rohansingh:
When we get an HTTP 429 (rate limit exceeded), back off for a few
seconds before trying again.


rohansingh:
Current behavior is not great, when we get errors we just retry continually.

image

PR #581, opened 2022-08-05
App: PUBG Stats (Fixes/Features/Readme)

joes-io:
Fixes
- Fixed bug that prevented app from working with non-Steam platforms.
- Data caching was incorrectly using player name or ID alone as keys. Now appends platform as well should names or IDs be duplicated across platforms.
- Additional error handling for 401 response.
- Hopefully fixed build error.

Enhancements
- All lifetime player stats provided by PUBG's Developer API are now selectable to be displayed on Tidbyt!
- Stats that have longer display labels now replicate on a marquee.
- README and better GIF added.

Everything tested well in dev environment, but API key decryption remains untested until app shows up in Tidbyt library. I'm very open to feedback if anyone sees anything that could be done more efficiently or elegantly.


joes-io:
@kconnell98 I saw you had a similar build-and-test failure on CLT Lightrail and were able to fix it. I've tried a couple of things for my own app, but I don't want to keep hitting the community repo with pull requests trying to troubleshoot. Do you have any insights from being able to get yours to pass? Everything was working great in dev environment and lint passes just fine.

kconnell98:
Hey @joes-io, I actually didn't fix the build-and-test, but instead it seems that @betterengineering did for me instead. Looking at your check, it seemed it failed because of Icons in your .star file. @betterengineering went into my file and changed my schema icon as the recent changed the naming convention for some.

In your case, I see you used 3 Icons; user, gamepad, and chart-line

The first two look good, but the last one seems to have changed as you can see here [link](https://github.com/tidbyt/community/blob/main/apps/icons.go#:~:text=%22chartGantt%22%2C-,%22chartLine%22%3A,-%22chartLine%22%2C).

My guess is that if you fix "chart-line", you check should work!

joes-io:
My hero @kconnell98! 🙌 That was it, thank you so much!

I did look at the log last night, but totally missed the `icons_test.go:36: app 'PUBG Stats' contains unknown icon: 'chart-line'`. Clearly see it now.

I even knew about changing the Font Awesome names from dashed to camel case, and then on a whim decided to change icons before doing the pull. Totally slipped my mind since "chart-line" does work on the local page served up by Pixlet.

PR #582, opened 2022-08-06
Sports Scores: Add sports and other adjustments

rs7q5:
Added some college sports, basically should be pulling from teams in top 25 for college sports
Also added EPL for soccer

Fixed a potential bug in highlighting winners in NHL games

Rearranged the schema since can only have one generated field, etc.

### DIFF
--- a/apps/sportsscores/sports_scores.star
+++ b/apps/sportsscores/sports_scores.star
@@ -6,7 +6,7 @@ Author: rs7q5
 """
 #sports_scores.star
 #Created 20220220 RIS
-#Last Modified 20220806 RIS
+#Last Modified 20220811 RIS
 
 load("render.star", "render")
 load("http.star", "http")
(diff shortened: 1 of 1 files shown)


matslina:
Hey. Having a bit of a hard time following all that's going on in this PR, but it looks like this will break all existing installations of the app, no? That's not ideal, and we unfortunately don't really have a mechanism for migrating existing installations.

To be more precise, users who've installed the currently live version of the app will have a record saying e.g. `sport=MLB`. With this change, I think those installations will just flat out fail until the user reconfigures the app. It'd be great if we could bring in this change without having that side effect.

rs7q5:
So previously I had the highlight team text entry as a `schema.Generated`, but you can only have one schema.Generated function and it made more sense to do the league instead of the text stuff. The other schema fields below it had to move into the `more_options` function because the `schema.Generated` field automatically gets put at the end of the schema fields (makes sense for sanity, however it makes no sense for my field to be at the bottom there).

Long story short, I shuffled stuff around so there is a sport and league option since there are so many soccer and basketball options in the app now. 

As for the issue you have raised, I can test locally if you merge and gave me access, however I think the latest commit should fix your concern?

rs7q5:
I guess I also could have added a frame to tell users to check existing configs too, but the current fix is probably a better solution.

PR #585, opened 2022-08-07
That would be an ecumenical matter

dinosaursrarr:
Christmas is only 25 December in certain parts of the world.

This extends @austinrfnd 's app to support the other possible dates to be more inclusive.


rs7q5:
December?

matslina:
This seems like pretty good change to me. Any comments @austinrfnd ?

PR #588, opened 2022-08-10
arcade-classics: Remove centipede from the options.

betterengineering:
# Overview
Arcade Classics is our 10th most popular app on the platform 🎉 . People really love it, as do we. The bad news is Centipede consuming a large amount of CPU across our server fleet to render apps. n the short term, I am removing Centipede from the options which will do two things:
- It will prevent new installations of Centipede from being installed
- It will remove Centipede from the random option

All existing Centipede installations will continue to work.

cc @sotteson1 

# Context
Only 3% of arcade classics have selected Centipede, however, 45% of arcade classics have selected Random which includes Centipede 1/3 of the time. We have caching internally to not render webp graphics if the tree of widgets is the same - which comes in handy with apps like weather that will look the same in a given region. However, because arcade classics has so much randomization, it will never hit this cache.

# Next Steps
@sotteson1 we're happy to add centipede back in if we can get it to render in under 500ms:
```
time pixlet render apps/arcadeclassics/arcade_classics.star animation=centipede
pixlet render apps/arcadeclassics/arcade_classics.star animation=centipede  4.12s user 0.10s system 342% cpu 1.231 total
```

We should also consider a limited set of seed values for the random number generator to be able to make use of our internal cache.

# Changes
- [arcade-classics: Remove centipede from the options.](https://github.com/tidbyt/community/commit/c95188e03baa52881358a236659eb32f2aa52e8f)
    - Centipede takes several seconds to render. Due to the randomization, it also skirts all of our internal caching mechanisms, which leads centipede to take a very large amount of the CPUs on our app-server fleet.


betterengineering:
Screen Shot 2022-08-10 at 3 19 12 PM

Centipede was accounting for 25% of our CPU usage 🙀 .

PR #590, opened 2022-08-13
New App: JSON Status for Tidbyt

wojciechka:
Displays JSON Status for one or more items by fetching a provided URL.

Retrieves a JSON with progress status that may return up to 4 elements, showing status of each element on the Tidbyt display.


wojciechka:
This currently works if the URL is set to https://raw.githubusercontent.com/wojciechka/community/feat/jsonstatus/apps/jsonstatus/example-status.json but should work with the default URL once the PR is merged.

wojciechka:

matslina:
Hey @wojciechka. Thanks for the contribution!

As much as I like the idea (and the code, and the design), I'm afraid we currently can't merge apps that make http calls to fully user-provided URLs. This policy is in place to keep our backend from being used to spam or otherwise cause trouble. I hope we can come up with a reasonable way around this in the future, but we're not there right now.

wojciechka:
> Hey @wojciechka. Thanks for the contribution!
> 
> As much as I like the idea (and the code, and the design), I'm afraid we currently can't merge apps that make http calls to fully user-provided URLs. This policy is in place to keep our backend from being used to spam or otherwise cause trouble. I hope we can come up with a reasonable way around this in the future, but we're not there right now.

Oh, I did not think that you're running this on the backend, not the device, I assumed that which is why I thought the URL has to be publicly accessible, but that would make sense that it's a security issue on your end.

What if I would limit this to specific website or websites. For example GitHub gists (just an example, here the problem would probably be API limits without a GitHub token) or a pastebin-like service that would allow API-based updates? I'd have to research which service(s) allow that, but would that even be a viable alternative?

This way I and perhaps others could push status / updates using an API and the Tidbyt app would just fetch the current status, hopefully without authentication.

One other alternative is for me to create a simple service for this purpose, perhaps with simple CRUD to add/remove progress indicators, but I am not sure I want to build an entire service, with auth and everything, just for the purpose of seeing status on Tidbyt.

matslina:
Hey! Yeah, limiting it to fetching data from certain sites where we can reason about request rates and limits would be a path forward.

wojciechka:
> Hey! Yeah, limiting it to fetching data from certain sites where we can reason about request rates and limits would be a path forward.

Ok. I think it's better I ask before I spend more time on this.

Would limiting it to MongoDB Atlas work?

It has a free plan, [data API](https://www.mongodb.com/docs/atlas/app-services/data-api/) that is rest and would make it easy to fetch the data.

Then the schema for the widget would be to ask for token (same as [todoistnext](https://github.com/tidbyt/community/tree/main/apps/todoistnext) is doing it) and any other parameters needed to make the calls.

I could also add snippets on how to initialize a database to README.

Would that work?

PR #593, opened 2022-08-16
Create the giphy app

DigitallyBorn:
Allows users to display random gifs based on a configurable search query.
giphy


DigitallyBorn:

apps/giphy/giphy.star
```diff
@@ -0,0 +1,132 @@
...
+load("encoding/base64.star", "base64")
+load("encoding/json.star", "json")
+load("http.star", "http")
+load("random.star", "random")
+load("render.star", "render")
+load("schema.star", "schema")
+load("secret.star", "secret")
+
+SEARCH_URL = "https://api.giphy.com/v1/gifs/search?api_key={}&q={}&limit=25&offset=0&rating={}&lang=en"
+API_TOKEN_ENCRYPTED = """
```
matslina, line 18: This secret, when decrypted, seems to be wrapped in json, like: `{"api_token":"..."}`. I think you only want the value of that key-value pair in there.

apps/giphy/giphy.star
```diff
@@ -0,0 +1,132 @@
...
+    API_TOKEN = secret.decrypt(API_TOKEN_ENCRYPTED) or config.get("dev_api_key")
+    search_query = config.str("search_query", "do it")
+    rating = config.str("rating", "g")
+
+    results = search(search_query, rating, API_TOKEN)
+    image_count = len(results["data"])
+
+    image_data = None
+    for _ in range(25):
+        image = results["data"][random.number(0, image_count - 1)]
```
matslina, line 46: This will error out when `image_count == 0`.

matslina (changes requested):
Wonderful app! I'm a bit worried about Giphy's rate-limits though. What are your thoughts on that?

matslina:
Should mention there's some risk we'll have to do some changes for performance reasons if the app becomes popular. The graphics are randomly selected on each run, and we'll very rarely be able to cache the data. That has some impact on how many devices we can support running it. One possible workaround is to only display the (e.g.) first result in a search, or to cycle through them more slowly (perhaps something like `(time.now().unix / 30) % image_count` as an index).

apps/giphy/giphy.star
DigitallyBorn: Yup ... I thought I was going to have another value to include in the secrets, so I created it as json ... ended up not needing it. Should be fixed now.

apps/giphy/giphy.star
DigitallyBorn: Should be handled now.

DigitallyBorn:
> Wonderful app! I'm a bit worried about Giphy's rate-limits though. What are your thoughts on that?

Since it isn't published and I have nothing to show Giphy, I can't promote the API key to a production key and have the rate limits lifted. Once it's on the device I'll take pictures of the device and screenshots of the iPhone app showing the required attribution. Hopefully, it's enough to get the limit lifted. (You have to follow brand guidance and I did the best I could.)

If that fails, I'll remove my API key and provide an input field for users to supply their own. That's not a great UX, but I imagine the TIDBYT audience is pretty techy, so far. It'd be super annoying for people who want to add the app multiple times.

If that fails, you could always take what I've built, put the extra little changes needed to the iOS and Android apps to properly show branding, and kick it out as a non-community app.

life finds a way

DigitallyBorn:
> Should mention there's some risk we'll have to do some changes for performance reasons if the app becomes popular. The graphics are randomly selected on each run, and we'll very rarely be able to cache the data. That has some impact on how many devices we can support running it. One possible workaround is to only display the (e.g.) first result in a search, or to cycle through them more slowly (perhaps something like `(time.now().unix / 30) % image_count` as an index).

It's worth mentioning that it's not really _really_ random. Each time you search you're going to get a fairly stable set of results .. I'm picking a random one of that set. If you've used the shuffle button in Slack's `/giphy` command, it's much the same.

As a napkin-math scenario, I still had Postman results open from when I was working on the app 2 days ago. I just re-requested the same search results and the same gif was the top result.

_HOWEVER,_ I'm caching based on the search query, so there's probably not going to be a high hit rate between app installations (what are the chances people are searching for the same stuff?)

matslina:
Ok, how about I merge this and make it available for beta testers only? If you shoot me a DM on [our Discord](https://discord.gg/rvHynSY3) I can get you added to the beta group and try to move things forward.

PR #594, opened 2022-08-17
Statuspage.io

DigitallyBorn:
Display simple status from an Atlassian Status Page (aka statuspage.io).

These are screenshots when the user configured site is https://status.atlassian.com/. The title "Atlassian" is the name of that statuspage -- it won't always say "Atlassian".

statuspage
statuspage
statuspage

### DIFF
--- a/apps/apps.go
+++ b/apps/apps.go
@@ -120,6 +120,7 @@ import (
 	"tidbyt.dev/community/apps/sportsstandings"
 	"tidbyt.dev/community/apps/spotthestation"
 	"tidbyt.dev/community/apps/stateflags"
+	"tidbyt.dev/community/apps/statuspage"
 	"tidbyt.dev/community/apps/steam"
 	"tidbyt.dev/community/apps/stepcounter"
 	"tidbyt.dev/community/apps/stockticker"
@@ -275,6 +276,7 @@ func GetManifests() []manifest.Manifest {
 		sportsstandings.New(),
 		spotthestation.New(),
 		stateflags.New(),
+		statuspage.New(),
 		steam.New(),
 		stepcounter.New(),
 		stockticker.New(),
--- a/apps/statuspage/statuspage.go
+++ b/apps/statuspage/statuspage.go
@@ -1,0 +1,25 @@
+// Package statuspage provides details for the StatusPage applet.
+package statuspage
+
+import (
+	_ "embed"
+
+	"tidbyt.dev/community/apps/manifest"
+)
+
+//go:embed statuspage.star
+var source []byte
+
+// New creates a new instance of the StatusPage applet.
+func New() manifest.Manifest {
+	return manifest.Manifest{
+		ID:          "statuspage",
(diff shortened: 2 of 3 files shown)


matslina (changes requested):
Wonderful idea, but I'm afraid we currently can't merge apps that call user-provided URLs. Since community apps run in our backend, we have to make sure HTTP calls hit predictable targets to make sure we're not inadvertently spamming anyone.

One option could be to predefine some popular/significant status pages in the code and make it available to the user as a Dropdown? Alternatively, if statuspage.io has some list of sites, then that could be used in a similar fashion.

DigitallyBorn:
> Wonderful idea, but I'm afraid we currently can't merge apps that call user-provided URLs. Since community apps run in our backend, we have to make sure HTTP calls hit predictable targets to make sure we're not inadvertently spamming anyone.
> 
> One option could be to predefine some popular/significant status pages in the code and make it available to the user as a Dropdown? Alternatively, if statuspage.io has some list of sites, then that could be used in a similar fashion.

That makes sense. Unfortunately, I don't know of a safe way to verify that a URL is an Atlassian-hosted status page with the tools on hand. They do host cheap/free statuspages on `[whatever].statuspage.io`, but nearly all businesses using it use their own domain.

Just thinking about ways of verifying ...
- Verification of CNAME entry in DNS won't be reliable. Customers could point traffic through their own CDN or flatten DNS.
- Verification of the SSL certificate won't even work -- they use Let's Encrypt and don't create a single certificate for each domain -- one of my statuspage has 40 domains included in the same certificate.
- They don't supply reverse DNS entries (ptr records), so we can't do a reverse lookup on the server's IP (it just identifies as a generic AWS EC2 instance/endpoint).

I'm out of ideas. That's sad ... I had memories of a big glowing green box in a data center showing that everything was happy.

matslina:
While I hope we'll be able to allow user provided URLs in the future, it'll take a while to get there as there are many competing priorities. The CNAME check would certainly do the job in this particular case, but ideally we'd find a more general solution to the problem.

But I'm thinking it might not be such a bad idea to have a predefined list of status pages to choose from. It'll certainly make the app a lot more accessible to non-technical users, while opening up for technical users to submit a PR to add a new status page to the list. It'd also provide some sense of discoverability to the app. For instance, here's what I found in a few minutes:

```
status.amplitude.com
status.atlassian.com
status.appsflyer.com
status.box.com
www.cloudflarestatus.com
status.datadoghq.com
discordstatus.com
status.fastly.com
status.grammarly.com
www.pintereststatus.com
www.redditstatus.com
status.sumsub.com
api.twitterstat.us
status.udemy.com
status.zoom.us
```

All appear to be on statuspage.io. I suspect few users would actively go look if Discord was on statuspage.io, but I bet quite a few would happily start tracking it if it was already in the list. WDYT?

jvivona:
we will watch for the commit to main and then add other services we know of (teamviewer, some of the large VoIP providers, etc)

PR #595, opened 2022-08-17
New App: TeslaFi

MrRobot245:
Uses the TeslaFI API to grab some info about your car. Gets the Name, range, and battery charge status and displays it using bars similar to date progress.


MrRobot245:

apps/teslafi/teslafi.star
```diff
@@ -0,0 +1,360 @@
...
+def rgb_to_hex(r, g, b):
+    return "#" + str("%x" % ((1 << 24) + (r << 16) + (g << 8) + b))[1:]
+
+def main(config):
+    if config.str("api") == None:
+        return render.Root(
+            child = render.WrappedText("Please enter the API info!"),
+        )
+    else:
+        rep_cache = cache.get("teslafi")
```
matslina, line 124: I think you want to include the api key in the cache key. The cache is shared across all instances of this app (but not with other apps), so the "teslafi" key will be written to (and read from) on every run otherwise, effectively showing you someone else's data most of the time.

apps/teslafi/teslafi.go
```diff
@@ -0,0 +1,25 @@
...
+var source []byte
+
+// New creates a new instance of the TeslaFi applet.
+func New() manifest.Manifest {
+	return manifest.Manifest{
+		ID:          "teslafi",
+		Name:        "TeslaFi",
+		Author:      "mrrobot245",
+		Summary:     "Shows charge/name/range",
+		Desc:        "Shows your Teslas current Name, Charge in Mi/KM and battery %. Also shows if its charging or not.",
```
matslina, line 20: Will it be obvious to a TeslaFi user how to get the API Key? Would it be helpful to add some pointers in the app description?

matslina (changes requested):

matslina (approved):

PR #596, opened 2022-08-17
New App:  Tautulli

MrRobot245:
Add Tautulli for Plex streaming information. Shows stream count, Transcode/Direct Play count, and Upload bandwidth.


matslina (changes requested):
This looks cool!

We're unfortunately unable to merge apps that call user-provided URLs right now. This is due to the risk of such functionality being abused to do calls to arbitrary servers, potentially causing damage or causing our servers to get blocked from calling certain API providers.

If there's some kind of workaround for that behavior then we'd definitely love to get this merged.

MrRobot245:
Hmm,  I don't think so as its a self-hosted thing. My other PR for Netdata is the same. But the Tesla one is a known-url so hopefully that ones good to go!

matslina:
I'm gonna start labelling PRs when we fail to merge for this reason. Kinda annoyed by it and hoping we can come up with a solution. When we do, we'll ping you about reopening the PR.

PR #598, opened 2022-08-17
Add ethstaker application

ColinCampbell:
This pull request adds an application for tracking the status of validators on the Ethereum beacon chain. You can read about that here: https://launchpad.ethereum.org/en/

This apps loads data using the [Beaconcha.in](https://beaconcha.in) API. Ideally this would use their OAuth API for logging in to get an access token but I'm unable to add an OAuth app at this moment in time. In lieu of that, the app can be configured with an API key. Currently the app shows a list of the recent slots with the status for each. There are a couple complicating factors here:

- there is a limit of 10 validators per request from the API, to get around this I issue multiple requests, and
- the API only returns a small amount of slots, to get around this the app caches the slot statuses so the app can provide a full history eventually

The API provides some summary stats about CPU and disk usage which I'll leverage if and when I get authenticate via OAuth.


ColinCampbell:

ColinCampbell:
There was a problem with Github Actions when I submitted this but I can't figure out how to retrigger the checks, fun

matslina:
Thanks for the PR!

Looking forward to oauth support! Should make it more accessible to other users.

PR #599, opened 2022-08-18
sportsscores: handle league passed in the sport field

matslina:


matslina:
@rs7q5 PTAL

rs7q5:
LGTM from what I can tell!

PR #603, opened 2022-08-22
New app: Outage Detector

joevgreathead:
# What is it?

This is a new app which helps a user by identifying whether a provided URI appears to be up or down by making regular requests and displaying up to the last hour of results.

# How does it work?

The app has three configuration options:
- the URI to track (default: tidbyt.com if none is entered)
- an optional label to display instead of the URI being requested
- a limited set of emoji sources for the chosen check/cross emoji used for display

On run, the app does the following:
- Sets up configurations.
- Cleans up the URI provided, including replacement for any values like `localhost` as a safety measure.
- Extracts the history for the given URI, if it exists, from the cache or initializes a data object if it does not.
- Checks to see if a request can be made (requests are limited to 5 times in 5 minutes, regardless of result).
- Makes a request if possible and updates the current status and data blob.
- Caches the request history.
- Renders.

# Checklist

- [x] I ran make app to generate template files.
- [x] I ran make lint and received no error messages.
- [x] I have included a sample screenshot of the app in action.

# Example

Pixlet render
outage_detector

Local dev render (included with readme)
outage_detector

# Thoughts & feedback

## Arbitrary requests
I am not aware if other apps exist which take arbitrary endpoints and make requests to them. That said, this app is definitely a risky type of situation because it allows a user to enter (just about) any URI and cause Tidbyt's servers to make requests to them. I've done some due diligence in limiting the requests (5 in 5 min) and removing some obvious hacks to reduce risk to the Tidbyt server (excluding localhost values), however, I would totally understand if y'all read this and decided, "we will not support this type of arbitrary execution". If there are alternative measures which you'd like to see included, then please let me know.

## Lack of structure leading to more code
This app was a quick little project over the weekend as I thought to make. I found the amount of code grew rather quickly due to the nature of things not available in Starlark that I'd typically expect such as classes, instance variables, and similar shortcuts. I tried to structure the majority of the methods so they did not rely on mutability (i.e. functional), but the code is still rather long and a lot of it is mostly due to management of the data structures I've applied. This is a good example, in my opinion, of applet code that could be greatly reduced in size if there were more modern OOP hooks or something to help with data structure management.


joevgreathead:
In reviewing recent PRs, I just saw this comment from @matslina 

https://github.com/tidbyt/community/pull/590#issuecomment-1219706325

> As much as I like the idea (and the code, and the design), I'm afraid we currently can't merge apps that make http calls to fully user-provided URLs. This policy is in place to keep our backend from being used to spam or otherwise cause trouble. I hope we can come up with a reasonable way around this in the future, but we're not there right now.

So, that answers that. 🙃

I'll close this out in a bit.

btjones:
You may be able to rewrite the app to use one of those "is the site down" websites so that the check doesn't happen on the TidByt server. Maybe one of them even has an API or as a fallback you could scrape the site. One of these maybe:

- https://www.isitdownrightnow.com/
- https://downforeveryoneorjustme.com/

joevgreathead:
Thanks for the input, @btjones! That is, indeed, a reasonable suggestion for evolving this app design to remove the arbitrary URI factor.

PR #605, opened 2022-08-22
bikeshare: use Dropdown for providers

matslina:


matslina:
@snorremd PTAL. Trying to publish this app today, and noticed the typeahead search isn't making use of the user-provider search pattern. Ends up being a pretty odd user experience in the Tidbyt mobile app, so I'm swapping it out for a Dropdown for now.

PR #608, opened 2022-08-22
Add price of HEX to pulsechain

bretep:
The community is enjoying this app. They have requested that the price of HEX be added to the display.

Can you add this to the "Crypto" sub-category within the app lookup?

https://twitter.com/hexpulseog/status/1561832379123122177


matslina:
Thanks for the PR! Should already be in the crypto collection.

PR #613, opened 2022-08-25
Add App: Random Cats

MrRobot245:
Shows Pictures/Gifs of random Cats


apps/randomcats/random_cats.star
```diff
@@ -0,0 +1,54 @@
...
+load("encoding/json.star", "json")
+load("render.star", "render")
+load("schema.star", "schema")
+load("http.star", "http")
+
+def main(config):
+    if (config.bool("gifs") == False):
+        imgSrc = http.get("https://cataas.com/cat").body()
+    else:
+        imgSrc = http.get("https://cataas.com/cat/gif").body()
```
btjones, line 17: It looks like you can add `width` and `height` query params to this url so that the returned image is already scaled down to the proper size. Doing that would reduce the payload size and speed things up.

i.e.
https://cataas.com/cat?height=32&?width=64
https://cataas.com/cat/gif?height=32&?width=64

btjones (commented):
Hi! This looks like a fun app. I just had a couple questions about the image sizes.

apps/randomcats/random_cats.star
```diff
@@ -0,0 +1,54 @@
...
+    children = []
+    children.append(
+        render.Row(
+            expanded = True,
+            main_align = "center",
+            children = [
+                render.Image(
+                    src = imgSrc,
+                    # width = 65,
+                    height = 35,
```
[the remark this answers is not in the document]
WeedMD-Support: I thought it was 35, 32 makes more sense!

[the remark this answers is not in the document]
MrRobot245: I thought it was 35, 32 makes more sense!

apps/randomcats/random_cats.star
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+load("encoding/json.star", "json")
+load("render.star", "render")
+load("schema.star", "schema")
+load("http.star", "http")
+
+def main(config):
+    if (config.bool("gifs") == False):
+        imgSrc = http.get("https://cataas.com/cat?height=32").body()
```
matslina, line 15: We'll have to add some caching here so the API call isn't made on every single run. In part to not overload the API, in part to simplify some of the caching that happens elsewhere in the Tidbyt backend. A ttl of say 20-30 seconds should do fine.

apps/randomcats/randomcats.go
```diff
@@ -0,0 +1,25 @@
...
+var source []byte
+
+// New creates a new instance of the Random Cats applet.
+func New() manifest.Manifest {
+	return manifest.Manifest{
+		ID:          "random-cats",
+		Name:        "Random Cats",
+		Author:      "mrrobot245",
+		Summary:     "Shows pictures of cats",
+		Desc:        "Shows random pictures of cats/gifs of cats.",
```
matslina, line 20: Would be great to give a call out to cataas.com as well. It's such a wonderful service they provide! =)

matslina (changes requested):
This is awesome! Can't wait to get it out! =)

apps/randomcats/random_cats.star
MrRobot245: Added!

apps/randomcats/randomcats.go
MrRobot245: Added!

matslina:
There's a minor tweak needed still but I'll do that before publishing in a separate PR and will tag you as an FYI. Merge ahoy.

matslina (approved):

PR #617, opened 2022-08-27
Add Blackout & How Old Am I?

mabroadfo1027:
Black out TidByt during evenings (or whenever)

### DIFF
--- a/apps/blackout/blackout.star
+++ b/apps/blackout/blackout.star
@@ -1,7 +1,7 @@
 """
 Applet: Blackout
 Summary: Blackout tidbyt
-Description: Black out TidByt during evenings (or whenever).
+Description: Black out Tidbyt during evenings (or whenever).
 Author: mabroadfo1027
 """
 
@@ -10,19 +10,18 @@ load("schema.star", "schema")
 load("encoding/json.star", "json")
 
 def main():
-
-    pulseList=[]
-    for i in range(0, 1000):
+    pulseList = []
+    for i in range(0, 10):
         pulseList.append(render.Text(".", color = "#8B0000" if i % 10 == 0 else "#000000"))
 
     return render.Root(
         delay = 1000,
         child = render.Box(
-                  render.Column(
-                        main_align="end",
-                        cross_align="end",
-                        expanded=True,
-                        children = [render.Animation(children = pulseList)],
-                     ),
-                  ),
-        )
+            render.Column(
+                main_align = "end",
+                cross_align = "end",
+                expanded = True,
+                children = [render.Animation(children = pulseList)],
+            ),
+        ),
+    )
--- a/apps/blackout/blackout.star
+++ b/apps/blackout/blackout.star
@@ -1,0 +1,28 @@
+"""
+Applet: Blackout
+Summary: Blackout tidbyt
+Description: Black out TidByt during evenings (or whenever).
+Author: mabroadfo1027
+"""
+
+load("render.star", "render")
+load("schema.star", "schema")
+load("encoding/json.star", "json")
+
+def main():
+
+    pulseList=[]
+    for i in range(0, 1000):
+        pulseList.append(render.Text(".", color = "#8B0000" if i % 10 == 0 else "#000000"))
+
+    return render.Root(
+        delay = 1000,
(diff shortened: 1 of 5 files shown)


apps/howoldami/howoldami.star
```diff
@@ -0,0 +1,147 @@
...
+def get_schema():
+    options = [
+        schema.Option(display = "US/Central", value = "US/Central"),
+        schema.Option(display = "US/Eastern", value = "US/Eastern"),
+    ]
+
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Text(
```
matslina, line 126: Have you seen the [`Datetime`](https://github.com/tidbyt/pixlet/blob/main/docs/schema/schema.md#datetime) schema field? Should be a great fit for this data and make it a lot easier for users to enter correct information.

apps/howoldami/howoldami.star
```diff
@@ -0,0 +1,147 @@
...
+                desc = "Birthdate (format: 2020-01-01)",
+                icon = "user",
+            ),
+            schema.Text(
+                id = "birthtime",
+                name = "Birthtime",
+                desc = "Birthtime (format: 13:00:00)",
+                icon = "user",
+            ),
+            schema.Dropdown(
```
matslina, line 138: I would suggest using the (kinda magic) `$tz` config parameter. It's provided automatically and will match the timezone of the location the user has selected for their Tidbyt device.

In cases where it's important to make timezone configurable per app installation, the simplest way is to use the `Location` schema field. It allows the user to pick any location in the world, and the associated timezone can be found in the config data passed to the app. But for this app I bet `$tz` will do just fine.

apps/blackout/blackout.star
```diff
@@ -0,0 +1,28 @@
...
+"""
+
+load("render.star", "render")
+load("schema.star", "schema")
+load("encoding/json.star", "json")
+
+def main():
+
+    pulseList=[]
+    for i in range(0, 1000):
```
matslina, line 15: I don't think there's any need to generate a full 1000 frames for this animation. Since every 10th frame gets the fully black color, it ought to be sufficient to have 10 frames in there. The Tidbyt device will loop the animation while it's being displayed.
```suggestion
    for i in range(0, 10):
```

matslina (changes requested):

apps/howoldami/howoldami.star
matslina: I thiiiiink it'll also give you the time adjusted to UTC, removing the need for juggling timezones.

matslina:
Thanks for the PRs! Seems like two fun apps that folks will enjoy. Just a couple of suggestions above that I think could be good to get in before merging.

apps/howoldami/howoldami.star
mabroadfo1027: Good to know! Very useful here

apps/howoldami/howoldami.star
mabroadfo1027: Also good to use here - updated.

matslina (approved):

apps/howoldami/howoldami.go
```diff
@@ -0,0 +1,25 @@
...
+
+// New creates a new instance of the How Old Am I applet.
+func New() manifest.Manifest {
+	return manifest.Manifest{
+		ID:          "how-old-am-i",
+		Name:        "How Old Am I",
+		Author:      "mabroadfo1027",
+		Summary:     "Calculates age",
+		Desc:        "Calculates age based on given date and time.",
+		FileName:    "how_old_am_i.star",
```
matslina, line 21:
```suggestion
		FileName:    "howoldami.star",
```

apps/howoldami/howoldami.go
```diff
@@ -0,0 +1,25 @@
+// Package howoldami provides details for the How Old Am I applet.
+package howoldami
+
+import (
+	_ "embed"
+
+	"tidbyt.dev/community/apps/manifest"
+)
+
+//go:embed how_old_am_i.star
```
matslina, line 10:
```suggestion
//go:embed howoldami.star
```

PR #620, opened 2022-08-29
Add astronomical method to Shabbat app

dinosaursrarr:
Follow up to https://github.com/tidbyt/community/pull/615. Adding as a second, dependent PR because it requires pixlet to be updated with https://github.com/tidbyt/pixlet/pull/381.

Adds a more precise way to calculate the end time based on the position of the sun. Really the idea is to identify when it's dark enough to see three small stars in the night sky. The other methods use heuristics, but they may not work right in all parts of the world. Computing the sun's position relative to the horizon for a given location should be more accurate.


dinosaursrarr:
Huh -- bit worrying that this passes all the tests. The starlark script relies on method that's unknown in the current version of the runtime.

dinosaursrarr:
Merged into https://github.com/tidbyt/community/pull/615 as there's now a tagged pixlet version with the updated sunrise module.

PR #622, opened 2022-08-30
Noaa tides app intial

tavdog:
New app for NOAA tides display.


tavdog:
> @tavdog - the URL for pulling the data from the API - the timezone needs to be changed to LST_LDT to account for stations that change between Standard Time and Daylight Saving Time. Line 122 should be changed to this (I tried to check it in, but permissions, etc and since it's not in tidbyt/community:main yet - I figured I would drop a note here.
> 
> url = "https://api.tidesandcurrents.noaa.gov/api/prod/datagetter?product=predictions&date=today&range=48&datum=MLLW&station=%s&time_zone=LST_LDT&units=english&interval=hilo&format=json&application=NOS.COOPS.TAC.TidePred"

Thanks. Adding that change now.

apps/noaatides/noaa_tides.star
```diff
@@ -0,0 +1,304 @@
...
+"""
+
+def debug_print(arg):
+    if print_debug:
+        print(arg)
+
+def get_stations(location):  # assume we have a valid location dict
+    stations_json = {}
+    station_options = list()
+    url = "https://api.tidesandcurrents.noaa.gov/mdapi/prod/webapi/tidepredstations.json?lat=%s&lon=%s&radius=50" % (location["lat"], location["lng"])
```
matslina, line 37: Would you mind adding some code to  truncate lat/lng before doing this call? We do this to avoid unnecessarily leaking/sharing the user's exact location. I'm thinking 1 or at most 2 decimal points would be sufficient for this application. You can find some info on how this affects accuracy [over here](https://gis.stackexchange.com/questions/8650/measuring-accuracy-of-latitude-and-longitude), and a few examples of how other apps handle it [over here](https://github.com/tidbyt/community/search?q=truncate).

matslina (changes requested):
Thanks for the PR! I think a bunch of people will really enjoy this! Just need that one location fix in first.

apps/noaatides/noaa_tides.star
tavdog: ok, all done.

matslina (approved):

PR #625, opened 2022-08-31
SF NextMuni Updates

dja852:
New Features:
* Route Filter: Filter predictions to only one route for a given stop
* Prediction Formats: Two Line w/Destination only shows predictions for the next two lines, two times, w/ destination. Larger Font
* Prediction Formats: Two Line w/4 Times only shows predictions for the next two lines, 4 times, no destination. Larger Font

Enhancements:
* Added line colours for E and F lines
* Display route letter for F and S lines as black to provide better contrast
* Improved KT route tag override to properly show K or T depending on which stop and to match Muni signage
* Better handling of default config values when pixlet commands are used as get_schema is not currently called then

### DIFF
--- a/apps/sfnextmuni/sf_next_muni.star
+++ b/apps/sfnextmuni/sf_next_muni.star
@@ -21,11 +21,14 @@ DEFAULT_LOCATION = """
 	"timezone": "America/Los_Angeles"
 }
 """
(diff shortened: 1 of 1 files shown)


dja852:

matslina:
@juniormonkey Would've love to get your feedback on this one.

juniormonkey (approved):
This is fabulous, thank you so much.

PR #626, opened 2022-08-31
cla: Add alternate renovate.

betterengineering:
This commit adds an additional renovate bot. At some point, renovate
changed from an account to an app on GitHub which broke this
integration.


betterengineering:
cc @dinosaursrarr I think this will do it 😅

PR #627, opened 2022-09-01
Adds maze drawing app

gstark:
Draws a random maze and then animates the solution.

Maze


gstark:
There is one warning with the linting process. It is claiming a syntax error on a `while` statement. However, the syntax appears to be ok.

gstark:

rs7q5:
> There is one warning with the linting process. It is claiming a syntax error on a `while` statement. However, the syntax appears to be ok.

Can't have while loops in the code as infinite loops are bad. Easiest solution would be just to switch it to a for loop with a max number of iterations and break out of it early when conditions are met.

gstark:
@rs7q5 Thanks for the heads up!

matslina:
Hey @gstark. Thanks for the PR, and apologies for the very long lead time here.

There's a bit of an issue with apps that generate completely random content on each run. In a nutshell, it makes it very hard for us to cache the generated graphic as it keeps changing on every execution.

We have a workaround for this which involves each app providing a seed for the random number generator. That "keeps the randomness down", while still keeping the app changing over time.

For now, I'll merge this and try to cut a new PR for the app to move it over to using this feature. It's a bit experimental, but if it works as well as I hope then we should be able to publish this.

Feel free to ping me on discord if you want to stay in the loop. I'll also make sure to tag you here on any follow up PRs I make.

dinosaursrarr:
I see I'm not the only one who's been at the [maze drawing exercises](https://github.com/tidbyt/community/pull/560) :) I really like the animated solution in yours with the changing colours.

PR #633, opened 2022-09-05
New App: Office Status

b-bell:
This application allows the user to hook up a calendar and messaging application using oAuth2.0. It reviews calendar events and their messaging app status to determine whether the user is busy, free, remote, or away. Depending on their status, it will display when they are next available or when they will be busy next. The purpose of this application is for a user to display their availability to coworkers. It allows coworkers to understand whether they are in the office, busy, or free.

HUGE shout out to [Matt Fisch](https://github.com/Matt-Pesce), who has an active PR open for an Outlook application. Their work saved me a lot of time figuring out oAuth with the Graph and Webex API's.

screenshot


b-bell:
After re-reading the docs on caching, it seems like I may need to update my cache settings to use a user-specific key when storing a user's calendar and Webex details. Let me know if that's accurate and I can make the change.

rs7q5:
All users share the same cache key, so they will need to be unique.

b-bell:
Okay, I've added user-specific cache keys based on their access token. In addition, I also included retry logic for the Webex API, and I fixed a bug with how the app handles some status combinations.

b-bell:
Hello! I just wanted to post a quick follow up message that this PR is ready for review.

I understand it may take some time to obtain a registered app for oAuth into Microsoft and Webex, but would it be possible to get a review of this PR so that I may correct any unrelated issues while those credentials are blockers?

kcwoodfield:
@b-bell nice work on this.  Hopefully this gets approved!

bpotter16:
@b-bell, will definitely use this as on my apps. Nice!

PR #634, opened 2022-09-05
feat: Add Textbyt applet

joshareed:
Add applet for displaying text messages on your Tidbyt.

screenshot


joshareed:

joshareed:
Yep: https://github.com/joshareed/textbyt/tree/main/api

It's running on Google's Firebase + Cloud Functions. The free tier is pretty generous. The text messaging is using Twilio.  All texts that come in get posted to the cloud function for parsing.

joshareed:
You can read the code but I exposed it as a simple REST API and then added the Twilio handling on top of that. So you can interact with the API programmatically as well without the text integration.

```
POST https://us-central1-textbyt-rest-api.cloudfunctions.net/textbyt/v1/<feed id>
{
  "author":"Josh",
  "message":"Some message"
}
```

joshareed:
Also, I'm not worried about the cost or reliability given it's Google. However, I totally get it if you think the idea is cool but would prefer the API were not provided by some rando on the internet. As someone who is on-call always for their day job, the idea might make me a bit twitchy.  If you wanted to reimplement on your own infra, the data model and API were designed to be dead simple.  I was just looking for a way to text my kids affirmations and dad jokes during the weeks they are at my ex's. 👍

btjones:
I don't work for Tidbyt - just another community contributor who was curious how you put it all together. I love the idea of the Tidbyt being a device we can interact with outside of the Tidbyt app so this PR definitely caught my eye.

joshareed:
Oh, nice. Yeah, Twilio is pretty straightforward to work with, especially for texting. I haven't done any Node or Typescript (so don't judge the code too harshly) but I was pleasantly surprised how easy it was to stand something up on Firebase.  Since it is an API behind the scenes, you could probably hook up other integrations/notifications via something like IFTTT, e.g. display a message when a package is delivered or something.

matslina:
Hey @joshareed. This is pretty awesome! =)

To be able to publish this, we'll need an updated app description in `textbyt.go`. That's the content users will see when they go to add the app on iOS/Android. Your README does a great job explaining things, so hopefully just copy pasting that over should be sufficient. We'd ideally want a not-very-technical user to understand how to make use of the app.

The second thing I'd like to see added in the app description is something to highlight that it's not Tidbyt, Inc. that's running the SMS service. Users should be very aware that if they use the app, the data they're sending will be passing through someone else's systems, and Tidbyt can't guarantee that the data is protected.

joshareed:
@matslina thanks for taking a look. I made the requested description updates and expanded the README.

matslina:
Thanks for the changes! Looking forward to play with this.

PR #635, opened 2022-09-05
Add LeetCode stats app

jmanske:
A new app for displaying your LeetCode stats.


jmanske:

matslina:
Hey @jmanske! This looks great, and I'm sure lots of leetcode folks will love it! =)

Wanted to ask if you could add a default username for when no (or an empty) `user_name` is passed in? That way the app previews shown in the Tidbyt iOS/Android app say more than just "no user name supplied", and help explain the purpose of the app to regular users.

jmanske:
Great idea. There is a user called "LeetCode" (an admin) who seems as good a choice as any. I pushed a change to use that user when no user is supplied in the config.

PR #636, opened 2022-09-05
Add solar elevation app

dinosaursrarr:
A clock for when you can't look out of the window or at an actual clock. How high is the sun above or below the horizon right now?


matslina:
So useful for folks without windows! =)

PR #644, opened 2022-09-07
Add HEX Daily Stats app

kmphua:


apps/hexdailystats/hex_daily_stats.star
```diff
@@ -0,0 +1,186 @@
...
+        return render.Root(
+            child = render.Box(
+                display_error,
+            ),
+        )
+
+    hex_price = str("$%f" % float(coin_data["market_data"]["current_price"]["usd"]))
+
+    # Get HEX daily data updates
+    daily_data_updates_response = http.post(HEX_SUBGRAPH_URL, body = DAILY_DATA_UPDATES_QUERY, headers = POST_HEADERS)
```
matslina, line 64: Would it be appropriate to cache these calls? I can't say I understand the queries, but if it's daily data then a TTL of  several hours would be fine right?

apps/hexdailystats/hex_daily_stats.star
```diff
@@ -0,0 +1,186 @@
...
+HEX_SUBGRAPH_URL = "https://api.thegraph.com/subgraphs/name/codeakk/hex"
+POST_HEADERS = {
+    "Content-Type": "application/json",
+}
+DAILY_DATA_UPDATES_QUERY = "{\"query\": \"{ dailyDataUpdates(orderBy: timestamp orderDirection: desc first: 1) { payoutPerTShare } }\" }"
+GLOBAL_INFOS_QUERY = "{\"query\": \"{ globalInfos(orderBy: timestamp orderDirection: desc first: 1)  { hexDay, shareRate } }\" }"
+
+def main(config):
+    # Get coin data for selected coin from CoinGecko
+    coin_data = get_json_from_cache_or_http(COINGECKO_PRICE_URL.format("hex"), ttl_seconds = 600)
```
matslina, line 40: I think this will be fine, but wanted to mention that there's some risk we'll end up hitting Coingecko's rate limit for the free API. This will be the third app using Coingecko with a 600s TTL, and I'm guessing it'll fit in the free tier, but if not we'll have to revisit and fold this (and the other) apps in under Tidbyt's paid plan (currently in use for the Crypto app).

apps/hexdailystats/hex_daily_stats.star
kmphua: Agreed. I added an 8 hour cache for the daily data.

apps/hexdailystats/hex_daily_stats.star
kmphua: Coingecko's free tier is rate-limited at 50 calls/min so I think should be ok.

PR #645, opened 2022-09-07
WiFi QR Code Generator App

misusage:
Hello! This is my first app for the Tidbyt. It will create a working WiFi QR code. If you scan the QR code with your phone, your device will automatically join the network!

This app would be amazing to display in your home, bar, restaurant, and/or shop!

There are encoding limitations with the QR library as well with Enterprise networks but I did mention them in the README. Please let me know if you have any questions!

Scan this QR code below as an example. Your phone will try to join the following network (SSID: Network, Password: Password123). Of course, it doesn't exist so your phone won't be able to join. Move your phone away from the screen if it doesn't read right away.
wifi_qr_code


misusage:

betterengineering (approved):
This is such a great use case! Thanks for putting this together!

betterengineering:
Oh, one thing to note. We typically don't take password fields in schema. They're stored in our database and we don't really want to be in the business of storing passwords. I think this one feels pretty low risk, given the user would creating a highly visible display to allow folks on their WiFi network, so it's not like their password is overly sensitive in this case.

PR #653, opened 2022-09-12
New App: Costco Gas

dja852:
Displays current gas prices for a selected [Costco](https://www.costco.com) warehouse in the US.

## Features

* Using an entered address, choose from a list of the nearest 20 warehouses with gas stations
* The applet will display the price for Regular (R), Premium (P), and if available, Diesel (D)
* Choose the colour scheme for gas prices: all white, red for gas/petrol and green for diesel, or green for gas/petrol and red for diesel
* Choose to display an icon (Costco logo or Gas Pump logo) or opening hours
* Opening hours will reference your device timezone ($tz) to detemine the current day and open status of the gas station
  * Device will display OPEN or CLOSED (based on current time) followed by the opening and closing times
  * Choose to display times in 24-hour clock (default) or 12-hour clock

## Data Source

Data is being sourced from the Costco Website Warehouse Finder API and as provided AS-IS only. Website updates made by Costco may cause this app to fail and may require updates. 

Even though gas (petrol) stations are available in other countries like Canada and the UK, only US warehouses reliably return prices at the present time.


apps/costcogas/costco_gas.star
```diff
@@ -0,0 +1,387 @@
...
+        if http_data.status_code != 200:
+            fail("HTTP request failed with status {} for URL {}".format(http_data.status_code, url))
+        response_data = [x for x in http_data.json() if type(x) == "dict"]
+        cache.set(url, json.encode(response_data), ttl_seconds = ttl)
+
+    return response_data
+
+def get_warehouses(location):
+    loc = json.decode(location)
+    warehouses = get_cached_data(API_WAREHOUSE_SEARCH.format("&latitude=" + loc["lat"] + "&longitude=" + loc["lng"]), 86400)
```
matslina, line 209: Would you mind adding some code to truncate latitude and longitude? This would help protect against leaking users' exact location to the API provider. As a secondary benefit, it also improves cache rate a bit in densely populated areas.

I'm thinking 1 or 2 decimal points would be sufficient, based on [this discussion](https://gis.stackexchange.com/questions/8650/measuring-accuracy-of-latitude-and-longitude)?

You can find some examples over here: https://github.com/tidbyt/community/search?q=truncate

matslina (changes requested):
Neat app! Just this one location fix and we should be able to merge and test it.

apps/costcogas/costco_gas.star
dja852: Done! Truncated to one decimal point as that should be resonably precise for a list of stores

matslina (approved):

PR #655, opened 2022-09-14
Remove Verge Taglines app

joevgreathead:
# What is this?

This is a request to remove the Verge Taglines app (original author: me).

# Why?

Today, [theverge.com](https://www.theverge.com) rolled out their new site design and, unfortunately for this app, it no longer includes a header image with a paired tagline. Currently, the app in production is either failing or is producing the last available header + tagline.

Work could be done to redesign this app to align with the Verge's new design in some novel way, however, I don't have insight into their design plans and rollout strategy so any guesses would lead to an unknown amount of work for myself and a degraded user experience. Instead, I've decided it's best to remove the app as-is.

I removed the following items which are, to my knowledge, the only dependencies:
- two references in `apps.go`
- `verge_taglines.star`
- `vergetaglines.go`


joevgreathead:
Actually, nevermind. There may be hope for this one yet.

PR #662, opened 2022-09-15
Add app: Premier League Scores

mabroadfo1027:
Special thanks to LunchBox8484 - heavily based on MLS Scores app


matslina:
Far out!

PR #663, opened 2022-09-15
California renewables app

sloanesturz:
Make an app that displays the percent of the CA grid that is powered by renewables.

ca_renewables


matslina:
Cool!

PR #667, opened 2022-09-16
feat: zoom call status applet

CFitzsimons:
It's a fairly simple application that just hits the Zoom API to check the user's current status.  If they're in a call, presenting, or on "Do Not Disturb" then the applet displays an "On Air" message.

I'm open to any feedback, suggestions, or alterations.  I've been using this locally for a bit and decided to open it out to the public on the chance some folks will find it of use!

The only caveat is that it might not be incredibly obvious for a non-technical user to figure out how to get a JWT token. Even if the Zoom developer docs are pretty straightforward it's still a concern.  I'm open to suggestions on how to make it easier.


CFitzsimons:

CFitzsimons:
recheck

apps/zoomcallstatus/zoom_call_status.star
```diff
@@ -0,0 +1,105 @@
...
+                render.WrappedText(message),
+            ],
+        ),
+    )
+
+def get_user(expected_email, token):
+    res = http.get(PAGED_URL, headers = {"Authorization": "Bearer " + token})
+    if res.status_code != 200:
+        return "Invalid token"
+    foundUser = cache.get("status")
```
matslina, line 51: The cache is shared across all users and installations of the app, so whenever the `"status"` key is present in cache all users will see the same result. I think the best way to deal with that is to include the JWT in the cache key. That way it a) becomes unique for each user and b) is only accessible for users who have access to the JWT.

If the JWT is large, it could be convenient to [hash it](https://github.com/qri-io/starlib/tree/master/hash).

apps/zoomcallstatus/zoom_call_status.star
```diff
@@ -0,0 +1,105 @@
...
+                render.Text("On Air"),
+                render.Image(src = RECORD_ICON, width = 20, height = 20),
+            ],
+        ),
+    )
+
+def main(config):
+    user_email = config.str("email")
+    token = config.str("token")
+    if user_email == None or token == None:
```
matslina, line 96: It'd be great if the app could show an illustrative default graphic when no config is present. That way users can get a sense of what the app does and what it looks like, before they make the leap to generate JWTs and set it up for real.

apps/zoomcallstatus/zoom_call_status.star
```diff
@@ -0,0 +1,105 @@
...
+        fields = [
+            schema.Text(
+                id = "email",
+                name = "Zoom User Email",
+                desc = "Your personal zoom email.",
+                icon = "user",
+            ),
+            schema.Text(
+                id = "token",
+                name = "Zoom JWT Token",
```
matslina, line 31: Looks like Zoom have support for oauth2 as well. It's a bit tricky to get it working, but thought I should mention it in case you hadn't seen it. =)

matslina (changes requested):

Issue #39, opened 2022-01-10
Provide "Color" picker as schema type


joeyhoer:
For apps which allow colors to be customized, it would be helpful if there was a color picker option available within schema. Right now, it appears that the best option is to utilize a list of predefined values with a `dropdown` schema.

```
  colors = [
    {"text": "White", "value": "#fff"},
    {"text": "Red", "value": "#f00"},
    {"text": "Green", "value": "#0f0"},
    {"text": "Blue", "value": "#00f"},
    {"text": "Yellow", "value": "#ff0"},
    {"text": "Cyan", "value": "#0ff"},
    {"text": "Magenta", "value": "#f0f"},
  ]
…
    {
      "type": "dropdown",
      "id": "color",
      "icon": "palette",
      "name": "Color",
      "description": "The color to display.",
      "options": colors,
      "default": DEFAULT_COLOR,
    },
```

My ideal state would be to have the Tidbyt controlled via HomeKit, and use HomeKit to set the color of the display. As long as the app accepts a single color variable, I believe this should be possible.

LukiLeu:
There was one added a few weeks ago: [https://tidbyt.dev/docs/reference/schema#color](https://tidbyt.dev/docs/reference/schema#color)

tavdog:
Time to close as completed ?

PR #668, opened 2022-09-17
Updating bigclock with RGB color pickers

joeyhoer:
Updating bigclock to use RGB color values rather than a predefined list of colors for full customizability.

In absence of an official color picker schema solution (https://github.com/tidbyt/community/issues/39), I've added RGB color values as a dropdown to allow for full color customizability rather than forcing users to select from a predefined list of colors. A number schema solution (https://github.com/tidbyt/community/issues/298) would also be superior to a drop down, but those tickets were closed due to inactivity.

I considered using a text field an having the user input a HEX string value for the color, but I decided that input validation would be too difficult. The "number selector" dropdown approach (while inelegant from a configuration UI perspective) ensures that the color will always be valid.

### DIFF
--- a/apps/bigclock/big_clock.star
+++ b/apps/bigclock/big_clock.star
@@ -9,6 +9,8 @@ Author: Joey Hoer
 
 load("render.star", "render")
 load("schema.star", "schema")
+load("math.star", "math")
+load("re.star", "re")
 load("time.star", "time")
 load("sunrise.star", "sunrise")
 load("encoding/base64.star", "base64")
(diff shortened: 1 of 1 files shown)


apps/bigclock/big_clock.star
```diff
@@ -150,8 +162,18 @@ def main(config):
     is_24_hour_format = config.bool("is_24_hour_format", DEFAULT_IS_24_HOUR_FORMAT)
     has_leading_zero = config.bool("has_leading_zero", DEFAULT_HAS_LEADING_ZERO)
     has_flashing_seperator = config.bool("has_flashing_seperator", DEFAULT_HAS_FLASHING_SEPERATOR)
-    color_daytime = config.get("color_daytime", DEFAULT_COLOR_DAYTIME)
```
matslina, line 153: A bit over half of all installations of `big-clock` are using a non-default value for `color_daytime` or `color_nighttime`. It would be great if we could make this change backwards compatible so that these installations continue to function as before.

joeyhoer: I was also a bit concerned about backwards compatibility, and I agree that most users will find the RGB values confusing. I was actually considering using HSL instead, as that's a little easier to understand, but ultimately against it.

I could not think of a way to support both the selector and the default color, unless the user types in a hex value — but again, that would require input validation.

I believe the best solution here is a color picker schema.

matslina: Color picker schema would be great. It's just a matter of finding the time to build it out. Lot of other priorities as always. =(

For backwards compatibility, the app should still receive the `color_daytime` parameter for existing installations (even when it's been removed from `get_schema()`), so it'd just be a matter of checking for that field first (with the same default) and then letting the new RGB parameters override it.

matslina: Yes, it should still be passed the same way for older installations.

When there's a colorpicker field, you'd need to introduce a new field ID, like `colors_daytime_rgb`, to keep backwards compatibility.

joeyhoer: Since the RGB values are not intuitive, and there would be more complexity switching back and forth between the various methods to achieve backwards compatibility, I think it would be best not to use the RBG color selectors for now, and wait until there is an official colorpicker schema.

I'm going to close the pull request, and hopefully we can come back to this at a later date. Thanks!

PR #670, opened 2022-09-18
Add London bike app

dinosaursrarr:
Santander Cycles are the official street bikes in London. But they're normally called Boris Bikes as he was the mayor at the time they were introduced. They were literally the only good thing he's ever done, and they were Ken Livingstone's idea anyway.

Features counts for e-bikes, which is very bleeding edge state of the art, as they've only just started introducing them last week. Also features a terrible drawing by me.


matslina:
I think the drawing looks great! Thanks for yet another useful, readable and easy to review app!

PR #674, opened 2022-09-20
New App: Plausible Stats Display

brettohland:
# Plausible Stats On Your Tidbyt

An app for the [Tidbyt](https://tidbyt.com) that uses the [plausible.io stats API](https://plausible.io/docs/stats-api) to display metrics about your website.

## Features

1. Each instance of the app shows a single metric (all metrics are supported)
2. The time period of the metric is fully customizable
3. An optional historical chart of that metric can be shown (and the chart's time period is customizable as well)
4. Your site's favicon is fetched and displayed automatically (or a custom path can be set)

Metrics are fetched every 10 minutes.
Favicons are fetched once per day.

## Options

The following options can be set via the Schema, or passed in via the `pixlet render` command:

| Option Key           | Accepted Value                                                             | Required | Description                                                                                                                                                                 |
| -------------------- | -------------------------------------------------------------------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `domain`             | Any String                                                                 | YES      | The domain of the website to display. The scheme "http://" and "www" are not needed.                                                                                        |
| `plausible_api_key`  | Any String                                                                 | YES      | Generate a new API key at https://plausible.io/settings#api-keys                                                                                                            |
| `metric`             | String: `pageviews`, `visitors`, `bounce_rate`, `visit_duration`, `visits` | NO       | The metric to display. Defaults to `pageviews`                                                                                                                              |
| `time_period`        | String: `day`, `7d`, `30d`, `month`, `6mo`, `12mo`, `custom`               | NO       | The time period for the stats to display. Defaults to `day`. **NOTE** `custom` will get the metric value count for all time                                                 |
| `should_show_chart`  | Bool                                                                       | NO       | Defaults to `true`. Controls the visibility of the historical chart                                                                                                         |
| `chart_time_period`  | String: `7d`, `30d`, `month`, `6mo`, `12mo`                                | NO       | Defaults to `7d`. The historical time period for the chart display                                                                                                          |
| `use_custom_favicon` | Bool                                                                       | NO       | Defaults to `true`. If yes, the app will attempt to download the first `png` file named `favicon`, `favicon-32x32`, or `favicon16x16` from the root of the domain passed in |
| `favicon_path`       | Any String                                                                 | NO       | The relative path to the favicon file to display eg. `/favicon/favicon.png`                                                                                                 |

## Screenshots

screenshot

---

screenshot-store

(As seen when adding in the app)

---

screenshot-no-chart

(When the chart is disabled)

---

screenshot-bounce-rate

(When a stat's name is longer and needs to scroll)

---

screenshot-missing-key

(Error: No API key)

---

screenshot-invalid-domain

(Error: No/Invalid domain)


apps/plausiblestats/plausible_stats.star
```diff
@@ -0,0 +1,569 @@
...
+""")
+
+def main(config):
+    # For the "add" screen, show only the logo.
+    if config.get(DOMAIN_KEY) == None and config.get(PLAUSIBLE_API_KEY) == None:
+        return render_error_screen(None)
```
matslina, line 122: It'd be sweet if the app could show an illustrative default graphic that shows off what the app looks like when fully set up. This would make it easier for users to understand what the app is all about, and to decide if they want to hop on over to Plausible.io, grab API keys and start exploring.

apps/plausiblestats/plausible_stats.star
```diff
@@ -0,0 +1,569 @@
...
+# Makes a call to the plausible.io stats endpoint.
+#    endpoint: the path to the API to call eg. "/aggregate" or "/timeseries"
+#    toke: The Auth token
+#    domain: The domain to check
+#    time_period: The time period to check (doesn't support custom date ranges)
+#    metric: The metric value to return
+def get_plausible_data(endpoint, token, domain, time_period, metric):
+    print("Getting data from Plausible API")
+
+    cache_id = "_".join([REQUEST_CACHE_ID, endpoint, domain, time_period, metric])
```
matslina, line 476: I think you need `token` to be part of the cache key. Otherwise any user will be able to call up cache records for any domain for which another user has provided the correct token.

matslina (changes requested):
Hey! This looks really cool.

There is one problem though: we're currently not able to merge apps that make calls to arbitrary URLs. In this case, the URL won't be entirely arbitrary, but the domain is, and I don't think we can merge it because of that in its current form. This is so to protect against malicious users using the app (and by extension the Tidbyt backend) to spam arbitrary hosts on the internet. We hope to come up with a good way of handling use cases like this, but we're not there yet.

apps/plausiblestats/plausible_stats.star
brettohland: That's a really good point. I think my mental model of the cache didn't include the idea that it was shared across everyone. But that makes sense. I've made that change to include it.

apps/plausiblestats/plausible_stats.star
brettohland: Very true. I've revamped the demo screen to better reflect how the app works:

screenshot-store

brettohland:
> Hey! This looks really cool.
> 
> There is one problem though: we're currently not able to merge apps that make calls to arbitrary URLs. In this case, the URL won't be entirely arbitrary, but the domain is, and I don't think we can merge it because of that in its current form. This is so to protect against malicious users using the app (and by extension the Tidbyt backend) to spam arbitrary hosts on the internet. We hope to come up with a good way of handling use cases like this, but we're not there yet.

I completely understand the reasoning behind this. The less chance for malicious actors, the better. However, I did take this into account while building this app and I'd like to give an overview of how it's been fully mitigated.

First off, the domain that is included in the schema is primarily used as a URL Query parameter when calling the Plausible API. It never connects to the supplied domain during the call. Plausible's API will throw a `401` error if the domain that is provided is not associated with the account who's API key was used at the authentication token. That's the first level of verification.

If that call returns a `200` HTTP status, we know that the user who added the app is guaranteed to be in control of the domain in question. At this point, the app will then attempt to fetch the `favicon` on the website using the set favicon file names inside of the app. If the user has the favicon in a custom folder on their web server, they can use the custom favicon path in the schema, but it's limited 1. Being a relative path from the `domain` provided, and 2. Still only looks for the standard favicon names provided in the app. Otherwise, it uses the fallback globe icon image.

Once the favicon is found, it's cached aggressively for a full day. This, combined with the above checks means that at most, any Tidbyt server will be making one request per domain, per day.

I'd love to hear your thoughts on this.

matslina:
> I'd love to hear your thoughts on this.

Well I'm sold. =)

matslina (approved):

matslina:
@brettohland When this is in production, it'll need some testing before we publish it for everyone. If you ping me on discord I can help you get beta access to it so you can take a look.

PR #677, opened 2022-09-22
borisbikes: updated secret

matslina:


matslina:
@dinosaursrarr Forgot to check the secret, and it too had to be updated to match the .star filename.

dinosaursrarr:
Thanks! I might need to check a couple of my other apps. 

I was never quite clear if the app name for encrypting was the folder name or the script name. Make app generated them differently when there's a space in the name. 

Might be a good one to add to the new docs site? :)

matslina:
You know what, I'm an idiot. This secret was perfectly fine as it was. My bad! I'll restore it in a bit.

matslina:
but yeah it'd be good to have written down on the doc site! =D

PR #684, opened 2022-09-23
costcogas: decode warehouse param, use value

matslina:


matslina:
@dja852 Wasn't able to release this app last week due to a bug. Found time to troubleshoot it today. I think this fixes it.

PR #685, opened 2022-09-23
jokeAPI applet - fixed marquee scroll height

jvivona:
fixed scroll height to handle 4 line jokes and scroll correctly.


rs7q5:
So did it not previously scroll with jokes that were 4 lines?

Or was it stopping prematurely and not getting the last scrolling line when there were 4 lines?

jvivona:
It would not scroll when you get exactly 4 lines - joke or joke + punchline.   Changing the marquee down to 24px from 32px solves it.    if you force it to get something that will be 4 lines (i used  &idRange=250  when testing) you'll see.

PR #687, opened 2022-09-24
MLB wild card race app

jmanske:
There are a few MLB stats/standings apps out there but I didn't see one for the ongoing Wild Card race standings so I made one myself!

The utility here may not be that great as the season is almost over. However, it will be ready to go for next season!

screenshot


betterengineering (approved):
Looks great!

PR #700, opened 2022-09-29
add guardiannews to apps.go

matslina:


matslina:
@jvivona Some fixes required for this to be publishable.

jvivona:
@matslina thanks for those - my brain operates at a different pace than my fingers some times.   I will get better at this.

quick question - i see the //go:embed line in the go file..  Is that something i should be putting in, or something you all do?  if I need to - I missed it in my NASCAR app and I need to update it.  And I'll update my workflow here.

matslina:
@jvivona Yeah, it needs to be there for nascar too! I think most of this will be added automagically if you run `make app`.

jvivona:
my windows dev environment is a little all over the place & I can't get that to work - so a lot of this I'm still doing by hand.   I need to step back and figure out my environment for building in vscode for windows

will fix NASCAR & do a PR real fast to get it in.

PR #707, opened 2022-09-30
amazing: reduce randomness with a time based seed

matslina:


apps/amazing/amazing.star
```diff
@@ -475,6 +476,9 @@ COLOURS = [
 ]
 
 def main(config):
+    # seed the RNG with a new value every 8 seconds
```
matslina, line 479:
```suggestion
    # seed the RNG with a new value every 15 seconds
```

apps/amazing/amazing.star
```diff
@@ -475,6 +476,9 @@ COLOURS = [
...
+    random.seed(time.now().unix // 8)
```
matslina, line 480:
```suggestion
    random.seed(time.now().unix // 15)
```

dinosaursrarr:
Awesome, thanks!

PR #711, opened 2022-10-03
Amazing: fix off by one error

dinosaursrarr:
This had been annoying me for a while. In the recursive division algorithm, it would close off too many sections by filling in a wall where an earlier stage had put a gap. Now it's a maze!


dinosaursrarr:
Failure looks unrelated to this change -- seems to be about Sound Transit.

PR #719, opened 2022-10-06
Chess Rating app

NickAlvesX:
This is a very simple app to follow your ratings on Chess.com.

chess_rating

I just got my Tidbyt (awesome product btw). This is my first, simple but useful app, which can soon have more animations and functionalities, like showing your peek performance, graphs, and so on, all related to your chess rating/stats.

### DIFF
--- a/apps/apps.go
+++ b/apps/apps.go
@@ -27,6 +27,7 @@ import (
 	"tidbyt.dev/community/apps/busytube"
 	"tidbyt.dev/community/apps/carenewables"
 	"tidbyt.dev/community/apps/charlestownferry"
+	"tidbyt.dev/community/apps/chessrating"
 	"tidbyt.dev/community/apps/chessviewer"
 	"tidbyt.dev/community/apps/climateclock"
 	"tidbyt.dev/community/apps/clockbyhenry"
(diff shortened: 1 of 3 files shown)


NickAlvesX:

matslina (changes requested):
Hey, thanks for the PR!

Looks great overall. We will need to have some caching added to keep down the request rate going to chess.com. A good place to add that could be `get_player_profile()`.

NickAlvesX:
@matslina ok! I will use the cache to also add a new feature to track the gain/loss of points in a day. Btw, what is the approach recommended to keep the data "updated"?

matslina:
@NickAlvesX

The cache has a time-to-live (TTL) for each record. You can set it with the `ttl_seconds` argument to `cache.set()`, e.g.: `cache.set("my key", "my data", ttl_seconds=3600) # cache for 1 hour`.

When a Tidbyt device is about to display an installation of this app, the `main()` function will be called with the associated configuration data. At that point, check for data in cache (`cache.get()`) and use that if available. If it's not available, hit the API and write a record with `cache.set()` for the next run. A good place for this could be `get_player_profile()`, and using the username as cache key.

PR #726, opened 2022-10-10
Adding Should I Deploy app

humbertogontijo:
This app retrives https://shouldideploy.today/ response and show it in tidbyt
should_i_deploy


humbertogontijo:

apps/shouldideploy/should_i_deploy.star
```diff
@@ -0,0 +1,66 @@
...
+
+load("render.star", "render")
+load("schema.star", "schema")
+load("http.star", "http")
+
+SHOULD_I_DEPLOY_URL = "https://shouldideploy.today/api?tz="
+DEFAULT_TIMEZONE = "UTC"
+
+def main(config):
+    tz = config.get("tz", DEFAULT_TIMEZONE)
```
matslina, line 16: Would be great if you could add some caching here. We don't want to hit the API too frequently, and the `main()` function will be called every time a Tidbyt running this app is about to show it in its rotation.

matslina (changes requested):

apps/shouldideploy/should_i_deploy.star
humbertogontijo: Thank you for the review. I've added a 2min cache to reduce api calls

PR #732, opened 2022-10-13
Currency converter

rjison:
New App: Currency Converter

Using data with a free account from ExchangeRate-API.com,  should be able to keep this free with caching as much as possible. 

This will display two flags, representing two currencies. It will show the current exchange rate between them as well.

Install the app multiple times if you want to see USA vs. Canada and USA vs. Mexico for example. This scenario will share the same cached data as well.


apps/currencyconverter/currencyconverter.star
```diff
@@ -0,0 +1,858 @@
...
+load("math.star", "math")
+load("schema.star", "schema")
+load("encoding/base64.star", "base64")
+load("encoding/json.star", "json")
+load("secret.star", "secret")
+load("http.star", "http")
+load("cache.star", "cache")
+load("time.star", "time")
+
+exchange_rate_api_key = "382c82c0e47531184b366111"
```
matslina, line 18: Kinda minor, but if the rate limit is per API key, then it would be good to create a new one and make it a secret in here. Otherwise anyone with access to this piece of code can fire off request and screw up our rate limit. I doubt anyone would bother OTOH, but still sort of a healthy practice I think. =)

matslina (approved):

matslina:
Btw, I changed a schema icon to get the build through. Hope it works. Cut a PR if not! =)

apps/currencyconverter/currencyconverter.star
rjison: I agree, anything I've submitted in the past with an API key, I was told Tidbyt would create their own account and replace my API key with theirs. I just left the unencrypted one in there for testing purposes because I wasn't sure the unencrypt call would work for me locally.

Let me know the policy, I'm happy to follow. I thought I was doing it right :( 

Thanks for considering my contribution!

matslina:
Check out https://github.com/tidbyt/community/blob/main/apps/icons.go

apps/currencyconverter/currencyconverter.star
matslina: Yeah we often do. In this case where no user data is passed to the API, and a free key seems to suffice, I don't think there's much need for it. If we run into rate limit problems we could pretty quickly replace it with a paid tier token.

PR #733, opened 2022-10-13
pathtrainschedule: disabling app due to API outage

matslina:


matslina:
@tsgreenberg1217 Disabling this app (in a hacky fashion) for the moment. The PATH train API seems to be completely down, and due to how our tests work it's blocking all other apps from being merged. We'd like to have better separation here so that one app acting up doesn't block all others, but that's not something I can swing right now.

PR #739, opened 2022-10-15
Add 'AirThings' environment sensor readings applet

joshspicer:
Adds environmental sensor readings from an [AirThings](https://www.airthings.com/) device.

Screen Shot 2022-10-15 at 11 35 08 AM

Utilizes the public API (https://developer.airthings.com/consumer-api-docs/), which only supports the client credential oauth flow.  Thus, each user must supply their own clientSecret and clientId, which can be generated for free from https://dashboard.airthings.com/integrations/api-integration.


joshspicer:
Is the cache global for all users of this, or somehow scoped to the individual user?  I am unclear on how existing OAuth are user-scoped based on my understanding of who this applets run on your servers.  

I think this applet may not be compatible with how your infrastructure is structured, but looking for feedback :)

joshspicer:
I think my commits above solve the user-scoped caching issue. @matslina is this the recommended pattern for user level caching?

matslina:
> Utilizes the public API (https://developer.airthings.com/consumer-api-docs/), which only supports the client credential oauth flow. Thus, each user must supply their own clientSecret and clientId, which can be generated for free from https://dashboard.airthings.com/integrations/api-integration.

Looks like they also implement [Authorization Code Flow](https://developer.airthings.com/docs/api-authorization).

> I think my commits above solve the user-scoped caching issue. @matslina is this the recommended pattern for user level caching?

That looks reasonable. The cache is scoped to your app, but shared across users to enable caching of shared data. Using the credentials to construct cache key guarantees that only a user in possession of said data can access the cached records. That said, I think it could be good to compute hash of the credentials and use that as cache key instead?

matslina (changes requested):

joshspicer:
> Looks like they also implement [Authorization Code Flow](https://developer.airthings.com/docs/api-authorization).

Not for consumers.  The docs you linked are only for businesses.

PR #748, opened 2022-10-22
New App: Visible Planets

rjison:
Pick a planet, your location and this will display information on where to look in the sky to see it. It'll tell you the direction, and the degrees above the horizon to look. It'll also tell you what constellation the planet appears in. If the planet is below the horizon, you have the option to skip displaying of that info. Skipping means the app only displays when it is possible to see it.

This app calls AstronomyAPI. I cache the data for at least an hour for each planet/location. If it is during daylight hours, the cache will be set to expire 30 minutes before sunset thereby reducing the number of needless api calls. However, the display can still give you an accurate indication of when the object will be visible, and how far the planet is currently.


apps/visibleplanets/visibleplanets.star
```diff
@@ -0,0 +1,399 @@
...
+        number: number we need to represent in two characters
+    Returns:
+        A two character string that represents the given number
+    """
+    if len(str(number)) == 1:
+        return "0" + str(number)
+    else:
+        return str(number)
+
+def get_bodies(app_hash):
```
matslina, line 320: This function doesn't seem to called. Just in case it's supposed to be. =)

apps/visibleplanets/visibleplanets.star
```diff
@@ -0,0 +1,399 @@
...
+    """
+
+    date_code = "%s-%s-%s" % (two_character_time_date_part(check_time.year), two_character_time_date_part(check_time.month), two_character_time_date_part(check_time.day))
+    tomorrow_date_code = check_time + time.parse_duration("24h")
+    tomorrow_date_code = "%s-%s-%s" % (two_character_time_date_part(tomorrow_date_code.year), two_character_time_date_part(tomorrow_date_code.month), two_character_time_date_part(tomorrow_date_code.day))
+    time_code = "%s:%s:%s" % (two_character_time_date_part(check_time.hour), two_character_time_date_part(check_time.minute), two_character_time_date_part(check_time.second))
+
+    params = {
+        "body": body,
+        "longitude": location["lng"],
```
matslina, line 286: We need lat/lng to be truncated when passed out to external APIs. This to avoid leaking users' location data. In this case, I think 1 decimal point will work just find. Did a few experiments and got pretty consistent results that way. As a bonus, it also means users who're located nearby can rely on each other's cached data so we get fewer API calls overall.

matslina (changes requested):

apps/visibleplanets/visibleplanets.star
rjison: removed -- I needed it to get a list of bodies I could request data on. Then hardcoded those bodies in the config.

apps/visibleplanets/visibleplanets.star
rjison: I originally kept the precision to get the right sunrise/sunset, but I refactored such that I have the precision for sunset/sunrise times, but use the rounded coordinates for api calls and the cache name.

matslina (approved):

PR #758, opened 2022-10-26
New App: MCServerStatus

jakeva:
Track the player count of any Minecraft server!

An example of tracking the Hypixel Network:
display

If a server has a logo, it will use that as the image displayed. If not, it will display a default image, which is a Minecraft Grass Block. If a server is offline, it will display a Minecraft Barrier Block icon and "Offline" text.

With this initial build, only Java Minecraft Servers are supported.


jakeva:

jakeva:
@matslina My apologies, had it as a draft PR due to the automatic testing action failing. Ready to be reviewed now! 😄

apps/mcserverstatus/mcserverstatus.star
```diff
@@ -0,0 +1,98 @@
...
+    rate_cached = cache.get("mc_rate")
+
+    if rate_cached != None:
+        rate = int(rate_cached)
+
+    minecraft_edition = config.get("minecraft_edition") or 0
+    server_ip = config.get("server_ip") or "mc.hypixel.net"
+    server_port = config.get("server_port") or None
+
+    req = http.get("https://api.mcsrvstat.us/" + ("bedrock/" if minecraft_edition else "") + "2/" + server_ip + (":" + server_port if server_port else ""))
```
matslina, line 24: Some caching here would be good! Don't want to unnecessarily hammer the API.

apps/mcserverstatus/mcserverstatus.star
```diff
@@ -0,0 +1,98 @@
...
+
+    if req.status_code != 200:
```
matslina, line 26: Would be great if the app could show some illustrative default graphic when no configuration has been passed. That enables users to see previews of the app before deciding to configure and install it.

matslina (changes requested):
Thanks for the PR! Two notes on changes we need in place before merging.

PR #763, opened 2022-10-29
fix: Bug fixes for office status applet

b-bell:
This PR contains a couple of minor changes to fix bugs in the office status applet.


b-bell:
@matslina Is it possible for the oAuth fields to be optional when this app is deployed? The application is designed to handle cases when a user connects only Outlook or only Webex. This aligns with best practices and allows me to expand this application to other services like Slack and Google Calendar. It would be great for users to be able to add Outlook OR Webex OR Slack, OR Google Calendar.

Issue #765, opened 2022-10-30
Automatic Build and Test Action Failing


jakeva:
Apps are failing the automatic build and test, usually with the following error:
```
FAIL
make: *** [Makefile:18: test] Error 1
Error: Process completed with exit code 2.
```

See: PR #758

jakeva:
@betterengineering I believe this is occurring again.

PR #767, opened 2022-10-31
office status: Add default client ID.

betterengineering:
This commit adds a default client ID so the tests for this repo will pass. There was a shuffle in how apps were created in this repo that lead to this app not having tests running on it.


betterengineering:
Closes #765

PR #769, opened 2022-11-02
fix: oAuth token refresh workflow fixes

b-bell:
The oAuth token refresh workflow was broken. This app only currently works until the first access token expires. These commits fix that issue.


b-bell:
@matslina would it be possible to get this PR reviewed and merged soon? The current office status app doesn't handle token refreshes properly so it only works until the first token expires. :(

matslina:
Apologies for the delay @b-bell. Had a few very hectic weeks over here and PR merging has suffered.

b-bell:
@matslina, not a problem at all! Thank you so much.

PR #770, opened 2022-11-03
States Visited applet

sloanesturz:
This applet shows an in-progress map of the US showing all the states you've been to. Check out the demo gif to see more.

The SVG files in `state-images/` are to help in case we need to regenerate the individual state PNGs later.


apps/statesvisited/states_visited.star
```diff
@@ -0,0 +1,135 @@
...
+    "WY": "Wyoming",
+}
+
+def main(config):
+    return render.Root(
+        render.Stack(
+            [
+                render.Image(base64.decode(STATES[state]), 64)
+                for state in STATES
+                if config.get(state) and config.get(state) != "false"
```
sloanesturz, line 122: There is a small bug in the `pixlet serve` tool. If you are on a URL like `http://localhost:8080/?FL=false`, the string `"false"` gets into the `config` object on this line. This is bad because the string `"false"` is truthy in Pythonic languages (all strings are True except the empty string).

I will file a bug to track this issue.

matslina: This behaviour is actually intentional. Check out `config.bool()` to handle type conversion for you. You can also do something like `config.get("FL") == "true"`.

matslina:
Wonderful PR! Thanks!

PR #776, opened 2022-11-07
New App: Marvel Facts

GothamGamer117:
This app uses Marvel's official API to randomly select a character and display either their character description, or the number of comics/series they have been in (if their description is not available). The character choice is set to change every 30 minutes in order to limit the number of API calls. There are 1,562 possible characters.

An example result can be seen below:
marvelfacts


GothamGamer117:

matslina:
Thanks for the PR!

PR #780, opened 2022-11-09
DateClockTime: added US date format to options

jvivona:
updated @AmillionAirs  app to add option to display date in US format vs Intl format.


AmillionAirs:
I'm cool with these changes since apparently i made an app that drives his wife crazy... hahaha

matslina:
=)

PR #782, opened 2022-11-09
Top Crypto Prices - Finished the app

playak:
Does all I wanted it to do now, including quite some config options.


apps/topcryptoprices/top_crypto_prices.star
```diff
@@ -0,0 +1,331 @@
...
+load("render.star", "render")
+load("http.star", "http")
+load("encoding/base64.star", "base64")
+load("encoding/json.star", "json")
+load("cache.star", "cache")
+load("math.star", "math")
+load("schema.star", "schema")
+
+# INIT
+credits = ["Data:Coingecko", "Code:Playak"]  # Credits required for using free Coingecko API
```
matslina, line 32: Giving credit in the app description should be sufficient to meet the API terms, and could perhaps make for a better user experience?

apps/topcryptoprices/top_crypto_prices.star
```diff
@@ -0,0 +1,331 @@
...
+Author: playak
+
+Version: 1.1
+Date: 2022-11-09
+Comments: Written to display the price info of top cryptocurrencies on a Tidbyt screen. We manually exclude stable coins and wrapped coins (so we only need one API call).
+Copyright: © 2022 Jeroen Houttuin, Playak - jeroen@playak.com - https://playak.com
+"""
+
+# CONFIG
+TTL = 60  # time to live for the API cache. Be nice to the free Coingecko API provider!
```
matslina, line 14: This is probably good enough, but should mention that there are other apps using the free API as well, so they'll all be sharing the same rate limit (although I'm not sure how Coingecko actually implements that, if at all). If we run into problems we can move this app over to Tidbyt's paid API key.

matslina (changes requested):
Thanks for the PR! This looks pretty good. Added a few thoughts in inline comments.

The main blocker for merging right now is just the `apps/topcryptos` directory that looks like a leftover from early development? Also need to remove the references to that app in `apps/apps.go`.

matslina (approved):

apps/topcryptoprices/top_crypto_prices.star
matslina: Looks like you got it. I'll merge this and try to get it published later today.

apps/topcryptoprices/top_crypto_prices.star
playak: Good idea. I made that configurable now, as an extra option. Default is to not show the credits on the Tidbyt.

PR #786, opened 2022-11-10
apps: register ncaamscores

matslina:


matslina:
@cmarkham20

matslina:
@cmarkham20 Hitting lint errors so I'll abandon this. Won't be able to publish the app until it's registered in `apps/apps.go` and passes lint etc.

PR #793, opened 2022-11-11
Made credits hidden. Added more fiat currencies.

playak:
As suggested by matslina, the credit screen is now hidden by default, but can be switched on. Added some major fiat currencies. This app should be pretty complete now imho. Thanks for your help guys.


matslina:
@playak lots of conflicts here apparently. I think you can move past them by either rebasing your branch and force pushing to this PR, or by creating a new fork of the repo and cutting a new PR that way.

playak:
Github remains an obscure mystery to me. All I did was I changed these 3 files and pushed them back up. Of course they conflict - I want to replace them. No idea what to do exactly, but I'll look into it. Writing the app is great fun, but I hate this Github process. But thanks for all your work.

PR #800, opened 2022-11-12
Test that all apps are registered

dinosaursrarr:
Seems like there's been a bunch of issues recently where changes aren't being reflected in apps.go, then people get confused. Or they accidentally removed existing apps when making their change due to sync issues.

Existing tests check that everything in apps.go is backed up by code in the repo.

This goes the other way and checks that what's in the repo is registered.

You can still quickly unregister existing apps if they start misbehaving. Just have to add the directory name to "exclusions" list and the test will ignore it.


betterengineering (approved):
This looks great! Thanks for the add!

PR #801, opened 2022-11-12
fix: Office Status Applet Beta Release Bugs and Refactor

b-bell:
A few clean-up tasks and bug fixes were found within the initial beta deployment of the office status applet. @matslina  The applet should be easier to maintain and deploy now as I've cleaned up how global variables are set for authentication.

1. Cleaned and simplified global variable declarations
2. Resolved a bug where all-day events from yesterday impacted today's availability
3. Improved the status message for cases where you are remote and busy all day


b-bell:
@matslina was there a change to the build-and-test check? It looks like some other apps in the branch are failing the check. The office status app runs properly on this branch in a local environment, it passes make lint, and then the build-and-test check fails due to errors from other apps.

My branch is up to date with main, so I have a feeling something changed in the check today.

b-bell:
@matslina thank you!

PR #802, opened 2022-11-12
feat: Add rotation speed option for NBA scores

mikeastock:


mikeastock:

mikeastock:
What's the process to get this merged @betterengineering or @matslina?

matslina:
@cmarkham20 Any thoughts on this one?

cmarkham20:
This defeats the purpose of installing the app multiple times and setting the Total Instances / Instance Number setting if you want it to go slower. That was a feature that a lot of people love and I think you'll get a lot of pushback (including from myself) if this gets approved. Plus it will break a lot of current installations.

jvivona:
i agree with @cmarkham20 - the multi-instance and having the math worked out for the user is the way to go.

mikeastock:
@cmarkham20 it defaults to the automatic rotation. This config is just for those that want just a single install but it to rotate much slower.

cmarkham20:
Even if you pick 30 seconds in your config, the Tidbyt will automatically switch apps every 15 seconds at the very slowest, and that is something you cannot change (hence the built-in 15 seconds). If you set it to 30 seconds, yes they will cycle more slowly, but you will only see half the games since it will cycle to the next app after 15 seconds. This is why I built in the functionality for separate app instances.

mikeastock:
@matslina what's the process for getting this merged? I feel like it's an addition that doesn't break existing flows and is more straightforward than having to do the multiple instance workaround. Can we have both?

cmarkham20:
The main issue I see is that if a user sees this setting and chooses the "30 seconds" option, the app is still only going to display for 15 seconds on the Tidbyt. Until Tidbyt makes it so you can adjust the app cycle time to be longer (which I see is potentially in the works), this will just cause more confusion for users.

mikeastock:
Gotcha. What if remove the 30 second option but leave everything else. Does that work?

cmarkham20:
I mean, I guess, but then it's not going any slower... the only option would be to make it go faster than normal, which has been the main complaint so far. I think we need to wait until Tidbyt makes it so a user can control cycle speeds before something like this is implemented.

matslina:
I think I agree with @cmarkham20 , but I totally see where you're coming from @mikeastock. We just recently started scoping out some potential changes on our end to enable longer display times for apps like these. I hope we'll be able to get something released before EOY, which hopefully provides as cleaner way of handling this issue.

betterengineering:
Thanks @mikeastock for the change! I'm going to close this PR for now, given there is not alignment. As @matslina we are looking to have architectural changes out sometime this year or early January that will provide better support for this style of information

PR #811, opened 2022-11-16
Sports Scores: Fix Basketball bug

rs7q5:
Had a small bug when getting the date in the basketball function, this PR fixes that so the app should work correctly now.

When I ran `make lint` it also modified the `sf_muni` app too for some reason. Just FYI.

### DIFF
--- a/apps/sfnextmuni/sf_next_muni.star
+++ b/apps/sfnextmuni/sf_next_muni.star
@@ -256,7 +256,7 @@ def fetch_cached(url, ttl):
             return (time.now().unix, {})
 
         # Trim off the UTF-8 byte-order mark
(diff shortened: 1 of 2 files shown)


apps/sfnextmuni/sf_next_muni.star
```diff
@@ -256,7 +256,7 @@ def fetch_cached(url, ttl):
             return (time.now().unix, {})
 
         # Trim off the UTF-8 byte-order mark
-        body = res.body().lstrip("\ufeff")
+        body = res.body().lstrip("\\ufeff")
```
matslina, line 259:
```suggestion
        body = res.body().lstrip("\ufeff")
```

PR #812, opened 2022-11-18
New App: BlueBomber -- randomly display one of the 8-bit Megaman bosses!

grex22:
This is my first app for Tidbyt (of many 🤞) and my first time writing in Go, but I'd love to share this app with the community!


grex22:
recheck

apps/bluebomber/bluebomber.star
```diff
@@ -0,0 +1,305 @@
...
+    "BLADE": "BLADE",
+    "SOLAR": "SOLAR",
+    "SHEEP": "SHEEP",
+    "COMMANDO": "COMMANDO",
+    "NITRO": "NITRO",
+    "STRIKE": "STRIKE",
+    "PUMP": "PUMP",
+}
+
+def main(config):
```
betterengineering, line 87: In order to make this usable to folks in the Tidbyt mobile app, we would need a `Dropdown` schema field to be able to select which boss and gender to display: https://tidbyt.dev/docs/reference/schema.

grex22: Good ideas! It wasn't really the intent of the app, however, it was just meant to be a random one displayed on each load (similar to the Nyan Cat app). If you think that'd be a good option for mobile users then I'm happy to add it! Gender isn't really a configurable option, as it's just there to catch the one case of "SPLASH WOMAN" being displayed.

There's I believe 64 boss options so do you feel a dropdown will be workable with that amount of choices? Not sure of a better option however.

betterengineering: Oh! My bad. I completely missed the randomized aspect of this change.

betterengineering (approved):

PR #813, opened 2022-11-18
Fix: Time zones for all day events

b-bell:
@matslina I'm sorry for the multiple bug fixes. :( I swear I'm testing, but I keep running into time zone issues. The code works properly until I try it at the end of the day. Then, I find a time zone conflict.

This should (safe harbor) be the last time zone bug fix.


matslina:
=) no worries

PR #815, opened 2022-11-18
Update verge taglines to work with non-hydrated home page

joevgreathead:
## What are you updating?

The Verge updated their site today and appear to be using hydration to render certain content only once it's loaded into a DOM. Since the HTTP module does not do anything beyond loading the raw text-based response body, I've had to update the logic here to parse out the tagline fro the underlying JSON blob.

This is just to fix the app because currently it displays a blank face in production.

## How are you doing this?

I'm targeting a script tag with an embedded JSON blob used for hydration and then iteratively checking the presence of the path I expect in that JSON lob since there doesn't appear to be a try-except/catch method supported in Starlark.

### DIFF
--- a/apps/vergetaglines/verge_taglines.star
+++ b/apps/vergetaglines/verge_taglines.star
@@ -8,6 +8,7 @@ Author: joevgreathead
 load("render.star", "render")
 load("http.star", "http")
 load("html.star", "html")
+load("encoding/json.star", "json")
 load("encoding/base64.star", "base64")
 load("cache.star", "cache")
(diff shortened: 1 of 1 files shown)


betterengineering:
> @betterengineering / @matslina, is there anything blocking this PR? This app has been broken since the day I submitted this fix.

Sorry for the delay @joevgreathead  😢. No issues on your side, we're just short on cycles.

joevgreathead:
No worries, @betterengineering. I know that feeling. Thanks for the quick response and g'luck.

PR #820, opened 2022-11-20
update: NASCAR next race

jvivona:
- rearranged layout to fit all series titles correctly 
- moved race name & track name from marquee to full text and used fade in / out logic to display data


jvivona:
i've got a different idea for the rotating text

jvivona:
reopening - this is currently the better way to handle it.   animation.transform with a stack isn't working the way I expected.  So let's go back this version.

jvivona:
@matslina#7592 I removed the app from my PR/branch and apps.go so I could build.   if you are merging - just see that I did it.

betterengineering (approved):

PR #822, opened 2022-11-20
Update SF Next Muni with service alerts from GTFS

juniormonkey:
Also update the encrypted API key, in case that helps fix the API issues. Manually using my unencrypted API key works locally, I don't know how to test the encrypted key...


apps/sfnextmuni/sf_next_muni.star
```diff
@@ -181,13 +163,33 @@ def get_schema():
                 default = "long",
                 options = formats,
             ),
-            schema.Dropdown(
-                id = "service_messages",
-                name = "Show service messages",
-                desc = "The lowest priority of service message to be displayed.",
+            schema.Toggle(
```
betterengineering, line 166: Note - this may cause unexpected behavior for current installations of this app. Any user who configured the drop down will now need to update their installation with the toggle.

betterengineering (approved):
Just an FYI, this app is decently popular:
```
app sf-next-muni has 83 installations and is installed on 59 devices
```

You'll want to be careful with schema changes in the future because it will change the existing installations and how they behave.

PR #829, opened 2022-11-21
Njtransitdepartureapp

Jason-J-Hunt:
App to fetch departures for a selected NJ Transit station


apps/njtransit/nj_transit.star
```diff
@@ -0,0 +1,374 @@
...
+
+def get_departures_for_station(station):
+    print("Getting departures for '%s'" % station)
+
+    station_suffix = station.replace(" ", "%20")
+    station_url = "{}/{}".format(NJ_TRANSIT_DV_URL, station_suffix)
+
+    print(station_url)
+
+    nj_dv_page_response = http.get(station_url)
```
betterengineering, line 184: Would you mind adding some caching here? Otherwise we would be hammering the NJ transit website if this app gets popular :scream_cat:

betterengineering (changes requested):
Thanks for the app @Jason-J-Hunt! It looks pretty good, we just need some caching around fetching departures to ensure we don't hammer the NJ transit website.

apps/njtransit/nj_transit.star
Jason-J-Hunt: Done added 1min ttl for website results.

betterengineering (approved):
Thanks for adding the cache @Jason-J-Hunt! LGTM

PR #830, opened 2022-11-21
New App: Apple TV (#817)

tjmehta:
Displays Apple TV's currently playing music, tv, video information on your Tidbyt.
Apple TV does not have Cloud APIs, so you will need to run a server locally to provide playing content information to Tidbyt. I wrote [tjmehta/apple-tv-now-playing-server](https://github.com/tjmehta/apple-tv-now-playing-server) on top of pyatv that will allow you to pair with your apple tv and read the "now playing" information and thumbnail. It has a Dockerfile so you can easily deploy it to any server running docker on your home network.


tjmehta:
recheck

tjmehta:
this should be good to go now

betterengineering (changes requested):
Hey @tjmehta! Thanks for the app! Unfortunately, we will not be able to support this app in it's current design. Our backend makes HTTP requests from Google Cloud and wouldn't have access to your local network. Even if you could open up the host to external requests, we don't currently support arbitrary URLs from user input and don't have a timeline on when we can support that. The only things we support at the moment are static, publicly facing URLs

tjmehta:
@betterengineering interesting... I assumed these tidbyt applications ran locally. I am curious, why does an application like this one require a backend at all? Is it not possible to run these applications directly on the tidbyt?

Thanks for your time. Last question, is there anyway (soon atleast) that we will be able to install locally running applications on the tidbyt?

tjmehta:
@betterengineering yeah I see how I could get it to work using the `tidbyt push` command but I am assuming rendering would interrupt other apps. 

I don't know the interface between the tidbyt and the official backend but it would be nice to be able to install "bring-your-own-backend" apps that at the very least mirror the tidbyt push behavior. The only piece that is missing is the tidbyt notifying a user server when it should push (without this you hit the interruption issue mentioned above).

If the tidbyt has any networking functionality at all, I think it would be cool to have an app that makes a request to request an image (even if the image is some custom format) from any server.

PR #836, opened 2022-11-26
Stupid Chat for Tidbyt

harrisonpage:
Send messages and pictures to your Tidbyt from anywhere.

Visit https://stupid.chat and click `Get Started`.  You will be assigned a random, unique token for your Tidbyt. Configure your stupid.chat token in the app to link the website to your device. Having done so, you may send messages or images from the stupid.chat website to your Tidbyt.  Tokens are free, you can always visit stupid.chat and generate a new one.

### DIFF
--- a/apps/stupidchat/stupid_chat.star
+++ b/apps/stupidchat/stupid_chat.star
@@ -1,0 +1,81 @@
+"""
+Applet: Stupid Chat
+Summary: Tidbyt Messaging
+Description: Send messages to your Tidbyt via https://stupid.chat or an API.
+Author: harrisonpage
+"""
+
+load("http.star", "http")
+load("render.star", "render")
+load("schema.star", "schema")
+load("encoding/base64.star", "base64")
+load("encoding/json.star", "json")
+
+VERSION = "1.0.0"
+
+def get_config(config, key, default):
+    value = config.get(key)
+    if not value:
+        value = default
+    return value
+
+def main(config):
+    token = get_config(config, "token", "WELCOME")
+    client_id = get_config(config, "client_id", "XXX")
+    response = http.get(
+        "https://api.stupid.chat/v1/" + token + "?client_id=" + client_id + "&format=base64&version=" + VERSION,
+        headers = {"User-Agent": "stupid.chat/" + VERSION},
+    )
+    if response.status_code != 200:
+        fail("failed with status %d", response.status_code)
+    cooked = response.json()
+
+    if cooked["upload"]:
+        return render.Root(
+            render.Image(
+                src = base64.decode(cooked["image"]),
+                width = 64,
+                height = 32,
+            ),
+        )
+
+    return render.Root(
+        child = render.Column(
+            children = [
+                render.Box(
+                    color = "#" + cooked["bgcolor"],
+                    child = render.Marquee(
+                        offset_start = 32,
+                        offset_end = 32,
+                        width = 64,
+                        height = 32,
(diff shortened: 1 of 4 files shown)


harrisonpage:
recheck

harrisonpage:

apps/stupidchat/stupid_chat.star
```diff
@@ -0,0 +1,81 @@
...
+
+def main(config):
+    token = get_config(config, "token", "WELCOME")
+    client_id = get_config(config, "client_id", "XXX")
+    response = http.get(
+        "https://api.stupid.chat/v1/" + token + "?client_id=" + client_id + "&format=base64&version=" + VERSION,
+        headers = {"User-Agent": "stupid.chat/" + VERSION},
+    )
+    if response.status_code != 200:
+        fail("failed with status %d", response.status_code)
```
matslina, line 30:
```suggestion
        fail("failed with status %d" % response.status_code)
```

apps/stupidchat/stupid_chat.star
```diff
@@ -0,0 +1,81 @@
...
+VERSION = "1.0.0"
+
+def get_config(config, key, default):
+    value = config.get(key)
+    if not value:
+        value = default
+    return value
+
+def main(config):
+    token = get_config(config, "token", "WELCOME")
```
matslina, line 23: The config (dict like) object can handle the default case for you:
```suggestion
    token = config.get("token", "WELCOME")
```

apps/stupidchat/stupid_chat.star
```diff
@@ -0,0 +1,81 @@
...
+    client_id = get_config(config, "client_id", "XXX")
```
matslina, line 24: I take it this is for debug purposes? Just wanted to mention since it's not in the schema.

apps/stupidchat/stupidchat.go
```diff
@@ -0,0 +1,25 @@
...
+//go:embed stupid_chat.star
+var source []byte
+
+// New creates a new instance of the Stupid Chat applet.
+func New() manifest.Manifest {
+	return manifest.Manifest{
+		ID:          "stupid-chat",
+		Name:        "Stupid Chat",
+		Author:      "harrisonpage",
+		Summary:     "Tidbyt Messaging",
```
matslina, line 19: If you don't mind it'd be great to rephrase this summary a little bit, just to highlight to users that this an external service not provided by Tidbyt. Perhaps something like `"Send messages to your Tidbyt"`?

apps/stupidchat/stupidchat.go
```diff
@@ -0,0 +1,25 @@
...
+		Desc:        "Send messages to your Tidbyt via https://stupid.chat or an API.",
```
matslina, line 20: Could you also add a note in the description highlighting that the Stupid Chat service is external and not run by Tidbyt? You can see a similar example in the "Textbyt" app.

matslina (changes requested):
Thanks for the PR! This looks great, and I'm sure folks will enjoy it a lot! =)

A few comments that needs to be addressed before we merge. Also wanted to mention that I've seen a lot of 500s when testing this out locally. Both with the default token and a new one I created on the site. Might be good to take a look at that, in case there something that needs fixing on your end for this to work more reliably.

apps/stupidchat/stupidchat.go
matslina:
```suggestion
		Summary:     "Send messages to your Tidbyt",
```

PR #841, opened 2022-11-29
rename todaysname to swedishnameday

matslina:


matslina:
@y34752 FYI. This updates the directory and file names to the name change, and also nudges the name a bit further to get it below the 17 character limit. LMK if you're not happy with the change and we can hold off from publishing it. Otherwise it'll go out later this week.

PR #844, opened 2022-11-30
Visible Planets fix

rjison:
Change display during the day for outer planets. Now show the visibility at sunset always.
Change display of large distances to something more readable (5.4 billion vs 5,400,000,000)


apps/visibleplanets/visibleplanets.star
```diff
@@ -363,9 +394,9 @@ def get_schema():
                 id = "planet",
                 name = "Planet",
                 desc = "Pick the planet you want to track.",
-                icon = "satelliteDish",  #skyatlas #moon, globe, mars, marsandvenus satelliteDish
+                icon = "globe",  #skyatlas #moon, globe, mars, marsandvenus satelliteDish
                 options = planet_options,
-                default = planet_options[0].value,
+                default = planet_options[5].value,
```
betterengineering, line 399: Just an FYI, anyone with this app installed with default values will see this change once this app is updated in our backend

betterengineering (approved):

PR #846, opened 2022-11-30
Add Shopify Order Moments and Shopify Animations

markmcintyre:
This PR adds two new Shopify Apps: **Order Moments** and **Shopify Animation**

Order moments shows a celebration animation when specific order milestones have been reached.
Shopify Animation is a totally useless, but enjoyable animation of Shoppy, the Shopify mascot!

Screenshot 2022-11-30 at 13 45 11

Screenshot 2022-11-30 at 13 48 24


apps/ordermoments/order_moments.star
```diff
@@ -0,0 +1,749 @@
...
+    (5000, 500),
+    (10000, 1000),
+    (100000, 5000),
+    (500000, 10000),
+    (1000000, 100000),
+]
+
+# MAIN
+# ----
+def main(config):
```
betterengineering, line 55: One thing to note is that we will call this app to render previews inside of the Tidbyt Mobile app. You should consider adding some default view for a preview so that users browsing on the explore tab or using search can find a representative preview

apps/ordermoments/order_moments.star
```diff
@@ -0,0 +1,749 @@
...
+
+# Get Schema
+# Return a Pixlet Schema for this Celebrate Applet
+# -----------------------------------------------------------------------------------------
+# Returns: A Pixlet schema
+def get_schema():
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Text(
```
betterengineering, line 331: Any chance this could be OAuth2? If you're installing this app through the API, you can still override the config value. The benefit is that existing users of Tidbyt who have a Shopify store can discover this app and install it without having to jump through hoops setting up their API key.

betterengineering (approved):
Going to get this into a beta so you can take a look awhile. I'd definitely consider previews and OAuth2 to keep the experience slick from the Tidbyt side.

apps/ordermoments/order_moments.star
markmcintyre: Good point! Is there a config value that's passed to the app for preview rendering?

PR #847, opened 2022-12-01
Earthquake Map Application

SpinStabilized:
Displays global earthquake data for, up to, the last 30 days. Applet preferences allow users to filter the display on a minimum earthquake magnitude and different time scales. Additionally, the user can select the central meridian of the map to be on the Prime Merdian, International Date Line, or the user's home location.

earthquake_map


betterengineering (approved):
Looks great! Thanks @SpinStabilized!

PR #850, opened 2022-12-03
Update default airports in METAR app

avalys:
Updating the default setting to show multiple airports so the purpose of the app is clearer when previewed in the store. Also clarified the description.


apps/metar/metar.star
```diff
@@ -6,7 +6,7 @@ load("secret.star", "secret")
 load("schema.star", "schema")
 
 ADDS_URL = "https://www.aviationweather.gov/adds/dataserver_current/httpparam?dataSource=metars&requestType=retrieve&format=csv&stationString=%s&mostrecentforeachstation=constraint&hoursBeforeNow=2"
-DEFAULT_AIRPORT = "KJFK"
+DEFAULT_AIRPORT = "KJFK, KLGA, KBOS, KDCA"
```
betterengineering, line 9: FYI - anyone who installed this with the default settings would have their installation change once this is live.

betterengineering (approved):

apps/metar/metar.star
avalys: Oh, interesting. Unless someone is specifically interested in JFK I'm sure they changed it.

The reason I'm changing it is that the image shown in the Tidbyt app seems to use the default settings and I think the view with multiple airports is a better representation of the app for new uers.

PR #788, opened 2022-11-10
Update shabbat app

NudnikShpilkis:
Update the shabbat app for Tidbyt based on discussion with @dinosaursrarr .

PR #858, opened 2022-12-05
Amended shabbat times with with new end methods

NudnikShpilkis:
Amended shabbat app with new end methods.


NudnikShpilkis:
@betterengineering I made a new PR rather than update #788 .

apps/shabbat/shabbat.star
```diff
@@ -1,8 +1,8 @@
 """
-Applet: Shabbat
+Applet: Shabbat Times
```
betterengineering, line 2: You will also want to update `shabbot.go` if you want these to update in the mobile app

apps/shabbat/shabbat.star
```diff
@@ -15,155 +15,352 @@ load("time.star", "time")
...
             schema.Dropdown(
                 id = "end_method",
-                name = "End time",
-                desc = "How to calculate the end time of Shabbat",
-                icon = "hourglass",
-                default = end_method_options[0].value,
+                name = "Method",
+                desc = "Method to calculate the end of shabbat",
+                icon = "moon",
+                default = end_method_options[1].value,
```
betterengineering, line 348: This app currently has:
```
app shabbat has 67 installations and is installed on 67 devices
```

If you update this default, it will cause anyone who installed this app with the default to get the new setting instead. Is that desired?

betterengineering (approved):

apps/shabbat/shabbat.star
NudnikShpilkis: That's a good point.  I don't want to get into too many halakhic arguments.  I'll allow for the old default.

PR #859, opened 2022-12-05
Change NS Error Cache to 30 mins

IsThisPaul:
This cache is used only when Nightscout returns an error. The cache ensures a blood sugar value is shown (and displayed as stale) for up to 30 minutes if NS is temporarily unavailable. It does not affect live glucose values if they are available as this app never displays cached data unless there is a NS error.

### DIFF
--- a/apps/nightscout/nightscout.star
+++ b/apps/nightscout/nightscout.star
@@ -38,7 +38,7 @@ GRAPH_WIDTH = 43
 GRAPH_BOTTOM = 50
 GRAPH_TOP = 275
 
-CACHE_TTL_SECONDS = 300  #5 mins
+CACHE_TTL_SECONDS = 1800  #30 mins
 
 PROVIDER_CACHE_TTL = 7200  #2 hours
 NS_PROVIDERS = "https://gist.githubusercontent.com/IsThisPaul/e976c41112d79383c51ed7315bb114ab/raw/170f967eba27587ffd289631c1a2ae9e31f1bee3/nightscout_providers.csv"


betterengineering:
Hey @IsThisPaul! Totally understood that this only happens if there was an error. The issue though is we're never telling the user that there was an error and are continuing to display stale data. How fast do glucose levels fluctuate? I don't want a Nightscout user to see their Tidbyt and think glucose levels are fine when reality is, our app is erroring because their nightscout setup is erroring. Ultimately, this app will be used to make healthcare decisions and I really want to make sure we're not causing someone to make poor healthcare decisions by displaying stale data. Let me know if I'm completely off base, more then happy to discuss or reconsider.

IsThisPaul:
It's not unusual to display stale data as long as it's clearly labeled stale. If a user sees a glucose number and a trend but knows it's 25 minutes old, they still have some idea what's happening. Generally, too, users would not be using the Tidbyt display to make healthcare decisions (they would always verify data with their CGM app and/or finger prick blood test). 
I'd be open to displaying "NS Error" in the reading age display if the reading is older than 5 mins but then the user has no idea how stale the data is. 
It's honestly an edge case that NS would error for longer than 5 mins but it happens. If it happens for longer than 30, they need to go troubleshoot NS. 
TLDR; Stale readings with clear age displayed is better than no reason due to minor outage.

IsThisPaul:
Happy to get on a quick call to discuss and resolve if needed.

IsThisPaul:
Oh. To answer your question... typical NS glucose readings are updated from the glucose monitor every 5 minutes.

betterengineering:
ok! Happy to merge if you find this is better for the end user. Thanks again for your improvements to this app!

IsThisPaul:
Thanks, Mark. I'm excited to see this go live.

PR #863, opened 2022-12-06
Added new applet for tracking portfolio value of a stock

gshipley:
I created a simple applet that will track the value of a particular stock that is owned. I mainly coded this to track my RSU value on my office tidbyt to constantly motivate me to work hard.

The user must provide a ticker symbol, number of shares owned, and a stock API key.

I set the cache value to 12 hours as the app is intended for a daily overview, not a real time account value. 

I am a bit concerned about the unfriendliness of the user having to provide an API key but it will default to mine if not provided. I have searched high and low for a unauthenticated simple stock quote service.

### DIFF
--- a/apps/stockvalue/stock_value.star
+++ b/apps/stockvalue/stock_value.star
@@ -1,0 +1,92 @@
+"""
+Applet: Stock Value
+Summary: Portfolio value
+Description: This app will allow you track the value of your portfolio for a single stock.
+Author: gshipley
+"""
+
+load("render.star", "render")
+load("http.star", "http")
+load("math.star", "math")
+load("humanize.star", "humanize")
+load("encoding/base64.star", "base64")
+load("schema.star", "schema")
+load("cache.star", "cache")
+load("secret.star", "secret")
+
+STOCK_PRICE_URL = "https://www.alphavantage.co/query?function=GLOBAL_QUOTE&symbol="
+DEFAULT_SHARES = 1
+DEFAULT_SYMBOL = "IBM"
+
+DOLLAR_ICON = base64.decode("""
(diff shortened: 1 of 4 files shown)


apps/stockvalue/stock_value.star
```diff
@@ -0,0 +1,92 @@
...
+            schema.Text(
+                id = "shares",
+                name = "Number of shares",
+                desc = "How many shares do you have?",
+                icon = "hashtag",
+            ),
+            schema.Text(
+                id = "alphavantage",
+                name = "API KEY",
+                desc = "API key for Alpha Vantage (https://www.alphavantage.co)",
```
matslina, line 88: The `desc` field is unfortunately currently not visible to end users in the iOS/Android app, so it would be great if you could change the field `name` to explain that the key is for Alpha Vantage.

matslina (commented):
Hey, thanks for the PR! Looks great overall. Just this one small change and we should be able to merge it.

apps/stockvalue/stockvalue.go
```diff
@@ -0,0 +1,25 @@
...
+var source []byte
+
+// New creates a new instance of the Stock Value applet.
+func New() manifest.Manifest {
+	return manifest.Manifest{
+		ID:          "stock-value",
+		Name:        "Stock Value",
+		Author:      "gshipley",
+		Summary:     "Portfolio value",
+		Desc:        "This app will allow you track the value of your portfolio for a single stock.",
```
betterengineering, line 20:
```suggestion
		Desc:        "This app will allow you track the value of your portfolio for a single stock. Get your API key for Alpha Vantage at https://www.alphavantage.co.",
```

apps/stockvalue/stock_value.star
betterengineering: @matslina I added a note to the app description to move this app along

betterengineering (approved):

PR #876, opened 2022-12-09
Github Status Badge App

Cavallando:
Hi!

I really wanted to be able to fetch the `badge.svg` from github but it appears that there is an issue with private repos where it can't be fetched. So I opted to write this using personal access tokens and GitHubs Rest API.

Here is a gif example showing the marquee and default preview for the app:

github_badge


matslina:
Nice! Thanks for the PR!

PR #877, opened 2022-12-09
Mastodon followers

meejle:
I've literally never done a pull request before. I'd never written a line of Python before, um... Monday? If you need a "here's what an inexperienced developer can do in just five days, before their first Tidbyt even arrived" testimonial, there you go.

(Actually, one day was taken up trying to get "make" to run on Windows 11, so the coding was really only four days.)


meejle:
recheck

meejle:

meejle:
recheck

mattgodbolt:
Very much interested in this now - thanks for posting @meejle

apps/mastodoncounter/mastodon_counter.star
```diff
@@ -0,0 +1,222 @@
...
+                            ),
+                        ],  #children
+                    ),  #render.Row
+                ),  #render.Box
+            ],  #children
+        ),  #render.Column
+    )  #render.Root
+
+def main(config):
+    theapi = "https://%s/api/v1/accounts/verify_credentials" % config.str("instance", DEFAULT_SERVER)
```
matslina, line 109: We can unfortunately not merge apps that perform http calls to arbitrary hosts. This is so to protect against our backend being abused to spam or DoS others.

Would it be possible to replace the `instance` field with perhaps a dropdown of popular mastodon instances instead?

apps/mastodoncounter/mastodon_counter.star
```diff
@@ -0,0 +1,222 @@
...
+        fields = [
+            schema.Text(
+                id = "instance",
+                name = "Your Mastodon instance",
+                desc = "The Mastodon server you use, e.g. ohai.social or mastodon.social",
+                icon = "mastodon",
+            ),
+            schema.Text(
+                id = "apikey",
+                name = "Your access token",
```
matslina, line 205: The `desc` field is currently not visible to users of the Tidbyt iOS/Android app, so be aware that any information users will need to set up the app has to be in the `name` field, or in the app description (in `mastodoncounter.go`).

matslina (changes requested):
Hey, thanks for the PR! This looks really great overall. There are a few issues we need to address before merging.

matslina:
Btw the most recent build error is due to this app not being registered in `apps/apps.go`. If you take a look there and add `mastodonfollowers` in the same fashion as all other apps you should make it past this particular error.

apps/mastodoncounter/mastodon_counter.star
meejle: There are apparently 13,394 Mastodon instances right now, so... probably not? 😅 Not without having to exclude a bunch of people and/or do _constant_ support haha.

Your other two suggestions seem perfectly do-able but I can't think of a good way of solving this one. I'll mull it over for a while.

betterengineering: Sorry for the back and forth @meejle! One thing that might be useful is https://instances.social/. They have an API here: https://instances.social/api/doc/. It might help get around the problem of arbitrary URLs

drudge: @meejle I did integrate with instances.social in this PR: https://github.com/tidbyt/community/pull/972

meejle:
I like @drudge's applet _way_ more than I like mine, ima let this one die and focus on other stuff. 😂 Thanks everyone for your help!

PR #881, opened 2022-12-12
Fix timezone off-by-one error in Is It Xmas applet

sloanesturz:
First, I love this applet and I loved discovering it in the tidbyt app! I am hoping seasonal things like this continue to make appearances.

There is a small bug in this applet where it uses a UTC date comparison instead of the local timezone. That means that in the evening west coast time today 2022-12-11, it shows "13 days until Christmas" instead of "14 days until Christmas".

The issue is that 
```
xmas_timestamp = time.time(year = today.year + 1, month = 12, day = 25).in_location(timezone)
```

Should be
```
xmas_timestamp = time.time(year = today.year + 1, month = 12, day = 25, location = timezone)
```

The former initializes a UTC 2022-12-25 and then translates it into PST - which is 2022-12-24. The latter creates a reference to 2022-12-25 in PST.

I also cleaned up some readability issues (I think the 26 on line 25 and 28 was patching around this issue).

@rohansingh @mgodfrey10 FYI!

### DIFF
--- a/apps/daystoxmas/days_to_xmas.star
+++ b/apps/daystoxmas/days_to_xmas.star
@@ -20,18 +20,15 @@ def main(config):
     today = time.time(year = now.year, month = now.month, day = now.day, location = timezone)
     current_xmas = time.time(year = today.year, month = 12, day = 25, location = timezone)
 
-    if today <= current_xmas:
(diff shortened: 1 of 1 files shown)


betterengineering (approved):
Thanks for the fix! Really wish we had a unit testing framework right about now 😅

PR #884, opened 2022-12-13
Soccer standings

M0ntyP:
Hi,

This is my first app!

I've created an app to display league tables from various soccer leagues around Europe and the world. Shows either a black & white option, or one with background colors for the teams. But doesn't always display nicely if the team color is yellow

PS - this is my 2nd pull request, I broke the first one :(


matslina:
Hey @M0ntyP, thanks for the PR! Build is breaking due to the app not being registered in `apps/apps.go`. Lint also appears to be failing due to a US vs UK spelling issue.

PR #890, opened 2022-12-16
feat: Create Hex Color Clock App

gabe565:
## Description
This PR adds a hex color clock, similar to [hex clock](https://www.jacopocolo.com/hexclock/) and [The Colour Clock](http://thecolourclock.com).

The hour, minute, and second portions of the current time are used to generate a hex code with hour being red, minute being blue, and second being green.  
The foreground color will change between white and black to be as visible as possible against the background.

## Screenshots
By default, it will scale the current hour, minute, and second to fill the full hex range:
hex_color_clock

But it can be configured to output the literal hex code of the current time by disabling "Use more colors":
hex_color_clock

## Supported Config
1. Location
2. Toggle: **Hide Clock**
4. Toggle: **24 Hour Format**
   - Defaults to 24 hours to match the hex codes more closely.
5. Toggle: **Hide Seconds**
   - I know seconds won't always be accurate, but it's more fun to see the seconds tick up along with the hex code.
6. Toggle: **Flashing Separator**
7. Toggle: **Hide Hex Code**
8. Toggle: **Use More Colors**
   - This is the config that toggles between the screenshots above. Would love suggestions on the name for this one. I also considered "use brighter colors" or "use full color range".


gabe565:

matslina:
Thanks for the PR!

PR #891, opened 2022-12-16
fix(nightscout): Remove trend variable

gabe565:
This PR removes the `trend` variable from Nightscout. This variable was not used during rendering, and not all CGM sources seem to set this variable so it could cause errors during render.

For example, here's an entry from my Nightscout `entries.json`:
```json
  {
    "_id": "639cc6bde169cee86173fe72",
    "device": "xDrip-DexcomG5",
    "date": 1671218875550,
    "dateString": "2022-12-16T19:27:55.550Z",
    "sgv": 139,
    "delta": -2,
    "direction": "Flat",
    "type": "sgv",
    "filtered": 0,
    "unfiltered": 0,
    "rssi": 100,
    "noise": 1,
    "sysTime": "2022-12-16T19:27:55.550Z",
    "utcOffset": -360,
    "mills": 1671218875550
  },
```

### DIFF
--- a/apps/nightscout/nightscout.star
+++ b/apps/nightscout/nightscout.star
@@ -106,7 +106,6 @@ def main(config):
     sgv_current = int(nightscout_data_json["sgv_current"])
     sgv_delta = int(nightscout_data_json["sgv_delta"])
     latest_reading_dt = time.parse_time(nightscout_data_json["latest_reading_date_string"])
-    trend = nightscout_data_json["trend"]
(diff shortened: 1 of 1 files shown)


IsThisPaul:
This looks right to me. "trend" is not available in all CGM sources and is not used by the script. "direction" is used instead.

IsThisPaul:
This issue is actually causing some installs to not work at all. This should be a priority.

PR #892, opened 2022-12-16
feat(nightscout): Add vertical axis bars every hour

gabe565:
This PR adds bars to the vertical axis at the beginning of every hour. The hour bars are subtle, but they add a lot of context at a glance!

nightscout


IsThisPaul:
This should be available as an option in the config schema if "show graph" is toggled on. Add the "show hours on graph" config schema logic, then wrap this new code in an if to display only when the option is selected.

gabe565:
@IsThisPaul Done!

PR #895, opened 2022-12-19
feat: Add starfield

gabe565:
## Description

This PR adds an app that simulates flying through a starfield. This one looks so cool on an actual device! The gif doesn't do it justice.

The generated webp is 20fps for 15 seconds and is typically ~85 KB, but if star count and tail length both get raised to higher values, the output can be larger than 128 KB, causing some frame drops. I limited these to try to keep frame drops to a minimum, but with the current maximums, the output can max at ~170 KB. Let me know if you'd like me to further restrict these or change anything else to keep file size down!

I also heavily optimized this one. Current profile output takes 280ms. Is this acceptable?

## Screenshot
starfield

## Schema

1. Dropdown: Background Color
2. Dropdown: Star Color
3. Toggle: Use Custom Star Color?
4. Text: Custom Star Color
5. Dropdown: Number of Stars
6. Dropdown: Star Speed
7. Dropdown: Star Tail Length


btjones:
I had been thinking that some old-school screen savers could work well and this looks really cool. Can't wait to try it out in my actual TidByt!

image

gabe565:
@btjones Haha, thank you! I recently purchased my Tidbyt and pretty quickly started thinking the same thing. I plan to add more, so feel free to ping me on the Discord or Discourse if you have any other ideas!

btjones:
I had a couple thoughts on colors options that could be interesting: 
- "random" which would select a different random color every time the app runs
- "rainbow" which would have each star be a different color

gabe565:
> I had a couple thoughts on colors options that could be interesting:
> 
> * "random" which would select a different random color every time the app runs
> * "rainbow" which would have each star be a different color

@btjones Great ideas! I've added both options.
- Random will choose a random color each time the starfield is rendered
- Rainbow will choose a random color for each star. It's pretty chaotic, but I think I like it? Either way, doesn't hurt to have 😂 

rainbow starfield

gabe565:
Now that it supports multiple colors at once, it might be cool to add color schemes? Or let users enter multiple comma-separated colors for the custom colors? I don't want to make the settings overwhelming, but that could be cool.

gabe565:
I went ahead and added support for palettes, along with adding a Christmas palette! The Christmas palette doesn't _really_ look like a starfield, but it still looks cool 🤷

starfield

btjones:
@gabe565 amazing.

apps/starfield/starfield.star
```diff
@@ -0,0 +1,413 @@
...
+Description: This app simulates flying through a starfield.
+Author: gabe565
+"""
+
+load("math.star", "math")
+load("random.star", "random")
+load("re.star", "re")
+load("render.star", "render")
+load("schema.star", "schema")
+
```
matslina, line 13:
```suggestion
load("time.star", "time")
```

apps/starfield/starfield.star
```diff
@@ -0,0 +1,413 @@
...
+        value = "1.3",
+    ),
+    schema.Option(
+        display = "Fastest",
+        value = "2",
+    ),
+]
+
+def main(config):
+    star_color = get_palette(config)
```
matslina, line 174:
```suggestion
    random.seed(time.now().unix // 10)
    star_color = get_palette(config)
```

matslina (changes requested):
Thanks for the PR! Animation is gorgeous and performance is great!

That said, we do rely heavily on caching of generated webps in our backend, and with completely random graphics on each execution we'll definitely be running into some problems there. I added suggestions intended to make the app deterministic over 10s time windows, although I suspect a longer time window may be totally fine. WDYT?

gabe565:
@matslina Thanks for the feedback! I attempted to optimize as much as possible, but was unclear on some of the MQTT image push details. Does the backend continually regenerate the image and skip the MQTT push if the image is unchanged? If so, then I agree that an even longer window would make sense.

Only thing to be aware of in this case is since each star gets placed randomly, sometimes a burst of stars can happen in a similar area. These look cool, but if the same burst happens too many times in a loop, then it starts to become obvious. Otherwise, I'm not opposed to a longer timespan that will give the same deterministic output!

I just pushed a commit that adds your suggestions. Let me know what you think!

matslina:
> @matslina Thanks for the feedback! I attempted to optimize as much as possible, but was unclear on some of the MQTT image push details. Does the backend continually regenerate the image and skip the MQTT push if the image is unchanged? If so, then I agree that an even longer window would make sense.

There are a few layers of caching. Devices have a local cache of graphics. The backend also checks the widget tree against a cache before drawing and encoding webps. This change should be especially helpful for the latter of the two.

PR #897, opened 2022-12-20
Increasing the amount of time the scores appear

Jwells2244:
Scores were flashing by too fast beforehand, and this improves the overall visual experience.


matslina:
@cmarkham20

cmarkham20:
@Jwells2244 - Please see the following thread: https://github.com/tidbyt/community/pull/802#event-7955697940

TLDR; The Tidbyt cycles between apps at a maximum of 15 seconds, so changing this will only cut off scores from being shown and not affect the time it stays on the device.

PR #903, opened 2022-12-22
Lord of the Rings random quote generator app

jmanske:
Displays a randomly chosen quote from the original Lord of the Rings trilogy.

Either choose a character or let the character be chosen for you at random. Quote will refresh every 10 minutes.

Uses [The One API](https://the-one-api.dev) for getting the quotes.

screenshot


matslina:
Wonderful!

PR #908, opened 2022-12-24
New App: GitHub repo

rs7q5:
Display stats for a GitHub repository.

the auth key is just done by creating a personal access token (classic) and the dev key is set to never expire currently. Should probably be swapped to a production key.

One big improvement that is necessary is the consistency of the icons and color, but I'm trying to figure out sourcing these with consistency.

Thanks to @Cavallando for the code for the GitHub icon keys from their PR in #876 , really saved me some time. 

github_repo


rs7q5:
Updated preview here.

github_repo

Cavallando:
@rs7q5 awesome! I love all of the stats you managed to include! Thanks for the shoutout 😄

rs7q5:
I appreciate you saving me some time with the badge logic. I ended up changing the icons colors a little, but great work. Your app was spot on in the design

Issue #781, opened 2022-11-09
Cryptotracker app does not display correct prices


jsakas:
Heads up @ethanfuerst it looks like the app is not displaying correct prices.

Seems like its frozen in time... currently stuck on BTC at $21,214 -0.43% but as of this morning BTC is trading around $17k.

ethanfuerst:
Good catch, will look in to this today

ethanfuerst:
@jsakas Got a chance to dig in to this - Looks like the [API call](https://www.alphavantage.co/documentation/#crypto-intraday) that I used to get coin prices was moved to the premium tier and the premium rates are expensive and allow way more calls than I need. I emailed them to see if they can get me on a lesser plan for cheaper, and will keep this thread updated! Thanks for opening this issue, hopefully I can get this set up soon!

ethanfuerst:
Also @rohansingh @betterengineering could one of y'all re-open this issue? I'll keep this thread updated

ethanfuerst:
PR ready!
https://github.com/tidbyt/community/pull/909

PR #909, opened 2022-12-25
Fixing bug with Crypto Tracker app and adding Please Stand By app

ethanfuerst:
Hey all! 2 things here:

1. Updating the API key for the Crypto Tracker App to one with extra request permissions. Now that the key has permission to do 150 requests per minute instead of 500 requests a day, I'm going to start work on adding functionality for coins worth less than 10 cents and show their price changes when it changes less than a cent. [Issue here](https://github.com/tidbyt/community/issues/781).

2. Creating a new app called Please Stand By, which displays a scrolling Please Stand By message over the classic multi-colored background. This app doesn't call any APIs and is very lightweight.

Thanks!


matslina:
Thanks for the PR!

PR #912, opened 2022-12-26
New App: hadisplay

ebmichel:
Added haDisplay, an app that displays reading from any [Home Assistant](https://www.home-assistant.io/) device on a local network.

Displays the value of a device from Home Assistant on your Tidbyt with optional graphing of historic values from the past 24 hours. 

Connects to your local Home Assistant setup via REST API and is authenticated using a Long-Lived Access Token generated by the user.

No data is cached as the API calls are to self hosted servers, and the data is unique for each individual user


ebmichel:

PR #916, opened 2022-12-26
Is it Christmas? Fixed bug in calculating christmas

### LINKED ISSUE #914: isItChristmas - bug with displaying days left until Christmas
@austinrfnd today (dec 26) the app is showing -1 days until Christmas with "display days left" enabled.   
image

rs7q5:
Fixed bug in calculating christmas by removing the if statement where a year was added only if the month was greater, not month and day (think year was getting added twice then).

Also set a default value for config.get("christmas_date") so don't get a None value when rendering locally

Should fix #914.

cc original author @austinrfnd.

edit: again not sure why njtransit and sf_muni got edited. occurred when I did make lint.

### DIFF
--- a/apps/isitchristmas/is_it_christmas.star
+++ b/apps/isitchristmas/is_it_christmas.star
@@ -103,7 +103,7 @@ def main(config):
     )
 
 def get_daysleft(today, timezone, christmas_month, christmas_day):
-    if today.month == christmas_month and today.day > christmas_day:
+    if today.month >= christmas_month and today.day > christmas_day:
         year = today.year + 1
     else:
         year = today.year
--- a/apps/isitchristmas/is_it_christmas.star
+++ b/apps/isitchristmas/is_it_christmas.star
@@ -50,7 +50,7 @@ def main(config):
     day = now.day
     month = now.month
 
-    christmas_date = config.get("christmas_date")
+    christmas_date = config.get("christmas_date", "12-25")
     christmas_month = int(christmas_date[:2])
     christmas_day = int(christmas_date[3:])
 
@@ -104,8 +104,6 @@ def main(config):
 
 def get_daysleft(today, timezone, christmas_month, christmas_day):
     if today.month == christmas_month and today.day > christmas_day:
-        year = today.year + 1
-    if today.month > christmas_month:
         year = today.year + 1
     else:
         year = today.year
--- a/apps/isitchristmas/is_it_christmas.star
+++ b/apps/isitchristmas/is_it_christmas.star
@@ -103,7 +103,9 @@ def main(config):
     )
 
 def get_daysleft(today, timezone, christmas_month, christmas_day):
-    if today.month >= christmas_month and today.day > christmas_day:
+    if today.month == christmas_month and today.day > christmas_day:
+        year = today.year + 1
+    elif today.month > christmas_month:
         year = today.year + 1
     else:
         year = today.year
--- a/apps/njtransit/nj_transit.star
+++ b/apps/njtransit/nj_transit.star
@@ -276,7 +276,7 @@ Function gets the destation froma  given depature
 
 def get_destination_name(data):
     nodes = data.find(".media-body").first().find(".mb-0")
-    destination_name = nodes.eq(0).text().strip().replace("\\u2708", "EWR").upper()
+    destination_name = nodes.eq(0).text().strip().replace("\u2708", "EWR").upper()
     return destination_name
 
 """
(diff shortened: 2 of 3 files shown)


apps/isitchristmas/is_it_christmas.star
```diff
@@ -50,7 +50,7 @@ def main(config):
     day = now.day
     month = now.month
 
-    christmas_date = config.get("christmas_date")
+    christmas_date = config.get("christmas_date", "12-25")
```
Yonodactyl, line 53: Im not sure how Tidbyt handles the dates for international users, but this kind of makes me feel like we shouldnt be hardcoding a date here?

For example, a user in Russia might celebrate Christmas on January 7 - so this hardcoded date would be incorrect.

Maybe I am misunderstanding how dates work here?

rs7q5: There is a config for different dates. This is just a default value.

apps/isitchristmas/is_it_christmas.star
```diff
@@ -105,8 +105,6 @@ def main(config):
 def get_daysleft(today, timezone, christmas_month, christmas_day):
     if today.month == christmas_month and today.day > christmas_day:
         year = today.year + 1
-    if today.month > christmas_month:
```
dinosaursrarr, line 108: Not sure this makes sense.

Let's say Christmas is 6 Jan, and today is February 2023. In that case, you need to add another year because we'll be counting down 11 months to 6 Jan 2024.

If Christmas is December, then this can't be reached anyway because it can't be a greater month than December.

rs7q5: You are correct, I only checked christmas in December.... I don't think both need to be there, or it needs to be a elif. Let me fix that. 

That's what I get for not doing proper edge checking...

Think I fixed it with the latest commit, the today.month needed >= in order to roll over correctly. Since any time the date has passed a year needs to be added. 

```
    if today.month >= christmas_month and today.day > christmas_day:
        year = today.year + 1
    else:
        year = today.year
```

dinosaursrarr: I see the issue you're fixing, but I think it needs to be something like this, right?

```
    if today.month == christmas_month and today.day > christmas_day:
        year = today.year + 1  # During month of Christmas, but after Christmas has passed, so look to next year
    elif today.month > christmas_month:
        year = today.year + 1  # Christmas is over this year, so count to next year
    else:
        year = today.year  # Christmas is still to come this year
```

Suppose you use `today.month >= christmas_month and today.day > christmas_day` and Christmas is 6 January. Then on 2 February 2023, you'd end up counting down negative days to 6 Jan 2023, instead of forward to 6 Jan 2024.

I did it this way because the day of the month only matters during the month of Christmas. After that, you don't care about days, you can just do it all with months.

But you're right that it should have been an elif.

rs7q5: Darn it, yeah, you're right. I was wondering when that elif edge case was gonna need to be present. 

Originally, a year was getting added twice anyways I think because it was merely if and then another if else statement, which was wrong in and of itself. 

The actual error originally then, which I thought was going to be the original fix was changing that to elif, and then I thought I could eliminate that case... 

Anyways I've implemented that simple change back in. Thanks for the sanity check!

austinrfnd:
Tested this out locally and looks to be good.  Thank you so much for looking into this and fixing it!

austinrfnd (approved):

apps/sfnextmuni/sf_next_muni.star
```diff
@@ -269,7 +269,7 @@ def fetch_cached(url, ttl):
             return (time.now().unix, {})
 
         # Trim off the UTF-8 byte-order mark
-        body = res.body().lstrip("\ufeff")
+        body = res.body().lstrip("\\ufeff")
```
matslina, line 272:
```suggestion
        body = res.body().lstrip("\ufeff")
```

apps/njtransit/nj_transit.star
```diff
@@ -276,7 +276,7 @@ Function gets the destation froma  given depature
 
 def get_destination_name(data):
     nodes = data.find(".media-body").first().find(".mb-0")
-    destination_name = nodes.eq(0).text().strip().replace("\u2708", "EWR").upper()
+    destination_name = nodes.eq(0).text().strip().replace("\\u2708", "EWR").upper()
```
matslina, line 279:
```suggestion
    destination_name = nodes.eq(0).text().strip().replace("\u2708", "EWR").upper()
```

matslina:
@rs7q5 This seems to include some unintentional changes to nj_transit. Happy to merge when those are removed and tests pass!

rs7q5:
Those appeared when I did make lint... 

It seems to happen frequently. With these apps.

PR #917, opened 2022-12-26
New App: valorantstats

ohdxnte:
The first commit of my new Valorant Stats Tidbyt app.

Takes in Riot ID + tag and parses to pull live rank to display on the Tidbyt. If no Riot ID is provided, it will show rank #1 on the VALORANT leaderboard.

This is my first Tidbyt app, criticism is welcome!

Screenshot of it in use:
valorant_stats


ohdxnte:

apps/valorantstats/valorant_stats.star
```diff
@@ -0,0 +1,186 @@
...
+        cache.set(val_tag_api_link, json.encode(rep.json()), ttl_seconds = ttl_seconds)
+        data = rep.json()
+        print("Hit API and cached data")
+    return data
+
+# get image from predefined json image url from stats api
+def getImage(url):
+    if url:
+        print("Getting image from: " + url)
+        response = http.get(url)
```
matslina, line 104: Would it make sense to add some caching around this call? Otherwise we'll be fetching the image every single time a Tidbyt is trying to display this app.

matslina (commented):
Thanks for the PR.

apps/valorantstats/valorant_stats.star
ohdxnte: Good call, that would definitely make sense. I'll implement it this weekend.

betterengineering: I've added some caching here

betterengineering (approved):
Thanks @ohdxnte for the app! I've addressed some PR comments to move this along

PR #918, opened 2022-12-28
New App: Paraland

Yonodactyl:
Hey All!

## Description

Just got a Tidbyt for Christmas and have loved all of the apps I have seen. I love pixel art and the idea of endless scrollers (games or animated art). I decided to create a couple of hand-drawn animations for the Tidbyt. I hope to add more throughout time (they are a little time consuming).

The app gives you a pick of **4** options to choose from and **3** possible speeds. These are adjusted through the configuration.

I hope to improve on the capabilities of this App in future iterations, but have loved looking at what I have made already on my Tidbyt. I hope you all find this useful and fun enough to add to the App List!

## Screenshot
This is the default Paraland screen that scrolls along for new users!

paraland


Yonodactyl:

matslina:
Thanks for the PR. Looks great!

Would it make sense to embed the graphics in the .star instead of fetching them over HTTP? Would reduce outgoing traffic by a bit, and also make it a bit easier for other devs wanting to modify/extend/tinker with the app.

Yonodactyl:
@matslina In trying to get these gifs as `base64` versions, I keep running into this error:
```
Error in Image: decoding image data: gif: unknown extension 0xef
```

I was able to get around this by just setting the base64 strings as just strings (no decoding done), and then decoding when I am rendering them. It looks something like this:

```starlark
# Arizona Desert - Day GIF
ARIZONA_DAY = "R0lGODdhQAAgAH..."  # <<<-- String representation of the base64
# 
# 
selected_img = base64.decode(config.get("image", DEFAULT_MORNING))  # <<<-- Grab the config and decode in main

```

Does this seem like a viable solution?

matslina:
Yeah that's the way to do it. Or even `ARIZONA_DAY = base64.decode("R0lG...")`. Base64 is a way to encode binary data to text, so the base64 encoded version will always be just a string. The decoded version is binary data on the other hand.

matslina:
> Just got a Tidbyt for Christmas and have loved all of the apps I have seen.

Btw: so happy you enjoy it! =)

Yonodactyl:
When I try it that way, I get the failure. I'm worried that the decode is not happening fast enough or accurately when you decode the string as a constant.

I do a comparison and apparently the base64 decoded version of the image from the schema != the base64 decoded version of the gifs constant value

matslina:
I would suggest using some short string value in the schema fields, and map them to the image data at runtime. E.g.
```
DEFAULT_MORNING = base64.decode("BASE64GOESHERE")
...
        schema.Option(
            display = "Default - Morning",
            value = "default_morning",
        ),
...
    img = config.get("image", "default_morning")
    if img = "default_morning":
        selected_img = DEFAULT_MORNING
    ...
```

The last part would be more elegant as a dict, e.g. `IMAGES = {"default_morning": DEFAULT_MORNING, ...` and `selected_img = IMAGES[config.get("image", "default_image")]`

matslina:
Sent you a DM on discord as well. Happy to chat there if helpful.

Yonodactyl:
@matslina ready for a re-review whenever you find some free time

PR #905, opened 2022-12-23
MN Metro Transit App

AmillionAirs:
Well...... we know how much you loved my original Minnesota Light rail app so buckle up boys cause now we got FULL ON BUS FUNCTIONALITY TOO!!!!

My buddy uses just the light rail in the summer, but in the winter time the 20 minute walk to the light rail stop is just to much for his little face to handle. So he then defaults to just getting on the bus that stops a block away from his house and riding to the stop instead. 

I'm submitting this app as a separate submission from the original light rail app I made as I only exclusively ride the light rail. Therefore, the original light rail app i made suits that need perfectly and is compact with only the info i need/want for it.

This new app I'm calling this the "MN Metro Transit" app as it encompasses everything within the Minnesota transit department offers (Bus and Train). This new app adjusts the look of the original app to accommodate for some of the longer naming conventions of the bus lines and also has some additional colors and directional features to go along with it.

https://user-images.githubusercontent.com/97411638/209259656-ecbc009e-8625-4c76-9356-9eed4695b407.mp4

PR #919, opened 2022-12-29
New app: MN Metro Transit

GE-Ninety:
tested updates to app #905, and this PR now replaces that one and is hopefully ready for release.

I'm submitting this app as a separate submission from the original light rail app made by @AmillionAirs as he exclusively rides the light rail, and requested it be separate.

This new app is the "MN Metro Transit" app as it encompasses everything within the Minneapolis - St. Paul Metropolitan area transit department offerings (Bus and Train). This new app adjusts the look of the original app to accommodate for some of the longer naming conventions of the bus lines and also has some additional colors and directional features to go along with it, including color matched name plates for BRT routes. Additionally, the Arterial BRT (ABRT) and standard bus routes have placards to mimic the signage on the respective busses. Light Rail Transit has been redesigned to have white lettering (as opposed to black lettering on BRT) to match the alteration in naming and direction within the app.

Also added to the app is an error screen for "Invalid Stop ID" so users can be certain if they have the right stop code entered. Finally, there are also indications for when there are no more transit departures at a stop for the night (which previously would cause the app to stop functioning due to calling an empty list).  The app should also be capable of displaying future ABRT lines to be added in 2023 and 2024 (letters B, and E-H). 

There are a lot of improvements coming to the Minnesota Metro area over the next 10 years to bring fast, frequent, and reliable transit to the region, and what better way to stay up to date on what's coming to a stop near you than this app bringing you arrival times of all forms, down to the minute!

A big thanks to @AmillionAirs for laying the foundation with the original app, and for all of the guidance along the way (as this was my first Tidbyt app/modification) with linting and formatting. Hope it is up to snuff for all to enjoy!

mn_metro_transit


GE-Ninety:
these are changes to #905 adding pages for no upcoming departures and bad stop ID, which would previously cause an error in the code due to empty lists being called.

AmillionAirs:
@GE-Ninety here is learning git so this is meant to be an improvement on the PR we already have submitted here:

https://github.com/tidbyt/community/pull/905 

I'm going to go ahead and converted it to a draft for now. Let me know if theres anything we need to do for approval on this PR.

apps/mnmetrotransit/mn_metro_transit.star
```diff
@@ -0,0 +1,214 @@
...
+load("schema.star", "schema")
+
+#Assign Default Stop Code
+DEFAULT_STOP_CODE = "15264"
+
+def main(config):
+    #Establish API URL
+    stop_code = config.get("stop_code", DEFAULT_STOP_CODE)
+    url = "https://svc.metrotransit.org/NexTripv2/" + stop_code + "?format=json"
+    MTT = http.get(url).json()
```
matslina, line 20: Going to need some caching for this call to keep request volume down.

matslina (changes requested):

apps/mnmetrotransit/mn_metro_transit.star
```diff
@@ -0,0 +1,247 @@
...
+
+    #Cache Data
+    MTT_cached = cache.get("MTT_rate")
```
matslina, line 24: I think this `cache.get()` is meant to happen before the `http.get()`? Otherwise the http call will happen on each run, regardless of what's in cache.

matslina (changes requested):

apps/mnmetrotransit/mn_metro_transit.star
betterengineering: I've resolved the caching bug

betterengineering (approved):
Hey @GE-Ninety! Thanks so much for the app! I've resolved some PR comments to help move this along. You may want to do some more testing with different stop codes and make sure everything works as expected

PR #921, opened 2022-12-29
New App: Fido

Yonodactyl:
## Description

I really love the idea of having a little companion with you - I have a dog of my own and he's my best friend. I decided that some of us might not have the physical pal, but would love a pixel pal.

Fido is the pixel Pal for your Tidbyt. Give them a name, and give them a Birthday and watch them grow with you!

I want to add more environments as well as include new and exciting options for a pal (cats, and exotics some day). I hope that you all find it as fun as I do to have a digital friend with you throughout the day!

My goal was to also include a "activity meter" but its not entirely possible without a backend to support it or variables that can be stored on device. One day we will have the abilities and then we can update Fido accordingly!

## Screenshot

Screenshot 2022-12-29 at 6 21 21 PM


apps/fido/fido.star
```diff
@@ -0,0 +1,265 @@
...
+""")
+FENCE = base64.decode("""
+iVBORw0KGgoAAAANSUhEUgAAAEAAAAAgCAYAAACinX6EAAAAAXNSR0IArs4c6QAAAUdJREFUaIHtlTFLAzEYht/EA5cDl8IpXUQcOpaCo3Tp4Og/8S/4F/wnjg5ZCt0OSsYudhQzCha8QeqUMx6XS9LkWqHfs1yekLt78+U7DiAIgiAIgiAIgiCODwYALw/3Wz1x9/TMfFxIhdm4+Lfuux8+HU1qAQBfn42LaBdSBXnI8333w/VASGXO78V1aCFVp/eZ5+RycPFYfX7VL1u8fsDmQiq8bb6T+zDPOj32+bfXZ7g6z1udA7/tprG5LorNXffb3PfEmh3hymNbb/qfDoit+Pp9s5MP86wOZvO+OoR3VSj0hGLcRcr3mR3I2ya1m4S6i1JVrWNfYvKZh93bX6CLUlWYr5ZMb9wc+5Iqb4YDMF8tmXk9JNy9JB365Jvz09Fke1Oc7jNKjbMDUgdrbnaX7z8EV34WWv1SVVFFCb2/7/UEQRBHzQ8p2EzduEqDNAAAAABJRU5ErkJggg==
+""")
+
+# Animated Actions
+BALL_THROW = "https://raw.githubusercontent.com/Yonodactyl/TidbytGIFs/main/Fido/ball_throw.gif"
```
matslina, line 41: Are these mostly static or meant to change over time? If static, perhaps it'd be better to just embed their base64 encoding in this script directly? Could make it easier for other devs to suggest changes, and eliminates some HTTP calls as well.

matslina (approved):

apps/fido/fido.star
Yonodactyl: For now - they are static. My goal is to allow these to be modifiable later on. I went with the link to have less code mostly.

I think removing the HTTP calls would be great though. To get over the hump of increasing LOC count, I can probably add the base64 at the end!

matslina: Tips to keep the size of the images down:
- Scale down the file to whatever dimensions you're using in the app (e.g. `convert -scale 32x32 large.png small.png` if you've imagemagick installed)
- Strip exif data from the file (e.g. `exiftool -all= small.png`)

PR #922, opened 2022-12-30
update dev container configuration for GitHub Codespaces and VS Code

joshspicer:
Updates the [dev container configuration](https://containers.dev) in this repository to use the latest features added by the dev containers team.

Dev container configuration is automatically picked up by [supporting tools](https://containers.dev/supporting) of the specification, such as GitHub Codespaces, and the VS Code Dev Containers extension.  Anyone using these clients will immediately have a containerized dev environment set up for quickly getting started.

This change drops an unmaintained `go` base image in favor of the `go` image actively supported by the team at Microsoft/GitHub (https://github.com/devcontainers/images/tree/main/src/go).  

This dev container also installs the latest version of `node` with the `node` dev container Feature (https://github.com/devcontainers/features/tree/main/src/node)

The `postCreateCommand` will trigger the `.devcontainer/install_pixlet.sh` script, which will auto-install (via git tags) the latest release of the `pixlet` CLI, with an optional variable `PIN_VERSION_TAG` that can be set to pin to a specific release.

Due to https://github.com/tidbyt/pixlet/issues/521, `pixlet` was pinned to release `v0.22.7`


betterengineering (approved):
Thanks for this @joshspicer! 🙌

PR #923, opened 2022-12-30
Add `Airthings` Applets

joshspicer:
Resurrected from: https://github.com/tidbyt/community/pull/739

Adds environmental sensor readings from an [AirThings](https://www.airthings.com/) device.

Utilizes the public API (https://developer.airthings.com/consumer-api-docs/), which only supports the client credential oauth flow. Thus, each user must supply their own clientSecret and clientId, which can be generated for free from https://dashboard.airthings.com/integrations/api-integration.

### DIFF
--- a/apps/airthings/airthings.go
+++ b/apps/airthings/airthings.go
@@ -1,0 +1,25 @@
+// Package airthings provides details for the AirThings applet.
+package airthings
+
+import (
+	_ "embed"
+
+	"tidbyt.dev/community/apps/manifest"
+)
+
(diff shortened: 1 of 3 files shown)


joshspicer:
Looks like another applet's tests are failing the CI on this PR.

jvivona:
if you dig deep enough into the test results - you'll see - your app description doesn't have punctuation at the end.
image
 that's the failure.

joshspicer:
oh I missed that, thanks @jvivona!

PR #925, opened 2022-12-30
New App: Hass Entity

InTheDaylight14:
New app that uses the [Home Assistant REST API](https://developers.home-assistant.io/docs/api/rest/) to grab entity state or an attribute value and display it. 

Cache does not make sense for this app as this is pinging user's self-hosted instances of Home Assistant.


InTheDaylight14:

matslina:
Hey @InTheDaylight14, thanks for the PR.

Community apps all execute in Tidbyt's backend environment, so they're gonna have a very hard time reaching a lot (the vast majority?) of HA instances.

InTheDaylight14:
The Rest API still works via an exposed HA instance (custom domain) or their subscription "Nabu Casa" domain. I tried to clarify in the schema that this app needs an exposed HA url to work. I was able to test both ways on my end and it worked. 

Most of us have HA exposed in some way so that mobile phones can send data to HA when away from home. Some people only use a vpn to send data back home. This app would not work for those people. 

Ideally the display would be able to reach out on the local network for information and it would work for all HA instances. But this will still work for the majority of HA users.

matslina:
Ok, great. I fixed a conflict for you, but there still seems to be some lint issues that'll need fixing.

InTheDaylight14:
When I do ```make lint``` it changes 242 files. Any advice on tracking down the lint issues for a first timer?

Looking into the logs of the checks it looks like other files are failing? I have no idea

InTheDaylight14:
@matslina I figured it out! Looks like a WoW Token app test is failing. I assume I don't have to worry about that for this PR

matslina:
I think this is the error you need to get past: `icons_test.go:36: app 'Hass Entity' contains unknown icon: 'inputText'`
Check out https://github.com/tidbyt/community/blob/main/apps/icons.go for a list of supported icons.

matslina:
So there's one pretty big remaining issue, which is that we currently can't merge apps that call user-provided URLs. This is so to prevent malicious actors from abusing our platform to spam or do other nasty stuff.

I really should've spotted that sooner instead of wasting your time on other fixes... Sorry about that.

We're hoping to figure out some reasonable rate limited solution in the future to work around this, but we're not quite there yet.

InTheDaylight14:
Hmm, that's fair. I can think of one potential work around. We could force users to use the nabu.casa url. The URL format is similar to an api key, long and random. As an example, mine is similar to following url:
```
https://4xxxxxxxxxxxxxxxxxxxxxxxxxxxx.ui.nabu.casa
```

Users could input the ```4xxx...``` part of the url and the app can take care of the rest of the URL. This will prevent users from inputting a fully qualified domain. I'm not sure if this counts as a work around to the rule. 

In the future we could open up the option to use a personal domain once the rate limiting is ready.

matslina:
I think that'd be ok from our perspective.

apps/hassentity/hass_entity.star
```diff
@@ -79,15 +79,15 @@ def get_schema():
         version = "1",
         fields = [
             schema.Text(
-                id = "ha_ip",
-                name = "Home Assistant External Url",
-                desc = "External HA Url (and optional port). Ex. https://abc.ui.nabu.casa or https://hass.mydomain:8123",
+                id = "nabu_casa_url_key",
+                name = "Nabu Casa Url Key",
+                desc = "The random letters and numbers in your Nabu Casa URL. Ex. Input 'abc123' for this nabu casa url https://abc123.ui.nabu.casa",
```
matslina, line 84: Just a heads up that the `desc` fields is currently not visible to users in the Tidbyt mobile app.

PR #927, opened 2022-12-30
New App: Idle Gardener

Yonodactyl:
## Description
I often times download and delete Idle games on my phone. I like the idea of a game where I dont have to do anything, but it will play itself. I think the idea of having a tree chopping (or planting) simulator is a really fun concept for an Idle Game.

Thus, enter Idle Gardener! This is a set and forget type of Idle game. You can watch the counter go up every time the app is re-run and the trees animate in from a `chopped` state, to `grown`. The counter will grow as you clear out the trees, and they grow back just in time for you to cut them down again!

**Note:** I considered that some users might not like the idea of deforestation, therefore I added a `Green Mode` that will plant the trees instead - same animation, just makes you feel like you are making a difference! They share the same cache, so you wont start over by toggling the schema.

I hope you all find this fun and enjoy the Idle Gardening lifestyle!

## Goals
- I hope to add in season recognition with future iterations - it would be nice to see the leaves change colors with the seasons. This was one of my favorite things about the game Stardew Valley - hope I do it justice!
- Add more stats (Oxygen reduction counter? Wood Production counter?)

## Screenshot
Screenshot 2022-12-30 at 5 10 51 PM

### DIFF
--- a/apps/apps.go
+++ b/apps/apps.go
@@ -80,6 +80,7 @@ import (
 	"tidbyt.dev/community/apps/howoldami"
 	"tidbyt.dev/community/apps/hurricanetracker"
 	"tidbyt.dev/community/apps/hvvdepartures"
+	"tidbyt.dev/community/apps/idlegardener"
 	"tidbyt.dev/community/apps/ifparank"
 	"tidbyt.dev/community/apps/indegostations"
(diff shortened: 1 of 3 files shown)


matslina:
Fun!

PR #928, opened 2022-12-30
Add 24 December option

dinosaursrarr:
My Swedish friends open their presents on Julafton on the 24th, so that's when they'd want to count down until.

### DIFF
--- a/apps/isitchristmas/is_it_christmas.star
+++ b/apps/isitchristmas/is_it_christmas.star
@@ -169,6 +169,10 @@ def get_schema():
     # https://en.wikipedia.org/wiki/Christmas#Date_according_to_Julian_calendar
     date_options = [
         schema.Option(
+            display = "24 December",
+            value = "12-24",
+        ),
+        schema.Option(
             display = "25 December",
             value = "12-25",
         ),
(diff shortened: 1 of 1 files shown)


rs7q5:
At this point does it make sense to add a schema.generated option to put other? and then let one manually choose a date? Like if they celebrate with one side of the family a different time of year.

Could either have them enter how the value is in the option fields, or have two separate integer entries for that?

Would have to change the `christmas_date` line to be something like
```
if config.get("christmas_date","12-25")=="None": 
 christmas_date = config.get("christmas_date","12-25")
else:
 christmas_date = config.str("custom_date",default)
```

the none is a string because the value is meant to be a string in the options. assumes the custom_date is text. 

This MAY defeat the purpose of the app though... and just a thought. cc @austinrfnd

dinosaursrarr:
I think it makes sense to have the predefined list because this is a specifically Christmas countdown app.

There are still only a handful of days that are Christmasses around the world. And still few enough to list them.

Setting a custom date would make it more of a generic countdown app, but a generic countdown app shouldn't have all the Christmas imagery.

rs7q5:
Yeah, I think that totally makes sense.

PR #931, opened 2023-01-02
Add Chicago CTA "L" Train Tracker App

samshapiro13:
### Description
Just got a Tidbyt for the holidays and was very impressed by the range of transit apps provided and that the community has already built. I saw that there was not a Chicago specific app, but that there was a generic Transit app that was able to load Chicago train times. However, when comparing with my CTA Ventra app on my iPhone, I noticed that the train times on the Transit app didn't match the live data shown in the Ventra app. I was then inspired to create a Tidbyt app that uses live CTA Train Tracker data to accurately show live train arrival estimates.

This app presents a list of all CTA stops to choose from, retrieved from the City of Chicago "L" Stops API. When a station is chosen, the app will show the two soonest train arrivals to that station returned by the CTA's Train Tracker API. Things to consider:

- Station selections will show all of the train lines that stop at that station in parentheses. The two soonest arrivals will be shown regardless of how many lines stop at that station.
- The default station selected is the first station alphabetically returned from the "L" Stops API. That station is currently 18th (Pink Line).

There are two types of arrivals returned from the CTA's Train Tracker API: live and scheduled. By default, arrivals shown for the selected station should be live. However, if there are no live trains, the CTA's API will return scheduled arrivals. These are not necessarily accurate reflections of when trains will arrive, but are rather the schedule by which they are supposed to. Since live arrivals are generally accurate and scheduled arrivals are not, any scheduled arrival is marked with an "S" to the right of the estimated arrival time.

### Screenshot
This is a screenshot of the CTA "L" Train Tracker app with the Paulina (Brown Line) station selected:

cta_l_tracker


samshapiro13:

apps/ctaltracker/cta_l_tracker.star
```diff
@@ -0,0 +1,289 @@
...
+Gets top 2 arrivals scheduled for the selected station
+from CTA Arrivals API
+"""
+
+def get_journeys(station_code):
+    api_key = secret.decrypt(ENCRYPTED_ARRIVALS_API_KEY)
+    if api_key == None or station_code == None:
+        return None
+
+    response = http.get(
```
matslina, line 247: Could you add some caching for this call?

matslina (changes requested):
Thanks for the PR! Great to have a dedicated and hopefully more well maintained app for the L!

One request for additional caching, but otherwise I'm happy to merge.

apps/ctaltracker/cta_l_tracker.star
samshapiro13: Sure, I'll add a 60 second cache to this

matslina (approved):

PR #933, opened 2023-01-03
New App: Top Twitter Trends

jre9158:
# Me
Hi everybody! My name is Joey Esposito. I'm a 4th year RIT student in Computer Science and started seeing what I can do with the Tidbyt.
# About
Displays the top Trending Hashtags on Twitter. The limit is set in a schema that maxes at 50, scrolling vertically through them.

The colors of the trend's text determine how many tweets it has. 
>White : No Volume Data
>Green : Less than 25K
>Blue : 25K - 100K
>Orange : 100K - 250K
>Purple : 250K - 500K
>Red : 500K - 1M
>Gold: Over 1M

This app does not post any tweets or change any data on an account. It also does not query any users or their tweets. Simply grabs the JSON file twitter provides for their top 50 trends from their API.

# GIF
twitter_trends

# API
Data is retrieved from Twitter's API v1.1. To use this, an elevated developers account is needed, which was free and takes 10 minutes. (If there is a better way of doing this please share). 

After creating a developer's account, then requesting elevation, add a new application to get the keys. (I tried to use the keys from my application that was created before I had elevated access, and it was causing issues. Adding a new application fixed it)
They provide you with a key and a secret key which is used to authenticate and get a Bearer Token to use for the GET request.
This app uses Twitter API v1.1 because that has a GET request for the top trends. I did not see a similar request in their v2 API.

Based on the limits for requests I see in the Twitter API. This app may be able to run off of one API account for everyone which could be something to look into.

# Issues
lint is failing in my get_schema function.
Right now it just asks for the key and secret key as text fields.
I couldn't use the schema.star in testing, so I am unable to verify that you can authenticate it using the schema.OAuth2.
In my implementation, I have all the schema code commented out in order to test it.
My hope is that this is possible but without testing it, my implementation could vary slightly and needs to be adjusted to work with the schema. I am not very well versed in APIs or authenticating them. I got the Tidbyt to start diving in. 

Any tips are appreciated!
I tried to follow all the steps for a healthy pull request.
# Screen shot
Twitter-Trends-Static-Image


jre9158:

apps/twittertrends/twitter_trends.star
```diff
@@ -0,0 +1,263 @@
...
+        #Commenting out Schema for tests
+
+        #Check for API keys
+
+        if config.get("key", None) != None or config.get("secret", None) != None:
+            api_parameters = {
+                "client_id": "{}:{}".format(config.get("key", None), config.get("secret", None)),
+            }
+        else:
+            return render.Root(
```
betterengineering, line 60: Any chance you can return a meaningful preview? When we display apps in the apps store, we render the app without any config values, which means all you would see in the app store is "No API Keys!" D:

betterengineering (changes requested):

apps/twittertrends/twitter_trends.star
jre9158: Hi! Should be all set to return a preview of the app now. Let me know if you need anything else!

kleban31:
Still working on caching

jre9158:
> Still working on caching

What do you mean?

kleban31:
> 

@jre9158 I sincerely apologize. I commented on the wrong PR. Please disregard.

PR #934, opened 2023-01-03
lordoftherings: new api key

matslina:


matslina:
@jmanske The original API key secret didn't decrypt correctly, so I'm swapping it out for a new one I just registered.

jmanske:
Thanks! I feared I had done something incorrectly there and there is no good way to test it.

PR #937, opened 2023-01-04
Nightscout: Fix vertical bars where hour crosses day

IsThisPaul:
Fix vertical bars where hour crosses day.
Vertical hour bar was not passing test to render when hour was midnight GMT


IsThisPaul:
@gabe565 for visibility

gabe565:
🤦 I can't believe I missed this. Thanks for the fix!

PR #938, opened 2023-01-04
New App: Any Progressbar for Tidbyt

wojciechka:
Displays one or more progress bars.

The idea is to install one or more instances and use Tidbyt API to update the configuration to show statuses.

This is a new take on #590 that depends on config instead of arbitrary URLs.


wojciechka:
Sample rendering of the progress bar:

any_progressbar

PR #942, opened 2023-01-06
New App - BBL Cricket

M0ntyP:
This app is for following your favourite BBL team (BBL = Big Bash League, Australian T20 cricket tournament).

Just select your team from the dropdown and it will show either the next upcoming game, the current game or previous game. 

For the upcoming game, you'll see a screen like this. Showing the team records and time of the next game, in your local timezone.
bbl_cricket

For a live game, you'll see something similar to this for the team batting first
bbl_cricket_live
And this for the chasing team
bbl_cricket_2nd
There is a slight difference in the information shown on the bottom row

And for a completed match, you'll see 
image

I had a lot of fun making this app especially dealing with all the nuances in the game of cricket. Please let me know if you find any issues as I'm sure there is a scenario I haven't allowed for 

Thanks!


apps/bblcricket/bbl_cricket.star
```diff
@@ -0,0 +1,771 @@
...
+    "locality": "Adelaide",
+    "timezone": "Australia/Adelaide"
+}
+"""
+
+def main(config):
+    SelectedTeam = config.get("TeamList", DEFAULT_TEAM)
+    SelectedTeam = int(SelectedTeam)
+
+    Standings_JSON = http.get(Standings_URL).json()
```
matslina, line 32: We'll need some caching around these calls to make sure we're not hitting the API to frequently.

apps/bblcricket/bbl_cricket.star
```diff
@@ -0,0 +1,771 @@
...
+
+            location = DEFAULT_LOCATION
+            loc = json.decode(location)
+            timezone = loc["timezone"]
+
+            MyTime = time.parse_time(StartTime, format = "2006-01-02T15:04:00.000Z").in_location(timezone)
+            Time = MyTime.format("15:04")
+            Date = MyTime.format("Jan 2")
+
+            Standings_JSON = http.get(Standings_URL).json()
```
matslina, line 544: Caching here too.

apps/bblcricket/bbl_cricket.star
```diff
@@ -0,0 +1,771 @@
...
+    schema.Option(
+        display = "Hobart Hurricanes",
+        value = "4845",
+    ),
+    schema.Option(
+        display = "Perth Scorchers",
+        value = "4846",
+    ),
+    schema.Option(
+        display = "Melboure Renegades",
```
matslina, line 703:
```suggestion
        display = "Melbourne Renegades",
```

matslina (changes requested):
Thanks for the PR. I bet a lot of cricket fans will really enjoy this one.

A few changes needed before we can merge and publish. Primarily need some caching to keep outgoing request rates as low as possible.

apps/bblcricket/bbl_cricket.star
M0ntyP: Can do, I'll fix that up

apps/bblcricket/bbl_cricket.star
M0ntyP: I can't believe I spelt that wrong! I must have looked at the dropdown 100s of times during testing 

Great pick up!

M0ntyP:
The spelling errors have been corrected. I have also implemented caching and will need to test it against a live game tonight to ensure that it works correctly

M0ntyP:
I have uploaded a new version of the .star file now. Hopefully its good to go!

matslina (approved):

apps/bblcricket/bbl_cricket.star
matslina: This one should be a `get_cachable_data()`, right?

matslina: Or just removed entirely since you're geting `Standings_JSON` above on L536?

M0ntyP: Yes, we can remove it as its already defined earlier on

M0ntyP: I've commented it out now

PR #945, opened 2023-01-10
Fix automatically resolvable lint issues.

betterengineering:
This commit fixes automatically resolvable linter issues by running:
```
pixlet lint -r --fix ./
```


betterengineering:
I think I'm going to have to disable the map order as well. It's a lot of changes and it feels arbitrary. I could see an argument that it's not safe to rely on map ordering, but it's also handy to set up a logical order in code

PR #949, opened 2023-01-11
Resolve lint issues.

betterengineering:
This commit resolves additional lint issues by manually editing the source code. I'm being extra paranoid when making changes, but there is a chance I've made a mistake. This is part of a larger effort to pay down our linter issues so I can enable `pixlet lint` for pull requests.


apps/amazing/amazing.star
```diff
@@ -338,14 +340,16 @@ def next_start(visited):
                 if neighbour in visited:
                     return cell
 
+    return None
```
matslina, line 343: :\

betterengineering: This resolves a linter issue where there is an implicit return - this felt the safest way to resolve it?

PR #957, opened 2023-01-12
nationalrail: config fix

matslina:


matslina:
@dinosaursrarr I think this will do the trick.

PR #962, opened 2023-01-12
Fix NES Quotes unused variable.

betterengineering:
The original method was not correct, and the linter complained about result being unused. The issue is, this method is used to create an ID for schema fields. If we change the ID, it will break any existing installations. Therefore, the safest option is to mimic the behavior of the original function. cc @markmcintyre

### DIFF
--- a/apps/nesquotes/nes_quotes.star
+++ b/apps/nesquotes/nes_quotes.star
@@ -23,16 +23,23 @@ CACHE_TTL = 604800
 
 # CREATE SLUG
 # -----------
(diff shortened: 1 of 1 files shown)


markmcintyre:
Dang, that's embarrassing. 😂 

I wonder if it's worth fixing this properly anyway, since the JSON file that's fetched to control the game toggles could add new games over time and therefore have conflicts with the names? Yeah, some users who toggled a game off might have it reactivated, but it'll be more future-proof going forward.

PR #964, opened 2023-01-12
martamap: fix config bools

matslina:


matslina:
@InTheDaylight14 Noticed some config fields weren't handled correctly when preparing to publish. Calling `config.get()` will give you a string value, so these won't be type bool by default.

InTheDaylight14:
Oh that's my bad. Thanks for the fix!

PR #966, opened 2023-01-12
Fix ethstaker lint warning.

betterengineering:
This commit resolves our last lint waring for status being undefined. I am not sure the intent here, and looking through the rest of the source code was no help. I presume this method already was erroring given this was not defined. My fix satisfies the linter but it likely does not resolve the logic error in this mehthod. cc @ColinCampbell

### DIFF
--- a/apps/ethstaker/ethstaker.star
+++ b/apps/ethstaker/ethstaker.star
@@ -75,7 +75,7 @@ def main(config):
         )
 
 def header_status(statuses):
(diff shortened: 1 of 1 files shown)


ColinCampbell:
Hi @betterengineering! When I go to the commit prior to this pull request, I can't get a lint warning/error to trigger on this line. Any tips on that? Tried running `make lint`. Also wondering what logic error you saw? AFAIK the app is working currently, I have my Tidbyt with it up right next to me :)

PR #967, opened 2023-01-13
Update CTA "L" Tracker - refactor and remove arrivals cache

samshapiro13:
Hey team,

Upon testing the CTA "L" Tracker app from within the Tidbyt app, I realized that having a 1 minute cache on the arrivals API response was very clunky when selecting a station from the dropdown. Because the arrivals API is called by default for the first station in the dropdown, the user must wait a minute after installing the app to select a station and see arrivals for their selected station. I think the user should be able to select any station from the dropdown and immediately see arrivals for that station. I noticed that some other transit apps do not maintain a cache for their arrivals (nycbus, mbta, etc.), so I think it is reasonable to remove the cache on arrivals for this app.

In addition, I refactored the code to be cleaner and avoid a duplicate call of the same function from both main and get_schema. Both functions were calling get_stations which makes an API call and caches the result. I instead pass each station's map id through the dropdown as the hidden value and am able to use that to skip the mapping of station name to map id.


matslina:
Hey there, thanks for the PR.

I think a better approach might be to include `station_code` in the cache key used for arrivals. The cache is shared among all users of the app (i.e. it's scoped to your app, but not to individual users), so regardless of selected station users will mostly just see what happens to be stored under the `ARRIVALS_CACHE_KEY`.

Removing arrivals caching will work, but if the app becomes popular there's a good chance a rate limit will kick in on the API providers side, breaking the app.

samshapiro13:
This is a great idea - I'm implementing this solution with a 60 second arrivals cache. Thanks for the review!

samshapiro13:
This is ready for review again, thanks!

PR #972, opened 2023-01-14
New App: Mastodon Follows

drudge:
This PR adds a Mastodon Follows app similar to my Twitter Follows app. I took a slightly different approach to this than #877, with these three goals in mind:

1.  I wanted to add the follower count of any username on an instance, not just my own
2.  I did not want to have to login to get the follower count
3.  I did not want to hardcode a list of instances into the app, so this searches [instances.social](https://instances.social) instead.

You can get your own free token for the search call [here](https://instances.social/api/token) if you'd prefer.

### Preview

mastodon_follows


matslina:
Thanks for the PR!

PR #975, opened 2023-01-16
New App: SolarEdge solar energy monitoring

marcusb:


matslina:
Hey @marcusb, and thanks for the PR! Hope you're doing well and so on. Been a while!

Would you mind adding a default graphic for the case where site id and api key aren't provided ("" or None)? Something illustrative of what it'd look like if they actually provided their own configuration. Makes it easier for users to decide if they're interested in setting up the app.

PR #976, opened 2023-01-17
[feature-update] Add travel time configuration to goodservice app

connernilsen:
Hello! 

I love using this and the default NYC subway time apps, but one thing I feel is missing is the ability to only show trains that will be arriving after a certain delay. I personally live around 10 minutes away from the subway, and I think being able to account for travel time to the subway would be an amazing addition!

I'm also looking to make the same edit to the default NYC subway time app if possible, but I'm not sure where to find it. Would I be able to get a pointer to where it lives in code?

Also, I'm very open to changing the way travel times are configured if there are any suggestions there.

Please let me know if there are any edits that I can make!


connernilsen:

PR #977, opened 2023-01-17
New App: Stripe Sales

jonbilous:
This app does one simple thing: displays a user's Stripe sales made so far today. 

stripe_sales

It only requires a user to provide a Stripe API Key, and all of the requests to the Stripe API are made directly, not through any external proxy.


jonbilous:

jonbilous:
recheck

matslina:
Hey @jonbilous, thanks for the PR!

Would love it if you could add a default graphic for the case when no api key is provided. That way users can get a sense of what the app looks like before they make a decision to go get their own API keys and install.

PR #979, opened 2023-01-18
New App: Last.fm Now Playing

MattyGroch:
As someone who doesn't use Spotify, I wanted an app that was capable of displaying what I was listening to based on my scrobbler for Last.fm. The API calls to Last.fm use my dev key, but users may supply their own if they want.


MattyGroch:

MattyGroch:
It clearly doesn't like having a period in the package name but I'm not sure what all would need to change to fix this. Assistance would be appreciated!

PR #983, opened 2023-01-19
Add max_age

dinosaursrarr:
Transport apps are also time sensitive, and solar elevation is really a clock.


betterengineering (approved):
Thanks @dinosaursrarr! This all seems reasonable. I'd hate for someone to miss a train 😭 .

PR #984, opened 2023-01-19
Added Summary Feature to Visible Planets

rjison:
Added a feature that displays all the planets visible at once. It's very compact, but the idea is that you can sketch what you see on the Tidbyt and go outside and find the planets that are visible. 

The bottom row shows directions, with "North" in the middle of the screen. There are 5 rows of letters, the higher the row, the higher in the sky that planet will appear. Each row representing about 18 degrees. So the letter "J" in the top middle would mean Jupiter is visible in your location if you look North, then 72 to 90 degrees above the horizon. 

If it is during the day, the display will be of the night sky as of sunset. Data is cached about an hour, and throughout the night, the night sky display will change as the earth rotates about its axis, and the planets appear to move across the sky.


apps/visibleplanets/visibleplanets.star
```diff
@@ -66,8 +86,8 @@ def main(config):
     system = config.get("system") or "metric"
     hide_quiet = config.get("hide_quiet")
     location = json.decode(config.get("location", default_location))
-    planet = config.get("planet", "mercury")  #default to mercury
-    is_inner_planet = planet == "mercury" or planet == "venus"
+
+    planet = config.get("planet", "all")  #default to mercury
```
matslina, line 90:
```suggestion
    planet = config.get("planet", "all")  #default to all
```
?

rjison:
Opps... I think it would be best to default to mercury and adjust the
comment.

The "All" is useful for folks who, like me, want a summary of everything I
can see  at night, but a lot of folks may have mercury on their Tidbyt by
default and switching to the "All" option would be confusing.

Let me make the change and check that in... unless you think otherwise?

rs7q5:
Just added this to my tidbyt! And was thinking about seeing all the visible planets. 

Could maybe have it be "All visible planets" be the option or something more explicit than just all? Unless All is "All planets" whether visible or not?

PR #989, opened 2023-01-20
New App: GitHub Unread

ElliottAYoung:
Adds a new app to display the User's current number of Unread Notifications on GitHub. Uses the GitHub Rest API to fetch the notifications, which requires a personal access token from the user as the only necessary argument.

Example Render:
github_notifications copy


ElliottAYoung:

ElliottAYoung:
@matslina Hello, just wondering if there is anything else I can do to help move this along?

betterengineering:
Hey @ElliottAYoung! This is what will show up in our store and is what @matslina is referring to:
image

I'm going to merge this a while, but you should consider following up with some form of demo content so users can see what the app could look like if they go through the setup.

PR #992, opened 2023-01-20
New app: Sky News

meejle:
skynews


betterengineering (approved):
Thanks for the app @meejle! I ran these two commands to fix the formatting and linting issues:
```
pixlet format apps/skynews/skynews.star
pixlet lint --fix apps/skynews/skynews.star
```

PR #998, opened 2023-01-23
[feature-request] Shorter shortnames for goodservice app

connernilsen:
This PR adds shorter shortnames for Goodservice. Instead of displaying the name of the last station on the route, a configuration is provided to allow for setting the short name to "Uptown" or "Downtown". This PR also slows down the scrolling motion when displaying stations to make reading easier for users.


connernilsen:
I'm going to be looking into the failing check, but I want to get the review process started on this so I'm opening it up to reviews.

rs7q5:
Looks like you're erroring out because `inputText` is not a free fontAwesome icon.

connernilsen:
> Looks like you're erroring out because `inputText` is not a free fontAwesome icon.

Thanks for that, @rs7q5! I thought I changed that when I last amended the diff, but it looks like it didn't go through 🙃

matslina:
@blahblahblah- any opinions on this one?

blahblahblah-:
Unfortunately, "Uptown" and "Downtown" really only makes sense within Manhattan and also does not work on the crosstown lines (i.e. 7, L, S), so I would prefer to use borough-based destinations or by cardinal direction (I prefer the former).

connernilsen:
Thanks for the feedback, @blahblahblah-! I've gone through and set the destination to be the capitalized direction instead.

betterengineering:
Thanks for the change @connernilsen! @blahblahblah- raises legitimate concerns. A Southbound 7 train would be nonsensical :sweat_smile:. I'm going to close this, but if you really wanted this type of change we could maybe support another app?

blahblahblah-:
I may revisit this in the future to have borough-based direction names using the geolocation of the station in the future, since that would require adding more information to the API.

blahblahblah-:
Actually @connernilsen, I'm open to merging the PR as it currently stands, if you would make certain lines be be "East" and "West" instead. Specifically 7, L, and GS.

PR #1000, opened 2023-01-23
New App - The Bored App

AndersHeie:
Are you feeling bored?
Good news ! The Bored App will make random suggestions for you and your friends to do.
Screenhots and readme.md included.


AndersHeie:

apps/bored/bored.star
```diff
@@ -0,0 +1,204 @@
...
+
+# Global defines
+BORED_URL = "https://www.boredapi.com/api/activity"
+DEFAULT_COLOR = "#FF00FF"
+DEFAULT_FRIENDS = "random"
+DEFAULT_FONT = "6x13"
+DEFAULT_DIRECTION = "vertical"
+
+def main(config):
+    activity = cache.get("activity")
```
matslina, line 23: Since cache key is the same regardless of the provided params, most users will just see the result for whatever params that happened to be provided when cache was populated.

matslina:
Thanks for the PR! Note the comment I added though.

apps/bored/bored.star
AndersHeie: Ah, thanks. My first app, I'll fix it.

PR #1002, opened 2023-01-23
App Update - The Bored App - Fix of caching key

AndersHeie:
- Fixed caching string to be depending on [relevant] settings.


AndersHeie:
Woopsie

PR #1004, opened 2023-01-23
Update - The Bored App - Fixed caching key

AndersHeie:
- Updated caching key to be unique for API call parameters combos
- Fixed a super important README.md typo


apps/bored/bored.star
```diff
@@ -20,26 +20,36 @@ DEFAULT_FONT = "6x13"
...
-            "participants": str(friends),
+            "participants": friends,
         }
+
         print("params", params)
         rep = http.get(BORED_URL, params = params)
         if rep.status_code != 200:
+            # if the APi fails, return [] to skip this app showing
             fail("Bored request failed with status %d", rep.status_code)
+            return []
```
matslina, line 49: Calling `fail()` will immediately halt execution, so the `return []` won't be reached.

PR #1005, opened 2023-01-24
Fido: Allow the ability to randomize the actions Fido takes

rs7q5:
I think there is a bug in either `pixlet serve` or something because I had to set random=="true" instead of it being a boolean. So am having to check against the string. Also required a toggle of True/False to get the extra config to show up when using `pixlet serve`

cc @Yonodactyl . FYI I refreshed the page on `pixlet serve` and it seemed to give me a different setting when you had the randomizer on, but if you could sanity check me that would be great!


rs7q5:
Looks like the "bug" in `pixlet serve` is intentional where it should be `"true"` or `"false"`, however one doesn't use `config.bool` when using generated. So whether that part is expected behavior I'm unsure. No issue though as I just check the string.

matslina:
Would it be easier to just add a "random" option to the dropdown?

rs7q5:
I thought about that and started that initially, but then I needed to deal with redefining `action_config` or whatever it's called from getting the option and dealing with not wanting that to be included in the randomization. But also so when you choose `random` based on the list of options or something, one would have to eliminate the random option. was trying to minimize the format of the actual code itself if that makes sense. 

I could remove the generated and keep the toggle and have both of them just exposed always? and so the selection of the dropdown doesn't matter. 

I would just need to move the generated code back into `get_schema()` and then I think everything else would be the same. 

Let me know if that makes sense.

rs7q5:
@Yonodactyl feel free to change the order of the schema or in fact have the random option. I am torn between just adding the random option to the dropdown, but am open to it if we think that is in fact easier. The way its coded I though the toggle was easier.

rs7q5:
@matslina realized I could do this with the dropdown pretty easily by offsetting my random number generator indices. Let me get this fixed... Will fix #1005 as well. Unless we like this toggle for clarity. But I understand not having unnecessary configs. Sorry for the confusion

rs7q5:
Okay I don't think anything I did screwed up the cache, but should check with @Yonodactyl . I did leave the defaults as they were and added the random option to the top

Yonodactyl (approved):
Love this. I like the idea of not having to watch Fido run across for days!

betterengineering (approved):

PR #1007, opened 2023-01-24
New App: Adelaide Metro

M0ntyP:
Inspired by all the other great transit apps out there, I made one for my home town

This app displays upcoming services for train stations and bus & tram stops around Adelaide

### DIFF
--- a/apps/ncaamstandings/manifest.yaml
+++ b/apps/ncaamstandings/manifest.yaml
@@ -1,7 +1,7 @@
 ---
 id: ncaam-standings
 name: NCAAM Standings
-summary: NCAAM football standings
+summary: NCAAM basketball standings
 desc: View NCAAM standings by conference.
 author: LunchBox8484
 fileName: ncaam_standings.star


apps/adelaidemetro/adelaide_metro.star
```diff
@@ -0,0 +1,1112 @@
...
+    STOP_ID = str(SelectedStation)
+
+    NEXTSCHED_URL = NEXTSCHED1_URL + STOP_ID
+
+    # Cache the next service times for 1 min
+    NextSchedCacheData = get_cachable_data(NEXTSCHED_URL, 60)
+    NEXTSCHED_JSON = json.decode(NextSchedCacheData)
+
+    INFO_URL = STOPINFO_URL + STOP_ID
+    INFO_JSON = http.get(INFO_URL).json()
```
matslina, line 49: Some caching here would be great!

matslina (changes requested):

apps/adelaidemetro/adelaide_metro.star
M0ntyP: Hi @matslina , thanks for the feedback. I didn't want to cache that call as its the initial call to check if the Stop ID entered by the user is valid. So if they do enter an incorrect ID it wouldn't be cached, they get the "Invalid Stop ID" message and they can try again straight away. If the ID is correct then the data does get cached a few lines later at 62 & 63

PR #1011, opened 2023-01-24
feat: The Big Migration

betterengineering:
# Overview
This change migrates the community repo to a brand new, unified tool chain built into `pixlet`. Instead of needing `Make`, `go`, `buildifier`, and `golangci-lint` installed, Community developers only need `pixlet` installed 🎉 .

# Notable Changes
- New apps can be created by running `pixlet create`
    - It's identical to `make app`, but it's now built in 😍 
- To see if your app is ready to publish, run `pixlet check apps/{{ app }}/{{app}}.star`
    - It's the same as running all of the formatting, linting, and testing tools BUT it's built in 😍 
    - Even better, it's the same command that's running in CI 💪 
- App manifests are now in YAML
    - These are the same as before, they just are written in YAML instead of a go source file 🙏 
- No need to update `apps.go`
    - We're now doing that step programmatically 💯 
    
# Why
Developers should only have to have one tool installed to develop Tidbyt apps: `pixlet`. Right now, we require so many toolchains installed to be able to create a Tidbyt app and it's a barrier of entry. We're looking to remove as many barriers as possible to publishing apps this year, and this is our first step.
    
# Existing Apps
This change is a no-op! We read in your manifest and wrote out a `manifest.yaml`. If your app worked before, it should continue to work the same way. If it doesn't, we have a bug 🙀 .

# Outstanding PRs
My hope was that we could land all outstanding PRs before this one. Reality is, that's not going to happen as it takes a few days on some PRs to go back and forth with the author. So instead, I'm going to update existing PRs to this standard. It's not going to be easy or 😭 . I'm open to feedback here.

# New PRs
New PRs should be opened using `pixlet create`. If you think your app is ready for review, run `pixlet check` on your app and make sure it's green before opening a PR.


betterengineering:
One thing to note, we need a tiny bit of `go` still. We will need to update the deployment mechanism in our backend before we can remove go entirely, but it's high on the list of things to tackle.

PR #1018, opened 2023-01-26
new app - display insight profile color blocks

amhefele:
New app to display Insight profile colors blocks based on your personality profile. Select your colors in priority and the shape of the display in case you lead with two colors. We use the large lego-like blocks at work and will be nice to show our personality colors on our Tidbyt with this app =) 

color_blocks


amhefele:

betterengineering (approved):

PR #1019, opened 2023-01-26
New App: Tar1090 ADS-B

cbattlegear:
App to view ADS-B data from a self hosted, publicly available, instance of tar1090. Tar1090 is pretty common in the ADS-B world and has a consistent API which can be easily reached via reverse proxies. 

for testing purposes feel free to use my instance @ https://flights.battagler.me/

I tried my best to make most items to go a non-self hosted item, but the primary information still has to come from the end users system. 

tar1090_ads_b

Happy to converse about ideas around this as I have a few other self-hosted, widely used tools I would love to build up apps for!


cbattlegear:

cbattlegear:
recheck

matslina (changes requested):
Thanks for the PR. Looks very cool, but I'm afraid we can't publish apps that perform calls to user provided URLs. This is so to protect against abuse where our platform is used to spam/flood arbitrary hosts on the internet. We hope to have a solution in place for this at some point this year, but I can't give you a timeline for it right now.

PR #1021, opened 2023-01-26
New app: Guardian News

meejle:
guardian_news


betterengineering (changes requested):
Hey there! Thanks so much for the app! You'll want to run the following to fix up the build:
```
 pixlet format apps/guardiannews/guardian_news.star
 pixlet lint --fix  apps/guardiannews/guardian_news.star
 ```
 
 You'll also need to add the `edition` to the cache key. Right now, the request is cached under the same key, which means changing the edition doesn't update the app and would cache for all users 🙀

meejle:
> Hey there! Thanks so much for the app! You'll want to run the following to fix up the build:
> 
> ```
>  pixlet format apps/guardiannews/guardian_news.star
>  pixlet lint --fix  apps/guardiannews/guardian_news.star
> ```
> 
> You'll also need to add the `edition` to the cache key. Right now, the request is cached under the same key, which means changing the edition doesn't update the app and would cache for all users 🙀

I think that should be it! Let me know if you need anything else doing to it. :)

betterengineering (approved):
Thanks for the updates @meejle!

PR #1025, opened 2023-01-27
App Update: Color selectors and minor changes for Desk Name Tag

b-bell:
This PR includes two new dropdowns for users to select custom colors. Additionally, I've updated a few app field names to make what users put on the additional lines of the name tag less opinionated.


stephenzeng55:
I would remove the requirement for putting the `#` symbol before the hex color since all of the colors have them.

b-bell:
@stephenzeng55, you're right. I've updated the app to work if you enter a valid hex code with or without the hashtag. It'll work in either format. Thanks for the feedback.

PR #1035, opened 2023-01-28
Tetris Clock: Dumb and fast optimization

MarkGamed7794:
Quick hotfix that *should* make rendering twice as fast. If this causes any issues then please tell me.


MarkGamed7794:
I have since added some more slight optimizations:
- fade_colour now integer divides by 1 instead of using int(), which *might* be faster but shouldn't be any slower.
- The rendering process now merges same-coloured boxes in the same row into one large box. This seems to reduce rendering time by a bit.

PR #1039, opened 2023-01-29
Remove Twitter follows app

drudge:
Twitter has silently killed the endpoint that was powering the lookup.


betterengineering (approved):
😢 This app has some decent usage, which is sad for two reasons. One, people like it and two we have to remove it from their devices:
```

2023/02/01 15:31:40 app twitter-follows has 337 installations and is installed on 308 devices
```

drudge:
If you'd prefer, I can switch to using Oauth and the Twitter api directly but I'm assuming this would require users to reinstall?

betterengineering:
> If you'd prefer, I can switch to using Oauth and the Twitter api directly but I'm assuming this would require users to reinstall?

Yeah, it would. Best to clean out this app and recreate if you wish to keep it going!

PR #1045, opened 2023-01-31
Attempt 2: Tetris Clock optimization

MarkGamed7794:
Alright so Git is weird. But this one should work!

- App is now optimized to run over five-ish times faster.
- More customization options have been added.
- Digits are now randomized more.
- Colour schemes slightly altered.


betterengineering (approved):

betterengineering:
Thanks for the fixes @MarkGamed7794!

PR #1053, opened 2023-02-01
A new app that displays the bus arrival times at specified DC-area …

JetgirlDC:
…(WMATA) bus stop(s).


apps/dcbustimes/dc_bus_times.star
```diff
@@ -0,0 +1,257 @@
...
+                        Bus[5],
+                        Bus[6],
+                        Bus[7],
+                    ],
+                ),
+            ),
+        )
+
+def GetTimes1(stopID, apiKey):
+    cached = cache.get("arrTimes1")
```
matslina, line 211: Since StopID is included in the API request, I'm guessing the response body is specific to a single stop? If so, then the cache key probably also needs to include the StopID to make sure later calls to `GetTime1` aren't just returning the cached data from previous calls with (potentially) different StopID.

apps/dcbustimes/dc_bus_times.star
```diff
@@ -0,0 +1,257 @@
...
+                desc = "Bus Stop ID Number (7 digit number located on Bus Stop sign e.g. 1001155)",
+                icon = "busSimple",
+            ),
+            schema.Text(
+                id = "StopID_2",
+                name = "Stop ID #2",
+                desc = "Bus Stop ID Number (optional)",
+                icon = "busSimple",
+            ),
+            schema.Text(
```
matslina, line 250: The `desc` field is currently not visible to end users in the phone app, so if the WMATA Key is optional it could be good to mention that in the field name (which is visible to users). Alternatively, since there's an encrypted key already available, perhaps this config field isn't needed at all?

matslina (changes requested):
Thanks for the PR! Looks great overall. A few comments inline.

Also wanted to ask if you could make the app provide a default graphic for the case when no configuration is provided? That would allow users to see an illustrative preview of the app when searching or browsing.

JetgirlDC:
Thanks for your inputs.  Let me make sure I understand your requests/suggestions before getting started.

1) Also wanted to ask if you could make the app provide a default graphic for the case when no configuration is provided?

So for this, would simply making the API call with a “hard-coded” default StopID and rendering the results back to the user do what you are requesting?

2) ...the cache key probably also needs to include the StopID to make sure later calls to GetTime1 aren't just returning the cached data from previous calls with (potentially) different StopID.

Yeah good point.  I incorrectly attempted this using the two caches (GetTimes1 and GetTimes2), but was careless not just adding the StopID.  Good catch.

3) ...so if the WMATA Key is optional it could be good to mention that in the field name.

Certainly.  I was using Pixlet which did show the description field so I was unaware iPhone users would not see it.  Thanks for heads up.

4) ...since there's an encrypted key already available, perhaps this config field isn't needed at all?

I thought about this since I did have some trouble getting the key into/out of config.  Ultimately, however, I liked the option of being able to debug the app from pixlet which required it to be in config (or hard-coded in the app).  You’re correct though, it makes sense to leave it out and hardcode it in when/if I need to test app locally.

While I have you, let me ask you a question since I struggled mightily getting the original PR to work correctly (thank god someone mentioned using Git Desktop).  Once I make these changes, do I simply make a new PR from Git Desktop?

Thanks again for the good inputs,
Steven

____________________________ 
Steven Pressnall
Senior Signal Processing Engineer
Resonant Sciences
***@***.*** 
703.971.6504

matslina:
> 1) Also wanted to ask if you could make the app provide a default graphic for the case when no configuration is provided?
>
>So for this, would simply making the API call with a “hard-coded” default StopID and rendering the results back to the user do what you are requesting?

Either that, or hard coding some sample data to remove the need for calling the API.

> 4) ...since there's an encrypted key already available, perhaps this config field isn't needed at all?
>
>I thought about this since I did have some trouble getting the key into/out of config.  Ultimately, however, I liked the option of being able to debug the app from pixlet which required it to be in config (or hard-coded in the app).  You’re correct though, it makes sense to leave it out and hardcode it in when/if I need to test app locally.

Got it. Another option is to remove it from `get_schema()` but still check for it in `config` and then pass it in via a query parameter. E.g., when using `pixlet serve`, you'll see something like `?foo=bar` in the URL of your browser. If you manually change that to say `?foo=bar&baz=bat`, you should see `bat` in `config.get("baz")`, even though the `baz` field is not defined in `get_schema()`.

> While I have you, let me ask you a question since I struggled mightily getting the original PR to work correctly (thank god someone mentioned using Git Desktop).  Once I make these changes, do I simply make a new PR from Git Desktop?

Afraid I've no experience of Git Desktop. In general, you can just add and push commits to an existing PR, and it looks like you got that working! =)

apps/dcbustimes/dc_bus_times.star
```diff
@@ -0,0 +1,240 @@
...
+load("encoding/json.star", "json")
+load("http.star", "http")
+load("render.star", "render")
+load("schema.star", "schema")
+load("secret.star", "secret")
+
+NEXTBUS_URL = "https://api.wmata.com/NextBusService.svc/json/jPredictions"
+DEFAULT_STOPID1 = "1001155"
+DEFAULT_STOPID2 = ""
+ENCRYPTED_API_KEY = "AV6+xWcEd0axGb4V9l3oSg/6rC84PkFXGwE3RMdhCsd3HYbmgqigkaNaks24APTvZ9KujTGybupJ0DYlAW8rE4IL7WVYkPhMHxcyXQ2QgrCrKYMYaVt+201YoMHI2ImivHlhrv2kPk2O2wdxRxpMu4vEH+QFfwg73zQgTl499VRVwj/msko="
```
matslina, line 18:
```suggestion
ENCRYPTED_API_KEY = 
"AV6+xWcEN/eoKQ8PCUXypIgPSO7IT4v5uu6/yNmDXvJjtwrYANFMNpXiU/ki/yEvf7u9wpNjLLsx2ab7y9SFBL15w/elg7AkgtJVueJbhGGOwnlGyvAnhUEEvcDQh57x5vQT186xKDDUo/RgMEjSIj0bGnigi2hJUET8ydOFsz1DGbwtP6c="
```

matslina (approved):

PR #1059, opened 2023-02-02
New App: Oura Ring

acvigue:
App to view the three scores from your Oura ring and it also shows a historical graph of the scores over time.

oura_ring


apps/ouraring/ouraring.star
```diff
@@ -0,0 +1,193 @@
...
+
+def errorView(message):
+    return render.WrappedText(
+        content = message,
+        width = 64,
+        color = "#fff",
+    )
+
+def main(config):
+    if "apikey" not in config:
```
matslina, line 100: Would it be possible to provide a default graphic for the case when no configuration is provided? That would allow the Tidbyt iOS/Android app to show users an illustrative preview for users before they've dug up an API key.

apps/ouraring/ouraring.star
```diff
@@ -0,0 +1,193 @@
...
+
+    if "days" not in config:
+        return render.Root(child = errorView("no days"))
+
+    now = time.now()
+    from_date = (now - time.parse_duration(str(int(config["days"]) * 24) + "h")).format("2006-01-02")
+    to_date = now.format("2006-01-02")
+
+    sleep_data = None
+    sleep_dto = cache.get("oura_sleep_data")
```
matslina, line 111: The cache is scoped to your app, but not to individual users, so this call will simply return the most recently written cache record for this key (which could be from a different user). Including the API key in the cache key should be sufficient to solve this.

matslina: (The same applies to other locations in the code where the cache is used)

matslina (changes requested):
Thanks for the PR! A few comments inline.

apps/ouraring/ouraring.star
acvigue: > The cache is scoped to your app, but not to individual users, so this call will simply return the most recently written cache record for this key (which could be from a different user). Including the API key in the cache key should be sufficient to solve this.

Thanks! I'll add that in.

acvigue:
Think I took care of those changes. Thanks for the info about the cache keys. I'll keep that in mind for future apps.

matslina (approved):

PR #1063, opened 2023-02-03
Updates to soccer apps to match new format

cmarkham20:


jvivona:
@cmarkham20  - for soccermens & soccerwomens - you've removed a whole bunch of functionality (days forward, days back for api results) I have added in the app with this merge you did - plus for some leagues - the feed will cause the app to break - because there's no logo in the API results for some teams at lower levels of soccer (plus I've removed some display options for this reason as well) - these 2 apps aren't just copies of your app - there's more to it.    at this point, I'd rather revert it back (for my 2 apps) & I'll look at merging in the new league name display inside the code with the new functionality.

cmarkham20:
Sure, whatever you want do.

PR #1068, opened 2023-02-05
fix: Formula1 2023

jvivona:
worked with @AmillionAirs  on updates

- add readme & screenshots
 - add in new 2023 tracks & fix Japan
 - temp override URLs to bypass blacklist on API (will go back to original API after they unblock us )
 - remove location from schema - this was causing too many API hits - no need to use location anymore - use $tz now
 - only hit API endpoint we need to each time, instead of all 3
 - only execute render / vars needed for selected display option
 - change f1 API endpoints to https
 - added in US Date / Intl Date option - only triggered when showing Next Race
 - added in 24 hour / 12 hour display option - only triggered when showing Next Race


matslina:
Thank you for this!

PR #1069, opened 2023-02-06
Jokes

AndersHeie:
- Since I couldn't always read the end of very long jokes, I added some settings to control scroll speed, and to ask to scroll until the end. All settings are user configurable.
- Lint did some reformatting for me as well. Thank you Lint. You are a trooper.


rs7q5:
I don't know if the show full animation needs to be a config. I think it should show regardless and should be set to true...

I don't really know why this isn't the default option now that this is a render.Root option. Either way, for this app I can see NO reason why a user wouldn't want to see the whole joke...

AndersHeie:
I think it's not default just  so it won't get abused accidentally by apps that happens to pull long texts, not sure. I like to make these things configurable, but I could take that out and just make it default. I do like the scroll speed option.. whatever you think.

rs7q5:
Yeah I think `show_full_animation=True` always (no need for config) and then keep the scroll speed option. Is the default speed option the current speed on the app (just to keep consistent for current installations).

AndersHeie:
OK.  And yes it is the default of course :)

PR #1077, opened 2023-02-07
build: Add git reference printing.

betterengineering:
I'm working to add target determination so we only build what's changed on pull requests. In order to do that, I need to know what references are available on PRs. It's especially confusing because I need to know what the build system sees on forks. This commit adds a reference printing so I can see what PRs have available today.


betterengineering:
TL;DR - I'm going to rebase existing PRs off of this change to ensure I get this right before enabling it

PR #1080, opened 2023-02-07
build: Add printenv.

betterengineering:
This commit adds printenv so that I can see what environment variables exist in the builds we have today.


betterengineering:
I double checked: GitHub is smart here and does not print secrets 👌

PR #1093, opened 2023-02-08
New app: Womens T20 World Cup

M0ntyP:
Another cricket app!

This time for the upcoming Women's T20 World Cup, running Feb 10-Feb 26

Very similar to the other T20 apps I've done but a little smarter with showing upcoming match info


matslina:
Thanks for the PR!

PR #1099, opened 2023-02-08
Animation fix baby steps

rjison:
Add a scroll speed configuration option.  Also, set show_full_animation = True so people can get the full quotes. They may be a little long, but not so long that folks who installed this app will be annoyed. They can speed the scrolling up if they really want to speed through this app.


betterengineering:
Sorry about the icons! It was moved to pixlet: https://github.com/tidbyt/pixlet/blob/main/icons/icons.go

rjison:
No worries, thanks for the updated link. Some other folks were asking on discord, I’ll post a follow up with this! On Feb 8, 2023, at 2:14 PM, Mark Spicer ***@***.***> wrote:﻿
Sorry about the icons! It was moved to pixlet: https://github.com/tidbyt/pixlet/blob/main/icons/icons.go

PR #1122, opened 2023-02-11
New App: Octoprint

cbattlegear:
Recreation of a 3D printer LCD status screen using the Octoprint APIs.

Does basic API calls to self hosted octoprint, no caching as they are live updates and minimal data is used per call.

octoprint
This shows what it looks like if the url can't be reached or no info is added.

### DIFF
--- a/apps/octoprint/manifest.yaml
+++ b/apps/octoprint/manifest.yaml
@@ -1,0 +1,8 @@
+---
+id: octoprint
+name: Octoprint
+summary: Octoprint Status Screen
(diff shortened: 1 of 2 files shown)


matslina (changes requested):
Thanks for the PR.

We can unfortunately not merge apps that call out to user provided URLs at this time. This is so to protect against our platform being abused to spam or flood arbitrary hosts on the internet. We're hoping to have a solution in place later this year that allows for apps like this one, but we're unfortunately not there yet.

PR #1123, opened 2023-02-11
New app: Flights Nearby

eddichen:
This app uses the flight radar API to find the closest flight to your location and displays an airline tail graphic, the flight origin, destination, flight number and aircraft type.

There is a default API key associated which is limited to 500 calls a month, so I've added a field for users to add their own flight radar key.

I'm testing out using the location field, but could swap it out for latitude/longitude for more precise locations. Also more tail graphics can be added to encompass more airlines.

flights_nearby


eddichen:

apps/flightsnearby/flights_nearby.star
```diff
@@ -0,0 +1,244 @@
...
+
+def is_key_present(k):
+    if k in TAILS:
+        return TAILS[k]
+    else:
+        return TAILS["Q4"]
+
+def main(config):
+    api_key = secret.decrypt("AV6+xWcEFKIfCw67zTcYusGXoTGKrc1bOSIdg8X8UuTkEYMbEdcN67Lh1R8PxQUqFf3QxoXI39bIZz0LV8LgL11JY+/uSxEYPlrVMMEqzJMUJngnn9ZW31mmc5Hk9iNnwgfKAZ6YOadc86TTyR3Pyg78hbUfrE2brp3zYWRpsLrfJpxtIlX4UVlHtuC4qF87pQjyeLQ1wHw=")
+    location = config.get("location", DEFAULT_LOCATION)
```
matslina, line 133: The location field will hold a json encoded location object. Passing in a location currently causes a crash on L141.

apps/flightsnearby/flights_nearby.star
```diff
@@ -0,0 +1,244 @@
...
+
+    flight_cached = cache.get("flight")
+    if flight_cached != None:
+        print("Hit! Displaying cached data.")
+        flight = json.decode(flight_cached)
+    else:
+        print("Miss! Contacting Flight Radar")
+        centrePoint = [float(location["lat"]), float(location["lng"])]
```
matslina, line 141: We'll need latitude and longitude truncated to a lower precision to protect against leaking users exact location to the API provider. Check out [this](https://gis.stackexchange.com/questions/8650/measuring-accuracy-of-latitude-and-longitude) for some guidance on how many decimals to keep.

apps/flightsnearby/flights_nearby.star
```diff
@@ -0,0 +1,244 @@
...
+    if k in TAILS:
+        return TAILS[k]
+    else:
+        return TAILS["Q4"]
+
+def main(config):
+    api_key = secret.decrypt("AV6+xWcEFKIfCw67zTcYusGXoTGKrc1bOSIdg8X8UuTkEYMbEdcN67Lh1R8PxQUqFf3QxoXI39bIZz0LV8LgL11JY+/uSxEYPlrVMMEqzJMUJngnn9ZW31mmc5Hk9iNnwgfKAZ6YOadc86TTyR3Pyg78hbUfrE2brp3zYWRpsLrfJpxtIlX4UVlHtuC4qF87pQjyeLQ1wHw=")
+    location = config.get("location", DEFAULT_LOCATION)
+
+    flight_cached = cache.get("flight")
```
matslina, line 135: The cache is scoped to individual apps, but not to individual users. This means that all users of the app will retrieve and write to the same cache record. You need to add something to the cache key that is unique for this API request. I suspect the (truncated) lat-lng is a good idea.

apps/flightsnearby/flights_nearby.star
```diff
@@ -0,0 +1,244 @@
...
+        print("Hit! Displaying cached data.")
+        flight = json.decode(flight_cached)
+    else:
+        print("Miss! Contacting Flight Radar")
+        centrePoint = [float(location["lat"]), float(location["lng"])]
+        boundingBox = get_bounding_box(centrePoint, config.get("distance", DEFAULT_DISTANCE))
+        rep = http.get(
+            FLIGHT_RADAR_URL,
+            params = {"bl_lat": boundingBox[0], "bl_lng": boundingBox[1], "tr_lat": boundingBox[2], "tr_lng": boundingBox[3]},
+            headers = {"X-RapidAPI-Key": config.get("key", api_key), "X-RapidAPI-Host": "flight-radar1.p.rapidapi.com"},
```
matslina, line 146: It's possible the Tidbyt mobile app will pass in an empty string when `key` is not set by the user. It would probably be a good idea to get `key` and check it against both `""` and `None` to determine if the default key should be used.

matslina (changes requested):
Thanks for the PR! There are a couple of issues that need to be addressed before we merge, but nothing that looks like it'll be too difficult to fix.

apps/flightsnearby/flights_nearby.star
```diff
@@ -0,0 +1,260 @@
...
+def is_key_present(k):
+    if k in TAILS:
+        return TAILS[k]
+    else:
+        return TAILS["Q4"]
+
+def reduce_accuracy(coord):
+    coord_list = coord.split(".")
+    coord_remainder = coord_list[1]
+    if len(coord_remainder) > 4:
```
matslina, line 134:
```suggestion
    if len(coord_remainder) > 3:
```

apps/flightsnearby/flights_nearby.star
```diff
@@ -0,0 +1,260 @@
...
+        coord_remainder = coord_remainder[0:4]
```
matslina, line 135:
```suggestion
        coord_remainder = coord_remainder[0:3]
```

matslina (approved):
Looks good! I'll nudge the lat/lng truncation to 3 decimals, for ~100m resolution. I imagine it's more than enough for this use case. Let us know if you disagree!

PR #1125, opened 2023-02-11
New app: SpaceX Launch

rytrose:
This PR introduces a new app "SpaceX Launch" that displays information about an upcoming SpaceX rocket launch.

- Can we expect the app to work reliably over time?
  - The [Launch Library 2 API](https://thespacedevs.com/llapi) has been around a few years, and seems to have a healthy community ([Reddit](https://www.reddit.com/r/TheSpaceDevs/), [Discord](https://discord.com/invite/p7ntkNA)).
- Is the app handling user data responsibly?
  - There's no real user data apart from API key, but that's passed in as config and not persisted outside of how Tidbyt persists configs.
- Is the app caching in an appropriate manner?
  - I believe so, it caches a single API resource and image for 5 minutes to comply with the API's [15 req/hr rate limit](https://github.com/TheSpaceDevs/Tutorials/blob/main/faqs/faq_LL2.md#is-ll2-free-to-use).
- If an external API is being used, is the app taking care to keep request rates at a reasonable level?
  - Yes, it makes one request every 5 minutes.


rytrose:

rytrose:
recheck

matslina:
Thanks for the PR. A little bit worried we'll exceed that low rate limit. When a cache record expires there can potentially be several instances of the app trying to refresh at the same time, resulting in a higher request rate than one might expect. But let's give it a try and see what happens.

rytrose:
Hey @matslina thanks for giving it a try! Yeah that makes sense, didn't think about that. 

I think the app could still have value even if updated e.g. once an hour, so if you do see issues perhaps we can just bump out the cache duration to much longer.

PR #1127, opened 2023-02-11
Vital Occupancy: new pixelated logo and colors

flip-z:


apps/vitaloccupancy/vital_occupancy.star
```diff
@@ -37,14 +39,34 @@ def main():
         logo = http.get(LOGO_URL).body()
         cache.set("logo", logo, ttl_seconds = 300)
```
matslina, line 40:
```suggestion
        cache.set("logo", logo, ttl_seconds = 3600)
```

PR #1133, opened 2023-02-13
shopify partners app

RishabhTayal:


RishabhTayal:

apps/shopifypartners/shopify_partners.star
```diff
@@ -0,0 +1,375 @@
...
+"""
+
+APP_ID = "shopify_sales"
+
+def api_fetch(partnermetricsCookie, request_config):
+    cache_key = "{}/{}".format(APP_ID, base64.encode(json.encode(request_config)))
```
matslina, line 205: Since `request_config` only holds the `now` field, which is set to today's date, this cache key will be the same for all users of the app. I think you'll want to include the `partnerMetricsCookie` in the key as well, to avoid one user reading another users data?

apps/shopifypartners/shopify_partners.star
```diff
@@ -0,0 +1,375 @@
...
+load("encoding/json.star", "json")
+load("http.star", "http")
+load("render.star", "render")
+load("schema.star", "schema")
+load("time.star", "time")
+
+# CONFIG
+SHOPIFY_COUNTER_API_HOST = "https://www.partnermetrics.io/chart_data?chart_type%5Bcalculation%5D=sum&chart_type%5Bcolumn%5D=revenue&chart_type%5Bdirection_good%5D=up&chart_type%5Bdisplay%5D=currency&chart_type%5Bmetric_type%5D=recurring_revenue&chart_type%5Btitle%5D=Revenue&chart_type%5Btype%5D=recurring_revenue&date="
+
+CACHE_TTL = 30
```
matslina, line 20: Would it be OK to use a (much) larger TTL?

apps/shopifypartners/manifest.yaml
```diff
@@ -0,0 +1,8 @@
+---
+id: shopify-partners
+name: Shopify Partners
```
matslina, line 3: I think we'll need to consider renaming this app before merging. It's not using Shopify's official API, so perhaps it'd be more appropriate to name it "Partnermetrics", or something like that?

matslina (changes requested):
Thanks for the PR! A few comments inline. Would also be great if you can add a default graphic for the case when no configuration is provided. That would enable users to look at a preview before deciding if they want to head over to partnermetrics.io to get their cookie and so on.

rishabhtpdl:
@matslina Thanks for reviewing the PR. I will take care of the comments.

RishabhTayal:
@matslina Thanks for reviewing the PR. I will take care of the comments.

apps/shopifypartners/shopify_partners.star
RishabhTayal: Increased to 3600.

apps/shopifypartners/shopify_partners.star
RishabhTayal: added `partnerMetricsCookie` to `cache_key`

apps/shopifypartners/manifest.yaml
RishabhTayal: The app doesn't use Shopify partner API however it is displaying data from shopify parter dashboard. I think this name is appropriate based on the data displayed.

apps/shopifypartners/shopify_partners.star
```diff
@@ -0,0 +1,375 @@
...
+                        ),
+                    ],
+                ),
+            ],
+        ),
+    )
+
+def main(config):
+    partnermetricsCookie = config.get("partnermetricsCookie")
+    timezone = config.get("timezone") or "America/New_York"
```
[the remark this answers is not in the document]
RishabhTayal: updated

RishabhTayal:
@matslina The PR is ready to review again.

RishabhTayal:
@matslina can you please review this again? Would love to get it merged.

apps/shopifypartners/manifest.yaml
matslina: Ok, I understand. However, we do need the name changed to merge. It's important for us to minimize any confusion with the official Shopify apps, since we've plenty of new users coming into the ecosystem specifically to use these.

RishabhTayal:
Thanks for merging. When are the applets published for community to use?

matslina:
Already available for beta users. Available for regular users when we've checked it out and hit the launch button.

RishabhTayal:
How can I signup as beta user? I would like to test it out as well. :)

matslina:
https://forms.gle/zGqUyCv9kUh6YUic9
We import the form data in bulk every few weeks, but if you get it in quickly I'll add you for this one app straight away.

RishabhTayal:
On it. 🏃

RishabhTayal:
Done.

matslina:
Restart the Tidbyt app and it should appear. Feel free to hit me up on our discord as well if you have any questions.

matslina:
Hey there @RishabhTayal 

Sorry to resurrect such an old PR. We heard from Shopify that there's some concern that this app's use of official looking Shopify assets, like the alien and the Shopfiy "bag", might cause confusion for users and give the impression that this is a "by Shopify" app.

I'd like to ask if you're able to submit a PR that replaces all these assets with new ones that more clearly conveys to users that the app is independent of Shopify? It would be greatly appreciated.

Otherwise we may have to take the app down for now. It would of course still be possible to run the app as a Private App.

LMK.

PR #1148, opened 2023-02-20
Add files via upload

Matt-Pesce:
Adds error message to the Tidbyt display when Auth failure (usually due to credentials being revoked).  Cleaned up render display function to avoid "fail" scenario.

### DIFF
--- a/apps/outlookcalendar/outlook_calendar.star
+++ b/apps/outlookcalendar/outlook_calendar.star
@@ -125,6 +125,7 @@ def main(config):
 
     calendar_start_time = current_time.format(RFC3339_FORMAT)
     calendar_end_time = midnight_time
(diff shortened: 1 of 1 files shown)


apps/outlookcalendar/outlook_calendar.star
```diff
@@ -198,7 +204,7 @@ def main(config):
             print(conflict_meeting_banner)
     else:
         conflict_meeting_banner = "No More Meetings for Today!"
-        display_calendar_date = time.parse_time(calendar_start_time).in_location("America/Detroit").format("Jan 2")
+        display_calendar_date = time.parse_time(calendar_start_time).in_location("America/Detroit").format("Jan 2")  # BUG HERE: Fix the Timezone (it's hardcoded)
```
matslina, line 207: There's a "magic" config variable `$tz` when running in our backend. That holds the device's configured timezone.

PR #1149, opened 2023-02-20
Goodreads yearly reading challenge progress tracker

panderson54:
### About:
This app displays the progress of the user towards their yearly reading challenge goals as tracked by the Goodreads site. Goodreads is integrated into kindle devices so the desired workflow is someone finishes a book on their kindle and automatically sees the progress on their tidbyt within one day. 

#### Display:
The app displays different graphics tracking your progress from 0 - 13 books, the average person reads 12 books a year so this seemed a natural stopping point. I may at a later point add more animations or some sort of dynamic image. 

bookdemo

#### Getting Data:
Goodreads had killed off its public API in 2020. Given what I am calling a challenge ID (not a user ID nor as far as I can tell derived from your users ID) https://www.goodreads.com/user_challenges/{challenge_id} will serve up any public profile's yearly reading challenge. The app makes a GET request to https://www.goodreads.com/user_challenges/{challenge_id and parses the returned HTML using starlark's regex libary to pull reading progress and goals from a div entitled 'progressText'.

Parsing HTML is more fragile than making a real API call, this page has been stable and Goodreads is not frequently updated but this is obviously more fragile than a proper API call. Should Amazon/Goodreads allow public API calls again I will update this app to use a more stable endpoint.


panderson54:

apps/readingchallenge/reading_challenge.star
```diff
@@ -0,0 +1,158 @@
...
+    )
+
+def get_schema():
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Text(
+                id = "user_challenge_id",
+                name = "Challenge ID",
+                desc = "Navigate goodreads.com/user_challenges and get this from the URL",
```
matslina, line 154: This description field is unfortunately not visible to users in the Tidbyt smartphone app, so you probably want to add a note about this to the app description.

apps/readingchallenge/reading_challenge.star
```diff
@@ -0,0 +1,158 @@
...
+
+    return images
+
+def main(config):
+    CHALLENGE_ID = secret.decrypt("AV6+xWcE/k5WZk3zYbxM/fuI6z4DJ5oQUNrHWbIu8HBOyeubTAzYcB9DxEMAWRoxDcryQFohadBlZneiSq94urEiiG53bpXexcibIvOyKOSH3Vi3qEYaY1vVXUxHQOmyk5odoCU9UtLX850GqOI=")
+
+    if CHALLENGE_ID == None:
+        CHALLENGE_ID = config.str("user_challenge_id", "0")
+
+    progress_cached = cache.get("progress")
```
matslina, line 105: The cache is shared by all users of the app. This will retrieve and use the most recently written progress record, regardless of what user create it. You probably want to include the challenge_id in the cache key.

matslina (changes requested):

apps/readingchallenge/reading_challenge.star
panderson54: Great, can do.

apps/readingchallenge/reading_challenge.star
panderson54: That makes sense, thanks for the clarification

apps/readingchallenge/reading_challenge.star
```diff
@@ -0,0 +1,158 @@
...
```
matslina, line 105:
```suggestion
    progress_cached = cache.get("".join(["progress", CHALLENGE_ID]))
```

apps/readingchallenge/reading_challenge.star
```diff
@@ -0,0 +1,158 @@
...
+    goal_cached = cache.get("goal")
```
matslina, line 106:
```suggestion
    goal_cached = cache.get("".join(["goal", CHALLENGE_ID]))
```

matslina (approved):

PR #1150, opened 2023-02-20
New App: FlightTracker

SamuelSagarino:
**About**
This app integrates with the FlightAware API, allowing users to input their FlightAware API key and track a specific flight's progress or display the recent arrivals or departures from a specified airport.

**Visual**
The top color bar changes based upon the flight status (onetime, delayed, cancelled, etc) - once a flight has taken off it also functions as a progress bar for the flight's completion.

The second section of information contains the departure and arrival cities and airport codes; as well as the scheduled time (or actual time) of departure or arrival (what time is displayed is determined based upon the current flight status; ie - if a flight has departed it won't show the scheduled departure; rather the actual).

The airline logos are sourced from the FlightAware library of logos that can be fetched utilizing the operator code provided by the API response.

The first marquee will change color based upon the current flight status; much like the top bar - and based upon the flight status it will display the next milestone in the flight's journey. (For instance after prior to takeoff it will show "Scheduled to Depart in x minutes", etc)

The lower marquee will display the flight number, aircraft registration and the aircraft type.

**Caching**
Due to the nature of tracking arrivals / departures and live flight data, the caching is currently set to just 60 seconds. Caches are stored depending upon whether or not you are viewing airport arrivals / departures or a specific flight so that if multiple users are tracking the same element the cache will be available for their use.

(Example: caches for arriving/departing flights are stored as `arrivals/KMCO` or `departures/KATL` or if you're tracking a flight it is stored as `flight/DAL093`)

flighttracker


SamuelSagarino:

SamuelSagarino:
Update on 2/23 - Saw notes on some other PRs about caching and how it is not individual to the user. Made some adjustments to account for this and correct it.

apps/flighttracker/flighttracker.star
```diff
@@ -0,0 +1,479 @@
...
+
+    actualArrival = time.time(year = year, month = month, day = day, hour = hour, minute = minute, second = second, location = "Europe/London")
+
+    actualArrival_humanized = humanize.time(actualArrival)
+
+    actualArrival_time = humanize.time_format("HH:mm", actualArrival.in_location(destinationTimezone))
+
+    #######
+
+    logo = http.get("https://flightaware.com/images/airline_logos/90p/" + operator + ".png").body()  # Get logo to display.
```
matslina, line 253: Would be great with some caching around this call.

matslina (changes requested):
Thanks for the PR! Need just a little bit of additional caching. Also, could you add a default graphic for the case when no config is provided? That would allow users to see a preview and get a sense of whether or not this is an app they're willing to go get an API key for.

SamuelSagarino:
@matslina Added caching for the flight data; up to 120 seconds (if you think it should be more is there opposition to making it be a user definable cache length?), added caching for the airline images and added a resting look when an API key isn't present.

matslina (approved):

matslina:
120 is probably a good TTL. I don't think there's any reason to make that configurable yet, unless it turns out to be an issue down the road.

PR #1151, opened 2023-02-20
New App: NSByt app 

tim-hanssen:
This app displays a real-time timetable for any train station in The Netherlands. It shows the departure time, delays and disruptions and destination of the next two trains that are leaving the chosen station.

The station can be set with the station parameter. Additionally, with the parameter skiptime, it is possible to show departures after the set skip time in minutes. If the destination station parameter is set, a trip is calculated and from the first two, the first leg of the trip will be shown instead al all trains.


tim-hanssen:

PR #1165, opened 2023-02-25
Feature: World Clock

M0ntyP:
I've added a feature to the World Clock app that gives an option to show the temperature for the selected location, either Celsius or Fahrenheit. The registered API can do up to 1 million calls per month

I've never tried to add a feature to an existing app before so hopefully I've done this correctly!

Also, never used secret.decrypt before, again hope that works too!


rs7q5:
This is super cool. Excited for this feature! cc @ejb for visibility though.

ejb:
So great! I've always wanted to add a feature like this. Thank you!

A couple of thoughts on the design:
- Is it possible to right-align the temperatures so they line up nicely?
- How does it look with long locale names? Do they truncate properly? (I can't even remember if or how that works..)

And what happens if the API is down or over the rate limit?

M0ntyP:
Thanks for the positive feedback! 

I can look into right-align for the temperatures and yes, I should add something if there is an API failure

With the long locale names, its set to marquee so long names will scroll across slowly as the animation changes. I can keep it the same, or look at truncating if its over a certain character length. I'll leave it up to you as the original author

apps/worldclock/world_clock.star
```diff
@@ -103,59 +122,121 @@ def main(config):
...
-                location_time,
-            ],
-        )
         rows.append(row)
         if (i < len(locations)):
             rows.append(horizonal_rule)
 
     return render.Root(
-        delay = 500,
-        max_age = 120,
```
matslina, line 158: Why remove `max_age`?

matslina:
Note that with 60s TTL on cached records, a single Tidbyt will produce 60*24*30=43200 requests over 30 days. That means 24 Tidbyts using the feature will be sufficient to exhaust the API quota. =(

apps/worldclock/world_clock.star
M0ntyP: Not sure, I must have deleted it accidentally, sorry! I will add it back in

M0ntyP:
> Note that with 60s TTL on cached records, a single Tidbyt will produce 60_24_30=43200 requests over 30 days. That means 24 Tidbyts using the feature will be sufficient to exhaust the API quota. =(

M0ntyP:
The weather part has a 30 min cache, could even go to 1hr

I'm hoping to get some time this weekend to review the suggestions - right-align and API error handling

M0ntyP:
I haven't forgotten about this one, its still on my to do list. Just need to put aside some time to get it fixed up.

PR #1168, opened 2023-02-26
Airthings: add option 'onlyDisplayNotNormal'

joshspicer:
Adds an option to the 'Airthings' applet to only display "not normal" values (values that are in the yellow or red range).

Example:

image

### DIFF
--- a/apps/airthings/airthings.star
+++ b/apps/airthings/airthings.star
@@ -154,7 +154,7 @@ def main(config):
             continue
 
         if onlyDisplayNotNormal and not displayName in nonNormalValues:
-                continue
+            continue
 
         items.append(
             render.Row(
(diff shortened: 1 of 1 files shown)


apps/airthings/airthings.star
```diff
@@ -235,19 +242,19 @@ def get_schema():
             schema.Text(
                 id = "clientSecret",
                 name = "AirThings API Client Secret",
-                desc = "API secret from https://dashboard.airthings.com/integrations/api-integration",
+                desc = "REQUIRED: API secret from https://dashboard.airthings.com/integrations/api-integration",
```
matslina, line 245: Note that these field descriptions strings are not visible to the end user in the iOS/Android app.

PR #1171, opened 2023-02-28
New app: Literature clock

alyshakwok:
This app displays the time using a quote from a piece of literature. It is based on work and idea by Jaap Meijers and Jene Voldsen.

It includes a toggle setting to hide any nsfw quotes (ie: quotes containing profanities).

If several quotes exist for the given time, one will be chosen at random. If it is currently a time for which there are no quotes or no sfw quotes (if setting is turned on), then will search backwards one minute until a quote is found.  

literature_clock


alyshakwok:

apps/literatureclock/literature_clock.star
```diff
@@ -0,0 +1,180 @@
...
+                ),
+        ),
+    )
+
+def get_schema():
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Location(
+                id = "location",
```
matslina, line 113: If you only need the device's current timezone, that's going to be available as the somewhat magic `$tz` config parameter when the app runs in the Tidbyt backend.

apps/literatureclock/literature_clock.star
```diff
@@ -0,0 +1,180 @@
...
+    file = JSON_ENDPOINT
+    if fileTime.hour < 10:
+        file += "0"
+    file += str(fileTime.hour) + "_"
+    if fileTime.minute < 10:
+        file += "0"
+    file += str(fileTime.minute) + ".json"
+
+    # check if quotes is not empty, else one min back
+    request = http.get(file)
```
matslina, line 45: It'd be great if you could add some caching around this call, to keep request rates down. Check out the `cache.star` module.

matslina (changes requested):
Thanks for the PR. Looks great. Just need some caching added and this should be good to go!

apps/literatureclock/literature_clock.star
alyshakwok: Thanks! Didn't know about this config. Code has been updated to leverage $tz.

apps/literatureclock/literature_clock.star
alyshakwok: Added in some caching!

matslina (approved):

PR #1173, opened 2023-03-01
Splatoon 3 Stages: Add Um'ami Ruins and Manta Maria stage images, fix missing stage fallback

MarkGamed7794:
At this point I should just add fallbacks for my fallbacks. Sorry for all the PRs.


matslina:
We like PRs!

PR #1177, opened 2023-03-01
Animation to scores apps

cmarkham20:


apps/ncaasoftscores/manifest.yaml
```diff
@@ -0,0 +1,8 @@
+---
+id: ncaasoft-scores
+name: Softball Scores
+summary: NCAA Softball scores
+desc: Creating multiple app instances is no longer necessary. Just install once and remove any other instances. You can set the rotation speed of scores in the settings.
```
matslina, line 5: Thinking this description might be redundant for a completely new app?

cmarkham20: Yeah, could be at this point, though there may be some people used to the old app instances. Maybe once the new apps have been live for say a month I can update all descriptions?

matslina: Sure, WFM. =)

PR #1183, opened 2023-03-03
New app: Airbnb Calendar

jed:
This app is a calendar that displays the coming week of reservations for multiple Airbnb listings, as pulled from Airbnb's iCal export URLs. It looks like this[^1]:

image

iCal URLs that do not start with `https://www.airbnb.com/calendar/ical/` are rejected to prevent abuse, and successfully fetched results are cached for five minutes. Please message me if you'd like to test using my private Airbnb iCal URLs.

This was a fun thing to build, thanks to the Tidbyt team for an enjoyable platform! Here are a few ideas that would help make building things like this easier:
- I'd love a way of selecting a timezone without using the `Location` widget, since I don't need lat/lon or other user information. Is the list of timezones available in the runtime somewhere, to populate a `Dropdown`?
- It would be great if the `Font` widget accepted embedded `.bdf` files. I ended up rendering my own 3x3 glyphs using `Box`.

[^1]: Pardon the `pixlet serve` screenshot, `pixlet render` seems to be barfing on url parameters, and my Tidbyt order hasn't arrived yet...

### DIFF
--- a/apps/airbnbcalendar/airbnb_calendar.star
+++ b/apps/airbnbcalendar/airbnb_calendar.star
@@ -6,7 +6,6 @@ Author: Jed Schmidt
 """
 
 load("cache.star", "cache")
-load("encoding/json.star", "json")
 load("http.star", "http")
 load("humanize.star", "humanize")
 load("math.star", "math")
--- a/apps/airbnbcalendar/airbnb_calendar.star
+++ b/apps/airbnbcalendar/airbnb_calendar.star
@@ -26,7 +26,7 @@ COLORS = {
     "PAST": ("#333", "#333"),
     "FUTURE": ("#aaa", "#666"),
     "PRESENT": ("#ff5a5f", "#ff5a5f"),
-    "BLOCKED": ("#666", "#000"),
+    "BLOCKED": ("#666", "#111"),
 }
 
 LETTERS = {
--- a/apps/airbnbcalendar/airbnb_calendar.star
+++ b/apps/airbnbcalendar/airbnb_calendar.star
@@ -99,13 +99,12 @@ def listing(url, height):
         children.extend([
             render.Padding(
                 pad = (left, 0, 0, 0),
-                child = box(width, height - 1, stroke)
+                child = box(width, height - 1, stroke),
             ),
-
             render.Padding(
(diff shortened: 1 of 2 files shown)


jed:

rs7q5:
Do you only need the local timezone or does the user need to choose the location? If so, you can use `timezone = config.get("$tz", DEFAULT)` to get the time zone without using location. `$tz` is an internal parameter on Tidbyt's side, so when rendering locally you will always get DEFAULT. But in production it would return something like `America/New York`.

jed:
Indeed, [this](https://github.com/tidbyt/community/commit/05878464990ac24793fb65af1db38ee22e6a6fb5#diff-614a8b1d7931f2877e6dced1d86deb3f074509fbfb1837ff80458c57175a2198R197) is what I'm doing now. Would be nice to let the user specify, but it's not a dealbreaker by any means.

rs7q5:
I mean honestly a lot of apps use `schema.Location` to get the timezone and don't use the other information. I mean you could pre-populate all of the popular timezones, but I feel like just letting them enter it would make most sense. I mean could you have it be chosen between the users current timezone or the timezone of the reservation? I presume the assumption here is they are your housing and you are listing them on airbnb. 

https://pkg.go.dev/time you'll likely be able to find info here and the file for loading the timezones. I think it just uses the tz database, so I'm sure there's an easier way to access the list.

jed:
Quick question for @matslina: if I wanted to make this more a more general app for anything iCal (not just Airbnb), what are the minimum requirements for allowing users to enter arbitrary URLs (ie, to avoid malicious DoS usage)?

matslina:
Hey @jed. Thanks for the PR. Looks good!

As for arbitrary URLs, we'd need the app to construct the URL from user input so that it's certain to match some predefined set of target hosts. It's a bit of a bummer as it blocks a bunch of interesting user cases. I hope we'll be able to support completely arbitrary URLs in the not too distant future, but we're not there yet.

jed:
I totally understand. It might make sense to create a repo of allowlisted URL patterns / frequencies, or something lightweight that's decoupled from the usual PR process (or whatever process you're currently migrating to. Thanks!

PR #1188, opened 2023-03-06
Create New Jersey PATH app

karmeleon:
# NJ PATH
new_jersey_path

Based closely on the many existing train arrival apps, allows Tidbyt owners to select their favorite PATH station and optionally a train direction to view arrival times for. I pull data from the PATH API, and it's got a 60 second timeout so it should stay up to date (though I don't actually own a tidbyt so I can't test that).


karmeleon:

karmeleon:
recheck

apps/newjerseypath/new_jersey_path.star
```diff
@@ -0,0 +1,238 @@
...
+
+    # sort the routes so the one with the soonest arrival is first
+    routes_ordered = sorted(routes_ordered, key = lambda route: route["arrivalTimes"][0])
+    return routes_ordered
+
+def main(config):
+    station = config.get("station") or "grove_street"
+    desired_direction = config.get("direction") or "both"
+    path_url_for_station = PATH_URL.format(station = station)
+    api_response = http.get(path_url_for_station)
```
matslina, line 152: We'll need some caching around this call to keep request rates down.

matslina (changes requested):

apps/newjerseypath/new_jersey_path.star
karmeleon: Makes sense, added a per-station cache with 30s TTL

PR #1189, opened 2023-03-06
[Enphase Solar] App to display energy generated by solar panels via Enlighten API v4

laodaochong:
## Summary

Tidbyt App to display energy generated by Solar Panels via Enlighten API v4.

I have been running this app on my local laptop and pushing to my Tidbyt device for about two weeks. I really want to share this app and I can run the app without keeping my laptop on.

Please refer to README about how to set up the application and necessary credentials.

## Expected output

Screenshot 2023-03-05 at 10 41 26 PM


laodaochong:

laodaochong:
Hi Team, can you help merge this PR? Please let me know if you have any questions. Thanks.

apps/enphasesolar/enphase_solar.star
```diff
@@ -0,0 +1,174 @@
...
+            schema.Text(
+                id = "access_token",
+                name = "Access Token",
+                desc = "Follow the App README to generate the access_token",
+                icon = "key",
+            ),
+            schema.Text(
+                id = "refresh_token",
+                name = "Refresh Token",
+                desc = "Follow the App README to generate the refresh_token",
```
matslina, line 170: These descriptions are unfortunately not visible to end users in the iOS/Android app. Could you update the the app description in the manifest to hold the relevant information for users to get going with the app?

apps/enphasesolar/enphase_solar.star
```diff
@@ -0,0 +1,174 @@
...
+    elif response.status_code == 401 and EXPIRE_MSG in response.json().get("details"):
+        print("Access token is expired and refresh token now. Call next time.")
+        return 401, None
+    else:
+        print("Request is failed with status code {} {}".format(response.status_code, response.json()))
+        return response.status_code, None
+
+def get_system_stats(api_key):
+    headers = {
+        "Authorization": "Bearer " + cache.get("access_token"),
```
matslina, line 35: Cache is scoped to the app, but not to individual users. You'll need to update all cache keys so that they're unique to the user/configuration.

apps/enphasesolar/enphase_solar.star
```diff
@@ -0,0 +1,174 @@
...
+        return render_msg("Switch is off, so take a rest.")
+
+    access_token = config.str("access_token")
+    refresh_token = config.str("refresh_token")
+    client_id = config.str("client_id")
+    client_secret = config.str("client_secret")
+    api_key = config.str("api_key")
+
+    if not all([access_token, refresh_token, client_id, client_secret, api_key]):
+        return render_msg("Missing credential information.")
```
matslina, line 100: It'd be great if you could provide a default graphic for the case when no configuration is provided. That would allow users to see a preview of the app, before they decided to go looking for all the require config to set it all up.

matslina (changes requested):
Thanks for the PR! A few comments in line that we need addressed before merging.

laodaochong:
Hi @matslina thanks for your review! I will address those comments 😄 soon.

laodaochong:
Hi @matslina I have addressed your comments. Please take another look.

apps/enphasesolar/enphase_solar.star
laodaochong: I am not sure how to provide a default graphic, but I add more information in the msg rendered including what information this app provides.

matslina: The way most apps do it is to check if the config params are present, and if they're not just use some mock data to render what the app would look like if it had config. Could be good to add that in a later PR, if you're up for it. =)

laodaochong: Aha, gotcha. I will address this in next PR! We are getting over 50 KWH from the solar panels now, still cold in Boston. Super excited to get the this app running on Tidbyt device without running my code from my laptop and push to the device using a script.

PR #1191, opened 2023-03-06
New app: SEPTA Transit

radiocolin:
This app displays departure times for SEPTA buses, trolleys, MFL/BSL, and NHSL in and around Philadelphia.
septatransit

Users can select a route and stop and see all upcoming departures. The app uses SEPTA's API and implements sensible caching of schedules.


matslina:
Thanks for the PR!

PR #1192, opened 2023-03-06
New app: F1 Results

M0ntyP:
With the start of the new F1 season, I've made an app that shows qualifying and race results for the latest round with some additional options to show for the race results. 

Note that this is not a live timing app! It just shows the results after the qualifying session or race is over

Credit to @AmillionAirs and @jvivona for the idea and providing a very handy F1 API


jvivona:
wanted to give you a heads up - the ergast folks still have the Tidbyt servers (IP) on their blacklist & so when this goes into production - you aren't going to get any data.    I've been working with him on showing that the new versions of the app(s) aren't going to flood his server (it's seems like it's literally 1 guy who does this).    For right now I've been caching the data locally on one of our servers & pointing the apps to our servers.

M0ntyP:
Ah that's a shame! 

My calls are once an hour just to check if the race is complete. Then if its more than 48hrs after the race has completed it calls the data for the next race and caches that for 24hrs. So this doesn't poll his API that much. His own responsible use policy says no more than 200 times per hour.

Anyway I might put this on hold then until you can assure him we're not bad guys!

Thanks for letting me know

jvivona:
I'm going to ping you on Discord...

M0ntyP:
FYI for the approvers - @jvivona has kindly offered to host the APIs I need as they have done for the other F1 app. So this one should be ready to go

matslina:
Thanks for the PR!

PR #1198, opened 2023-03-08
New app: "Langton's Ant" (cellular automata)

alewando:
An app to run the Langton's Ant automata on Tidbyt. Supports multiple ants and rule sets.

See: [Langton's Ant (Wikipedia)](https://en.wikipedia.org/wiki/Langton%27s_ant)


alewando:
Closing this PR, the commits are associated with an incorrect account

PR #1199, opened 2023-03-08
New app: "Langton's Ant" (cellular automata)

alewando:
An app to run the Langton's Ant automata on Tidbyt. Supports multiple ants and rule sets.

See: [Langton's Ant (Wikipedia)](https://en.wikipedia.org/wiki/Langton%27s_ant)
langtons_ant


alewando:

apps/langtonsant/langtons_ant.star
```diff
@@ -0,0 +1,254 @@
...
+
+def log(message, vars = None):
+    if DEBUG_ENABLED:
+        if not vars:
+            print(message)
+        else:
+            print(message % vars)
+
+def main(config):
+    num_frames = int(config.get("num_frames", DEFAULT_FRAMES_PER_VIEW))
```
matslina, line 66:
```suggestion
    random.seed(time.now().unix // 15)
    num_frames = int(config.get("num_frames", DEFAULT_FRAMES_PER_VIEW))
```

apps/langtonsant/langtons_ant.star
```diff
@@ -0,0 +1,254 @@
+"""
+Applet: Langton's Ant
+Summary: 2-D Cellular Automata
+Description: Automata with simple rules but complex emergent behavior.
+Author: alewando
+"""
+
+load("random.star", "random")
+load("render.star", "render")
+load("schema.star", "schema")
```
matslina, line 10:
```suggestion
load("schema.star", "schema")
load("time.star", "time")
```

matslina (changes requested):
Thanks for the PR! The animation is very pretty!

Suggested a change to make it all a little bit less "random". This would provide new graphics over time, while still allowing us to cache them well in the Tidbyt backend.

Also a bit concerned about performance. It takes several seconds to render, which may be a bit of a problem for us. If there are any optimizations you can think of then that'd be very welcome.

alewando:
I looked at another similar app (`life`), and saw that it uses (and reuses) predefined `Cell`s instead of creating lots of new objects. I implemented a similar scheme and it helped a lot. 
`pixlet profile` went from 9s to 4.5s, so about a 50% improvement.

apps/langtonsant/langtons_ant.star
dinosaursrarr: I had similar issues with my game of life app. Creating and reusing a single `Box` pixel for each colour, rather than creating a new `Box` for every pixel made a big difference.

https://github.com/tidbyt/community/pull/584

alewando: Thank you for the Conway's Life app @dinosaursrarr! Seeing it in the app list was definitely an inspiration for submitting another cellular auotmata. And then seeing your performance solution was brilliant, too!

dinosaursrarr: @alewando Inspired by your PR to make this one: https://github.com/tidbyt/community/pull/1240

dinosaursrarr: @alewando I went one further and updated the PR with an second app for 1D, three-state CAs.

matslina (approved):

PR #1202, opened 2023-03-08
Nearby Flight app

kevinelliott:
The following adds the ability to show a nearby flight (sometimes overhead) of your location using data from a private Airframes API (https://airframes.io) made specifically for the Tidbyt. I am the founder of Airframes, so I am authorized to release this.

nearby_flight


kevinelliott:

apps/nearbyflight/nearby_flight.star
```diff
@@ -0,0 +1,196 @@
...
+AIRFRAMES_NEARBY_SORT = "distance"
+AIRFRAMES_NEARBY_SORT_ORDER = "asc"
+ENCRYPTED_API_KEY = ""
+
+def main(config):
+    api_key = secret.decrypt(ENCRYPTED_API_KEY) or config.get("api_key")
+    location = json.decode(config.get("location"))
+    print(location)
+
+    flights = get_nearby_flights(api_key, location["lat"], location["lng"], AIRFRAMES_NEARBY_RADIUS)
```
matslina, line 31: We'll need these coordinates truncated before submitting to the API. Even though I'm sure the API provider is most trustworthy, we try to guarantee that exact user locations aren't shared unless it's absolutely necessary. For this use case, I think truncating to 3 decimals would be good enough? That would provide resolution of around 100m.

apps/nearbyflight/nearby_flight.star
```diff
@@ -0,0 +1,196 @@
...
+
+# AIRFRAMES_NEARBY_URL = "http://localhost:3001/v1/tidbyt/nearby/flights"
+AIRFRAMES_NEARBY_RADIUS = "5"
+AIRFRAMES_NEARBY_LIMIT = "5"
+AIRFRAMES_NEARBY_SORT = "distance"
+AIRFRAMES_NEARBY_SORT_ORDER = "asc"
+ENCRYPTED_API_KEY = ""
+
+def main(config):
+    api_key = secret.decrypt(ENCRYPTED_API_KEY) or config.get("api_key")
```
matslina, line 27: Unlike in pixlet, `secret.decrypt()` failure will actually result in a hard error when running in our backend environment. Since `ENCRYPTED_API_KEY` is blank, this will always fail.

apps/nearbyflight/nearby_flight.star
```diff
@@ -0,0 +1,196 @@
...
+                            ],
+                        ),
+                    ],
+                ),
+            ],
+        ),
+    )
+
+def qr_code_for_flight(flight):
+    url = "https://app.airframes.io/flights/%s" % int(flight["id"])
```
matslina, line 142: Probably want to add some caching around this call too? Just to keep request rates down. Your call, since you're running the API. =)

apps/nearbyflight/nearby_flight.star
```diff
@@ -0,0 +1,196 @@
...
+    return nearby
+
+def get_schema():
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Text(
+                id = "api_key",
+                name = "Airframes API Key",
+                desc = "Your Airframes API key. You can find this in your Airframes account settings.",
```
matslina, line 186: Note that these field description are currently not visible to users in the Tidbyt iOS/Android app. You'll need to add the information to the app description in the manifest file if you want people to see it.

apps/nearbyflight/nearby_flight.star
```diff
@@ -0,0 +1,196 @@
...
+ENCRYPTED_API_KEY = ""
+
+def main(config):
+    api_key = secret.decrypt(ENCRYPTED_API_KEY) or config.get("api_key")
+    location = json.decode(config.get("location"))
+    print(location)
+
+    flights = get_nearby_flights(api_key, location["lat"], location["lng"], AIRFRAMES_NEARBY_RADIUS)
+    if flights == None or len(flights) == 0:
+        fail("No nearby flights found")
```
matslina, line 33: You may want to `return []` here instead. That will send a "blank screen" to the user, which results in the app not being displayed in the rotation.

apps/nearbyflight/nearby_flight.star
```diff
@@ -0,0 +1,196 @@
...
+AIRFRAMES_NEARBY_URL = "https://api.airframes.io/v1/tidbyt/nearby/flights"
+
+# AIRFRAMES_NEARBY_URL = "http://localhost:3001/v1/tidbyt/nearby/flights"
+AIRFRAMES_NEARBY_RADIUS = "5"
+AIRFRAMES_NEARBY_LIMIT = "5"
+AIRFRAMES_NEARBY_SORT = "distance"
+AIRFRAMES_NEARBY_SORT_ORDER = "asc"
+ENCRYPTED_API_KEY = ""
+
+def main(config):
```
matslina, line 26: Would be great if you could add an illustrative default graphic for the case when no config has been provided. That will allow us to display a preview of the app so that users get a sense of whether or not this app (and airframes.io) is something for them.

matslina (changes requested):
Thanks for the PR. Looks very cool!

A few comments inline that we should probably address before merging.

kevinelliott:
Thank you @matslina for the comments and review. I will look at the feedback and begin to address.

apps/nearbyflight/nearby_flight.star
kevinelliott: The locations are already obfuscated.

apps/nearbyflight/nearby_flight.star
kevinelliott: Cached over here, but yeah, I will consider!

apps/nearbyflight/nearby_flight.star
kevinelliott: Definitely, makes sense.

apps/nearbyflight/nearby_flight.star
kevinelliott: Good idea, thanks for the thoughts here. Definitely think a "No nearby flight" or similar message is best.

apps/nearbyflight/nearby_flight.star
kevinelliott: Oh interesting. Let me think about this.

apps/nearbyflight/nearby_flight.star
matslina: Could you elaborate a bit on this? I don't see any sign of it in the code here, which is where it would need to happen.

PR #1205, opened 2023-03-10
American Cinematheque Movie Showtimes

plattthompson:
# Overview

This app displays showtimes for your local American Cinematheque theater. As the showtimes approach, they'll change in color until the time passes and they're grayed out completely.

The app uses data from an American Cinematheque API, and the caching is set up such that calls for new data will happen relatively infrequently.

## Demo
AC app demo 1
AC app demo 2
AC app demo 3
tidbyt-ac-error


plattthompson:

plattthompson:
There's currently a linting error coming up from both `pixlet check` and `pixlet format`. I've tried everything to resolve it, but perhaps you could give me some insight into what's going on.
Screen Shot 2023-03-10 at 12 19 37 AM

rs7q5:
Did you use `pixlet create` You should have a `manifest.yaml` file too. Not sure if that's why it's mad, but you'll def need one of those.

plattthompson:
> Did you use `pixlet create` You should have a `manifest.yaml` file too. Not sure if that's why it's mad, but you'll def need one of those.

I did, but it looks like Pixlet went through a major change here, so I'll try updating it and see what happens: https://github.com/tidbyt/community/pull/1011

plattthompson:
@betterengineering @matslina @rs7q5 I changed the indentation from tabs to spaces and it appears that the linter is working properly now. Is this a known issue?

plattthompson:
Created an issue [here](https://github.com/tidbyt/community/issues/1207) and a PR [here](https://github.com/tidbyt/community/pull/1208) to resolve the tab indentation error.

Issue #1207, opened 2023-03-12
Linter gives nonspecific error when indenting with tabs


plattthompson:
Currently the `pixlet check` command will give the following error if a `.star` file is indented with tabs:

<img width="719" alt="Screen Shot 2023-03-10 at 12 19 37 AM" src="https://user-images.githubusercontent.com/43945336/224527365-1bb6ffc9-cd21-4fbe-8b12-183b30937840.png">

Upon changing the indentation to spaces, the linter works as expected.

PR #1208, opened 2023-03-12
Add .editorconfig file to clarify indentation requirements

plattthompson:
I’ve added an `.editorconfig` file to help future developers avoid [this issue](https://github.com/tidbyt/community/issues/1207). It isn’t a panacea - `.editorconfig` doesn’t work out of the box with VS Code and some other editors, and [this extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) is needed to apply the rules. However, I think it's a good first step.

Starlark itself specifies [some editor config settings](https://github.com/bazelbuild/starlark/blob/master/.editorconfig).

Additionally, perhaps this limitation could be mentioned on the FAQ page [here](https://tidbyt.dev/docs/publish/FAQ#why-is-the-linter-so-opinionated).

Long term, I would heavily recommend adding support for tabs for accessibility reasons and general DX (yes, I’m extremely biased but we don’t need to continue the indentation holy wars here).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+[*.{star}]
+indent_style = space
+indent_size = 4
+insert_final_newline = false


rs7q5:
It's weird that `pixlet format` or something doesn't fix this for you or tell you it's a tab error already.  I use `tab` to indent all the time and have not run into this issue. I do use VS code though.

betterengineering:
Thanks for the change! Happy to accept it. One thing to note though is `pixlet format` is the definitive solution to formatting. If something in the editorconfig and pixlet disagree, pixlet will have the "correct" answer.

PR #1222, opened 2023-03-17
New App: Houston Metro

DanielJRutledge:
Hello all! Adding a new transit app in the form of tracking 1-4 arrivals at Houston Metro bus and train stations! I wasn't able to do the pixlet create and check steps, but I copied other repos to have the same file structure.

Cribbed from https://github.com/tidbyt/community/blob/main/apps/nycbus/nyc_bus.star heavily.

Peek 2023-03-16 21-58


DanielJRutledge:

apps/houstonmetro/houstonmetro.star
```diff
@@ -0,0 +1,228 @@
...
+        ),
+    )
+
+def time_string(full_string):
+    time_index = full_string.find("T")
+    return full_string[time_index + 1:len(full_string) - 4]
+
+def get_stations(location):
+    loc = json.decode(location)
+    coordinates = str(loc["lat"]) + "|" + str(loc["lng"])
```
matslina, line 196: Would be great to have the coordinates truncated to avoid sharing user's exact location with the API provider. Based on the use case, I would guess 3 decimals for ~100m accuracy is sufficient here?

Check out this post for some additional info on how the number of decimals affect accuracy:
https://gis.stackexchange.com/questions/8650/measuring-accuracy-of-latitude-and-longitude

matslina (changes requested):
Thanks for the PR. Looks great! Just this one comment on location truncation and we should be ready to merge.

apps/houstonmetro/houstonmetro.star
DanielJRutledge: Done!

matslina (approved):

PR #1224, opened 2023-03-17
New App: TempestWx

epifinygirl:
New app showing more data from your Tempest Weather station. This app shows the current temperature, current wind chill temperature (feels like temperature), current humidity, inches of rain in current day, pressure, and wind. You will need to log into your Tempest account via API. This new app used the original Tempest app as the foundation and added more data points in the display.
Screen Shot 2023-03-17 at 7 19 20 PM


epifinygirl:
@rohansingh @betterengineering @matslina Hi, I saw this PR was closed by the stalebot after no activity. Could this PR be reopened and then reviewed? Thanks!

rohansingh:
@epifinygirl Looks good and happy to merge this. But what would you think about just replacing the existing app instead of making this a new app?

epifinygirl:
@rohansingh that could work but not sure if folks that currently use it would want the added data points but I’m for it if you agree.

epifinygirl:
@rohansingh This may be a silly question but can I do that via this existing PR (and just update the summary/name) or do I have to create a new PR that merges it to the existing Tempest folder?

rohansingh:
@epifinygirl You should be able to just add a commit to this same PR. Alternatively, I'm happy to amend this PR from my end as well. 

Let me know what you'd prefer.

epifinygirl:
@rohansingh If you can do it on your end, then please do! :) Thanks again!

epifinygirl:
@rohansingh The screen blank on my app and being skipped on the Tidbyt rotation. Is the same occurring on yours?

jvivona:
I can see the new version on mine & I didn't make any settings changes & it's working just fine.

rohansingh:
I am seeing a large number of errors on our server that look like this:

```
2023/04/25 15:27:19 running app tempest: running: Traceback (most recent call last):
  tempest:61:36: in main
Error: key "feels_like" not in dict
```

Perhaps `feels_like` is not always available in the response from Tempest for some reason?

rohansingh:
OK I think I see the issue. Looks like we're looking for "feels_like" units here:
https://github.com/tidbyt/community/blob/a29d47b5b4c92b94653820ceba56e7af10312db2/apps/tempest/tempest.star#L61

But I don't think "feels_like" units are a thing. Looking at the docs:
https://weatherflow.github.io/Tempest/api/swagger/#!/observations/getStationObservation

I'm guessing the units for "feels like" should just be the same `units_temp`.

jvivona:
there's no input parameter called feels_like  for getbetterforcecast

image

all the other ones are there.   I don't believe it's needed to pass it in.

rohansingh:
OK, I removed it and deployed.

@jvivona  @epifinygirl Can you check if this is working for you now? My Tempest weather station is down due to Internet issues so I can't test it at the moment.

jvivona:
yeah - I  see the new graphics now   2 lines with temps + humidity & rain.    awesome.
image

rohansingh:
Great. Not seeing any more errors or the server either.

Thanks all!

epifinygirl:
Yes I’m seeing it now, thanks for updating! 

2 quick questions - 1) you said you removed the feels like so is the units temp just showing up twice now? If so I may remove it so it’s not deceiving to users. 2) I may want to update the pressure so it shows as inHg (multiply it by 0.0295)

twx23:
This is working for me now too. Quick question, is there any chance you could add an option for dewpoint ? Either to toggle with the feels like value or humidity value ? It’s certainly more useful than relative humidity for measuring actual humidity.

rohansingh:
It's just the units request param that was removed since it's not needed. That param says what unit (C/F) to use,  but "feels like" always uses the same temperature units as the main observations.

So we're still getting back the "feels like" data as reported by the server and displaying it.

rohansingh:
> This is working for me now too. Quick question, is there any chance you could add an option for dewpoint ? Either to toggle with the feels like value or humidity value ? It’s certainly more useful than relative humidity for measuring actual humidity.

I don't have the bandwidth for this but if anyone else wants to add it, tag me and I'm happy to review the PR.

twx23:
> > This is working for me now too. Quick question, is there any chance you could add an option for dewpoint ? Either to toggle with the feels like value or humidity value ? It’s certainly more useful than relative humidity for measuring actual humidity.
> 
> I don't have the bandwidth for this but if anyone else wants to add it, tag me and I'm happy to review the PR.

Really appreciate it ! I'm actually a meteorologist and it does a much better job of measuring the actual humidity levels in the air, regardless of Temperature levels (RH is somewhat of a function of temperature). I'm sure there are many others who would appreciate this data point.

TY all!

codylerum:
Seeing some incorrect values here #1383 after this change

epifinygirl:
It’s showing up as MB units instead of inHG (conversion is taking the MB and multiplying by 0.0295). I’m hoping to update it quickly when I get a minute.

connorkeef:
@epifinygirl Excellent update!!  Love it so far. Is it possible to add decimals to the rainfall?  We have received nearly .4” here at my location and it still shows “0 in.”

epifinygirl:
PR created for updates [https://github.com/tidbyt/community/pull/1389] to fix rain decimal and add schema option for dew point/feels like.

PR #1243, opened 2023-03-23
mtatraintime: Fix timezone issue

Rai:
I noticed after the app was deployed that the timezone was UTC. This PR addresses that.


Rai:
recheck

PR #1244, opened 2023-03-23
mtatraintime: Fix timezone

Rai:
This fixes a timezone issue I noticed after the app was merged earlier.

### DIFF
--- a/apps/mtatraintime/mtatraintime.star
+++ b/apps/mtatraintime/mtatraintime.star
@@ -282,11 +282,26 @@ LINE_COLORS = {
 }
 
 API_ENDPOINT = "https://backend-unified.mylirr.org"
+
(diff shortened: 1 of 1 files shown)


drudge:
Hi @Rai! Rather than hardcoding the timezone to EST, you can use the special $location config value to pull the timezone for the Tidbyt device itself. You can add the default timezone so previews render in EST too.

If you look at any of the clock examples in this repo you can see how it's done:

https://github.com/tidbyt/community/blob/38ff6d009e80deedb08e4583391ae94b8d23cca4/apps/datetimeclock/date_time_clock.star#L13-L31

Rai:
Thanks for the detailed response, I'll take a look. However, i don't think anyone outside of the EST timezone will be riding on metro north or long island rail road.

betterengineering (approved):

PR #1246, opened 2023-03-23
Use new colour picker

dinosaursrarr:
It's much nicer than only offering 12 colours in a drop down.

Question: is it OK to change the type of existing fields? Worried that you have values saved in your DB that won't be compatible with the updated schema. I _think_ it should be OK here if the old values were a hex string, and the new values are too. But would appreciate confirmation.


betterengineering (approved):
It _should_ be a drop in replacement for strings. This looks great! Thanks @dinosaursrarr!

dinosaursrarr:
Wow, that was fast! Cheers!

PR #1251, opened 2023-03-24
New App: Todoist Tasks

TheSchultz:
todoisttasks

Shows up to 3 tasks on your To-Do list, sorted by priority. Use the filter option to further sort your tasks by using parameters such as 'today', tomorrow', 'overdue' etc. 

Let me know if there are any issues. That was all done with ChatGPT4 by modifying existing code. I have no idea how any of this actually works.


TheSchultz:

TheSchultz:
Cheers @rohansingh The process behind it isn't exactly what I'd call streamlined yet. There were many prompts used and in total all of this was done over 5-6 long conversations back and forth. But it's definitely getting there, it's a great sparring partner. Here's an example of what one of those conversations would look like.  https://shareg.pt/7HT31Iu (Don't mind the user token, was a temporary one I generated for testing)

apps/todoisttasks/todoisttasks.star
```diff
@@ -0,0 +1,212 @@
...
+def main(config):
+    token = config.get("auth") or config.get("dev_api_key")
+    children = []
+    circle_children = []  # Initialize circle_children here
+
+    if token:
+        filter = config.get("filter") or DEFAULT_FILTER
+
+        cache_key = "%s/%s" % (token, filter)
+        content = cache.get(cache_key)
```
matslina, line 34: This cache key is never written to.

matslina:
Curious if this will work, so I'll go ahead and merge. At least have to fix that cache problem before publishing.

PR #1255, opened 2023-03-26
New App: Weather Dashboard for displaying weather info from an Ambient Weather station

ckelley-1992:
This app pulls from an Ambient Weather station to display temperature, wind, and precipitation for both rain event and weekly totals.


ckelley-1992:

ckelley-1992:
recheck

ckelley-1992:
weather_dashboard

apps/weatherdashboard/manifest.yaml
```diff
@@ -0,0 +1,8 @@
+---
+id: weather-dashboard
+name: Weather Dashboard
```
matslina, line 3: Would be great if the metadata could be updated to reflect that it's specifically weather stations from Ambient Weather that's supported.

apps/weatherdashboard/weather_dashboard.star
```diff
@@ -0,0 +1,215 @@
...
+                icon = "brush",
+                default = "#1167B1",
+            ),
+        ],
+    )
+
+# pragma mark Helper Methods
+
+def get_weather_data(config):
+    cached_data = cache.get("cached_data")
```
matslina, line 79: Cache is scoped to the app, but shared among users. This cache key will need to incorporate something unique to the user, such as their API and application keys.

ckelley-1992: Ah I didn't realize that, just updated it to use the api key to disambiguate

apps/weatherdashboard/manifest.yaml
ckelley-1992: I updated the description to reference Ambient Weather specifically. I'm running into character limits on the name but will be happy to change it if you have anything specific

matslina (approved):

PR #1258, opened 2023-03-27
New App: Vercel dashboard

croossin:
This gives ability for developers to see the latest deployment information from their Vercel dashboard.

It includes:
- Commit message
- Project name
- Author name
- Success/Failure
- Time it was created

### DIFF
--- a/apps/binaryclock/binary_clock.star
+++ b/apps/binaryclock/binary_clock.star
@@ -25,26 +25,8 @@ MAX_VALUE = 2048
 DEFAULT_COLORS = {
     "Black": "#000",
     "White 100%": "#fff",
-    "White 50%": "#777",
     "White 20%": "#222",
     "Red 100%": "#f00",
-    "Red 50%": "#700",
-    "Red 20%": "#200",
-    "Green 100%": "#0f0",
-    "Green 50%": "#070",
-    "Green 20%": "#020",
-    "Blue 100%": "#00f",
-    "Blue 50%": "#007",
-    "Blue 20%": "#002",
-    "Yellow 100%": "#ff0",
-    "Yellow 50%": "#770",
-    "Yellow 20%": "#220",
-    "Cyan 100%": "#0ff",
-    "Cyan 50%": "#077",
-    "Cyan 20%": "#022",
-    "Magenta 100%": "#f0f",
-    "Magenta 50%": "#707",
-    "Magenta 20%": "#202",
 }
 
 DEFAULT_BARWIDTH = {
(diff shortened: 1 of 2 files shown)


apps/verceldashboard/vercel_dashboard.star
```diff
@@ -0,0 +1,168 @@
...
+
+def main(config):
+    apikey = config.str(CONFIG_API_KEY)
+    showIn24hr = config.bool(CONFIG_CLOCK_FORMAT, True)
+
+    # Figure out how to show if no api key
+    if apikey == None:
+        fail("No API Key provided")
+
+    cached_data = cache.get(CACHE_KEY)
```
matslina, line 32: Cache is scoped to your app, but not to individual users. This record will hold whatever happened to get written to cache last time. You likely need to incorporate something like the API key into the cache key.

matslina (changes requested):

apps/verceldashboard/vercel_dashboard.star
croossin: Oh very interesting, I did not realize that when reading the docs - must have overlooked. Will fix that.

apps/verceldashboard/vercel_dashboard.star
```diff
@@ -0,0 +1,168 @@
...
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Text(
+                id = CONFIG_API_KEY,
+                name = "API Key",
+                desc = "Your Vercel API key generated via the Vercel dashboard",
+                icon = "key",
+            ),
+            schema.Dropdown(
```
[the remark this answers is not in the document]
croossin: Also was not aware of `$tz`. I guess I could use that as default, but I do like the idea of having the option to change as some developers may enjoy seeing it in another timezone - UTC as an example.

apps/verceldashboard/vercel_dashboard.star
```diff
@@ -0,0 +1,168 @@
...
+CONFIG_CLOCK_FORMAT = "time-format"
+CONFIG_TIMEZONE = "time-zone"
+
+def main(config):
+    apikey = config.str(CONFIG_API_KEY)
+    showIn24hr = config.bool(CONFIG_CLOCK_FORMAT, True)
+
+    # Figure out how to show if no api key
+    if apikey == None:
+        fail("No API Key provided")
```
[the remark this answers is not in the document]
croossin: Certainly. I think my comment to myself even suggests that but must have totally slipped my mind. Will add that as well.

apps/verceldashboard/vercel_dashboard.star
croossin: @matslina Do we need to encode/decode the API key if we are going to use that as part of the cache key? Or is the security in place that the cache key is not exposed anywhere?

matslina: The cache key will not be visible to other apps, so using it directly should be secure. That said, there's nothing wrong with computing a hash of it and using that instead.

croossin: Makes sense - I have went ahead and updated it with a hashed version of the api key!

croossin:
@matslina I believe I took care of all the requests - let me know if there is anything else you would recommend changing. Thanks again for the review!

croossin:
@matslina It seems that the Actions are backed up and not being queued properly? Just a heads up.

PR #1263, opened 2023-03-28
Bug Fix: FlightTracker: Taxiing Status

SamuelSagarino:
Taxiing status not displaying properly; patch to fix that.


apps/flighttracker/flighttracker.star
```diff
@@ -71,7 +71,7 @@ def main(config):
     airportCode = config.get("airportCode") or DEFAULTAIRPORT
 
     # Date utilities for the API calls. These dates are not utilized in the render.
-    now = time.now().in_location("America/New_York")
+    now = time.now().in_location("Europe/London")
```
matslina, line 74: In case this was meant to be GMT, note that London is currently in BST. You can use "Etc/UTC" here for GMT.

PR #1265, opened 2023-03-28
New App: IPL Cricket

M0ntyP:
New T20 app that shows scores for the Indian Premier League, starting April 1

Similar to my other T20 cricket apps with a couple more new details
- Shows what's happened over the last 12 balls
- Shows what the team batting second need to win and off how many balls


matslina:
Thanks for the PR!

PR #1266, opened 2023-03-28
New App: Weightgurus

grantmd:
Displays your most recent weigh-in from a [Weight Gurus](https://weightgurus.com/) smart scale, along with a configurable trend plot.

Weight Gurus does not have an official OAuth api, so this uses reverse-engineered calls to the same API their mobile app uses.

Based on/steals from the existing FitBit applet to do the plotting and other layout.


grantmd:

grantmd:
recheck

apps/weightgurus/weight_gurus.star
```diff
@@ -0,0 +1,489 @@
...
+
+    #setup our variables for holding the data we get back and display
+    operations = None
+
+    if not email or not password:
+        print("No email and password set, using example data")
+        operations = json.decode(EXAMPLE_DATA)
+    else:
+        #get an access token
+        access_token = cache.get("access_token")
```
matslina, line 49: Cache is scoped to the app, but not to individual users. This will use the same cache key for all users, likely resulting in some users data being shared with all others.

matslina (changes requested):

apps/weightgurus/weight_gurus.star
grantmd: Oh, I thought the cache was per-device. So keying this off of email address is probably a good idea?

matslina (approved):

PR #1267, opened 2023-03-28
feat (whosthatpokemon): Complete list of pokemon names w/ spaces and Pokemon names w/ hypens

### LINKED ISSUE #1268: [Bug] WhosThatPokemon - Some pokemon names with spaces don't appear correctly
The hardcoded list of pokemon names with spaces and names with hyphens haven't been completely updated with Violet/Scarlet data, so some pokemon names appear incorrectly on the app because the space or hyphen isn't being accounted for

dragid10:
I noticed that I got "Flutter Mane" on the app, but the name only appeared as "Flutter". This should fix that for any other pokemon that appear that way.

Fixes #1268 

(ps: This app has made my work day less boring)


matslina:
@N-Brooks

N-Brooks:
Porygon-Z has a hyphen?! I have it in the spaces one because I definitely thought it had spaces. It checks for hyphens first so it'll still work with it in both, but might be worth it to remove it from spaces just for posterity. I'm betting `iron valiant` should be `iron-valiant` too. Good catches though, I added the new ones without a thought for all the paradoxes with spaces in their names.

dragid10:
> Porygon-Z has a hyphen?! I have it in the spaces one because I definitely thought it had spaces. It checks for hyphens first so it'll still work with it in both, but might be worth it to remove it from spaces just for posterity. I'm betting `iron valiant` should be `iron-valiant` too. Good catches though, I added the new ones without a thought for all the paradoxes with spaces in their names.

My mind selectively ignored the fact that Porygon-Z was already in the spaces category 😅
And the paradox pokemon definitely have some odd names. I honestly forgot they were pokemon until I saw Flutter Mane pop up on the tidbyt (I guessed incorrectly too)

PR #1277, opened 2023-04-01
New App - Rochester Bindicator

nolan-lynch:
Use address located in Rochester, NY to query the City website and then display what trash bins to put at the curb for collection.


nolan-lynch:

apps/roc_bindicator/roc_bindicator.star
```diff
@@ -0,0 +1,201 @@
...
+    Title_Texts = config.get("User_title_text", "Bindicator")
+    Bar_Color = config.get("bar_color", "#de00cb")
+    Positive_Color = config.get("Positive_Color", "#2f7000")
+    Negative_Color = config.get("negative_color", "#700000")
+
+#Convert config Location to Float from String
+    LatAsString = loc["lat"]
+    LatAsFloat = str(LatAsString)
+    LongAsString = loc["lng"]
+    LongAsFloat = str(LongAsString)
```
matslina, line 37: How many decimals are needed in the lat-lng pair for the app to function properly? We try to make sure apps truncate them as far as possible to reduce the risk of unnecessarily sharing/leaking user's exact location with API providers.

apps/roc_bindicator/roc_bindicator.star
```diff
@@ -0,0 +1,201 @@
...
+
+#Set conversion URL. The Location needs to change from reference type 4326 to 2262 (This is what the City GIS system needs the Location data formatted in.)
+    Conversion_URL = "https://epsg.io/trans?x="+LongAsFloat+"&y="+LatAsFloat+"&s_srs=4326&t_srs=2262"
+
+#Get Contents of Conversion URL and if failure, tell us what the code is. Also set lat & long variables for use in next step
+    GIS = http.get(Conversion_URL)
```
matslina, line 43: It would be good to add some caching arouns this (and other) HTTP calls so we keep request rates at a reasonable level.

matslina (changes requested):

apps/roc_bindicator/roc_bindicator.star
nolan-lynch: Due to pickup being based on your house address, the Lat/Long needs to be pretty specific. There are some neighborhoods where the left side of the street is a different date than the right side. So we need to pinpoint exactly where to query data for.

apps/roc_bindicator/roc_bindicator.star
nolan-lynch: I don't know if I can add caching. Based on the documentation, isn't caching for the entire app installation, not a per-user cache? If we cached the location, it would be wrong for people not located at the same address.

apps/roc_bindicator/roc_bindicator.star
matslina: Got it, thanks for the clarification!

apps/roc_bindicator/roc_bindicator.star
matslina: Correct, the cache is scoped to your app, but shared among users. You can construct your cache key so that it contains the location users in the call. That means only a user in the very same location will get the same data, which is ok.

nolan-lynch: Great. Thank you for the clarification

nolan-lynch:
Hey @matslina 
I added caching as requested and I think I did the commit correctly to send it back for review. But just letting you know because this is my first time with GIT, and still not sure if I am doing the right workflow.
Thanks

matslina:
Hey @nolan-lynch. There are some issues with the caching code here. Since the cache key "Identifier" is used across the board, the data in cache will only be useable for one location at a time. If you incorporate lat/lng into the cache key, then that should fix it, and also let you simplify the code a little bit.

Another issue is that `pixlet format` has to be run on the code. When you run that locally, it'll possibly choke on a couple of lines where there are tabs used instead of spaces for indentation. If so, those might need manual editing.

nolan-lynch:
@matslina I fixed the issue with formatting and tab characters. 
For the cache, I think it's set up right. The cache is looking for the Identifier variable, and it's value is Lattitue and Longitude provided in config. Is that the correct way to do a cache lookup for this instance type?

matslina:
Hey @nolan-lynch.

There are some lint issues failing the build right now. First was order of package loads, which I fixed. The second is warnings about potentially unused variables:

```
apps/roc_bindicator/roc_bindicator.star:162: uninitialized: Variable "TrashColor" may not have been initialized. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#uninitialized)
apps/roc_bindicator/roc_bindicator.star:166: uninitialized: Variable "isRecycle" may not have been initialized. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#uninitialized)
apps/roc_bindicator/roc_bindicator.star:166: uninitialized: Variable "RecycleColor" may not have been initialized. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#uninitialized)
apps/roc_bindicator/roc_bindicator.star:176: uninitialized: Variable "PrePickupText" may not have been initialized. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#uninitialized)
apps/roc_bindicator/roc_bindicator.star:176: uninitialized: Variable "PickupDayName" may not have been initialized. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#uninitialized)
```

I can try to fix these too, but I'm not familiar with the logic and the API, so it might be easier for you to do so. Just making sure they have sane default prior to that sequence of if-statements should suffice.

For the cache issue mentioned previously, if you can just get the PR to build cleanly and pass tests/lint, then I'd be happy to take a stab at fixing up the caching for you.

dinosaursrarr:
I just want to say that "bindicator" is a fantastic name

noahpodgurski:
Haha I love the idea of having a trash app for each town across the US

PR #1282, opened 2023-04-03
Fix Reading Challenge

theUnBurn:
I and another member of the community has found that this app does not work properly; it does not update with a new user_challenge_id.

I suspect that the secret.decrypt sets a value to CHALLENGE_ID, and because that's the case, it never checks the config for what the user types in.

My fix removes the decryption, and instead uses a default value of an Emily May, one of the most followed community members on Goodreads (list here: https://www.goodreads.com/user_following/most_followed?country=GB&duration=a). This way, an effectively anonymous user can be used as a default value while also allowing users to input their own challenge ID.


theUnBurn:
cc @panderson54 making a change to your project. It's a great idea; noticed a bug described above. Hope this solution is acceptable for you!

PR #1284, opened 2023-04-03
Initial commit of Yahoo Fantasy Football app

jweier:
This PR is for a new app for Yahoo Fantasy Football. The app has the option to display the score from the current fantasy matchup or scroll through the standings. This app depends on OAuth2 authentication with the Yahoo API. If you create a new Yahoo App to use then simply update the YAHOO_CLIENT_ID and YAHOO_CLIENT_SECRET values. I believe I've set TIDBYT_OAUTH_CALLBACK_URL to the correct URL but please correct if needed.

I have requested access to the beta and would love to be able to beta test this before it goes live due to the limited testing I could perform relating to OAuth and Secrets. 

Thanks! Really excited to get this one going and then I'm planning on a similar one for Yahoo Fantasy Baseball.


jweier:
Just some more info if you create a new OAuth app...You can create one here: https://developer.yahoo.com/apps/. 

Client Type = Confidential Client
API Permissions = Fantasy Sports (Read)

matslina:
Sorry for the long review time. This should be avaialble for beta users in a minute or so. I'll set up our own set of credentials as well before we launch it. Hit me up on discord when you've tested this out and we'll take it from there.

PR #1305, opened 2023-04-08
XScreenSaver on the Tidbyt

gknauss:
# Description
Show 265 different XScreenSaver (https://www.jwz.org/xscreensaver/screenshots/) hacks on the Tidbyt. Scaled way, way down.

They can be turned on or off individually, and are divided into groups for easy selection ("Classics", "Colorful", "Jarring", "Mathematical", "Nerdy", "Soothing", "Weird"). The ones that render the best -- about 175 of them -- are in an initially selected Default group.

The GIFs are hosted on my server, and are re-generated every two hours.

Some examples:

noof

crumbler

bouncingcow

surfaces

skytentacles

# Copilot

copilot:all


gknauss:

apps/xscreensaver/xscreensaver.star
```diff
@@ -0,0 +1,551 @@
...
+                    align = "left",
+                ),
+            )
+
+        else:
+            # If we're randomly picking a hack...
+            if config.bool("random", True):
+                # Pick one, but make sure it's not the same as the last one we ran
+                hack_last = int(cache.get("hack") or 0)
+                hack = random.number(0, len(hacks) - 1)
```
matslina, line 334:
```suggestion
                random.seed(time.now().unix // 15)
                hack = random.number(0, len(hacks) - 1)
```

matslina: This will allow for graphics to be chosen at random, while making the choice consistent over 15s time windows. Doing so will help our backend to effectively cache the rendered graphics.

apps/xscreensaver/xscreensaver.star
```diff
@@ -0,0 +1,551 @@
...
+
+            # And save the index of the current hack for a long time
+            cache.set("hack", str(hack), ttl_seconds = 60 * 60 * 24 * 365)
+
+    # If we're running from the command line, give some feedback on what we're running
+    if (config.get("hack") or config.get("hackname")):
+        print("Hack %s: %s" % (hack, hacks[hack]))
+
+    # Pull the GIF from the remote source
+    response = http.get(
```
matslina, line 354: Would suggest some caching around this call to keep request rates down. No need to refetch the same graphic over and over again.

apps/xscreensaver/xscreensaver.star
```diff
@@ -0,0 +1,551 @@
...
+                    content = "XScreenSaver: Nothing is enabled. Add some!",
+                    align = "left",
+                ),
+            )
+
+        else:
+            # If we're randomly picking a hack...
+            if config.bool("random", True):
+                # Pick one, but make sure it's not the same as the last one we ran
+                hack_last = int(cache.get("hack") or 0)
```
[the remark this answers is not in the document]
gknauss: Oh! Duh. I guess I should have expected that.

Is there a way to store a value across invocations of the app per installation outside of `config`?  If not, I'll probably remove the serial option and leave the random selection to run the risk of picking the same hack twice in a row.

apps/xscreensaver/xscreensaver.star
gknauss: I was expecting each user to be pulling different animations, but with the 15-second random seed, this makes sense. Thanks.

gknauss:
@matslina The animation generation does a lot of work to keep them as high-quality as possible while still fitting into the 192K limit for the WebP. If the bigger ones cause problems, the upper limit can be dialed down, at the cost of frame rate, colors, and duration.

apps/xscreensaver/xscreensaver.star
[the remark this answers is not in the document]
gknauss: I went ahead and removed the serial option.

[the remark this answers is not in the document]
matslina: There unfortunately is not. =( We're discussing some changes to how cache is scoped which might enable this in the not too distant future though.

matslina (approved):

matslina:
Got it. Let's give it a try.

PR #1306, opened 2023-04-08
worldclock: fix location count handling

matslina:
# Description

# Copilot

copilot:all

### DIFF
--- a/apps/worldclock/world_clock.star
+++ b/apps/worldclock/world_clock.star
@@ -47,12 +47,7 @@ def main(config):
         ]
 
     location_count = int(config.get("location_count") or 3)
-
-    if location_count < 4 and len(locations) > 3:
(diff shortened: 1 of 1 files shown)


apps/worldclock/world_clock.star
```diff
@@ -47,12 +47,7 @@ def main(config):
         ]
 
     location_count = int(config.get("location_count") or 3)
-
-    if location_count < 4 and len(locations) > 3:
-        locations.remove(locations[3])
```
matslina, line 52: If we have two distinct locations A and B, and pass in location 1,2,3,4 = `A,B,A,A`, then this line will remove the first occurrence of location `A` (i.e. `location_1`) instead of the last one (`location_4`). Besides removing the wrong locations, this also breaks the assignment of the optional `location_x_label` fields.

PR #1337, opened 2023-04-17
indycar: move graphics to app dir

matslina:
# Description

# Copilot

### Summary
📱🌎🛒

Moved `base64Maps.md` file to a higher-level directory in the `indycar` app. This file contains base64-encoded images of different race tracks for the app.

> _`is_valid` checks_
> _refactored for clarity_
> _autumn leaves falling_

### Walkthrough
* Move `base64Maps.md` file to a higher level directory to avoid duplication ([link](https://github.com/tidbyt/community/pull/1337/files?diff=unified&w=0#diff-129b15c3633a8fb947c648c8518ff0e93eee6bff3af213a80e6e5c866dac0786))


jvivona:
@matslina we were just stashing those in case someone else needed them / wanted to make tweaks - they aren't actually used in the app.

jvivona:
if we need to remove them because of the way the build happens - not a problem.  I can do it and put in another PR.

PR #1347, opened 2023-04-19
new app

skalum:
# Description
Uses the OpenWeather API to get the weather description, temperature, and "feels like" temperature for the provided location, along with an icon. Requires an OpenWeather API key.

feels_like

# Copilot

### Summary
🌡️🌎📜

This pull request adds a new applet called Feels Like, which shows the "feels like" temperature for a given location. It includes the applet code in `feels_like.star` and the manifest file in `manifest.yaml`.

> _`Feels Like` applet code_
> _Shows weather with OpenWeather_
> _A cool winter gift_

### Walkthrough
* Create and register a new applet called Feels Like that shows the "feels like" temperature for a given location ([link](https://github.com/tidbyt/community/pull/1347/files?diff=unified&w=0#diff-69aef20eba8c04879e9f60bad41efecdf649ab137ddb5695e41ab8188a4690a4R1-R8), [link](https://github.com/tidbyt/community/pull/1347/files?diff=unified&w=0#diff-5c18284c35976a8b4eaad847dc3ade382265542a817522728d33aa86bd473a85R1-R84))


skalum:

skalum:
recheck

apps/feelslike/feels_like.star
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+def main(config):
+    apiKey = config.get("apiKey", DEFAULT_API_KEY)
+
+    location = json.decode(DEFAULT_LOCATION)
+    lat, lng = float(location["lat"]), float(location["lng"])
+
+    url = "https://api.openweathermap.org/data/3.0/onecall?lat={}&lon={}&appid={}&units=imperial".format(lat, lng, apiKey)
+    weather = http.get(url).json()
```
matslina, line 52: I think it would be good to add some caching for these http calls. Even if the user provides their own API key, it'd be unfortunate if all their request quota gets eaten up faster that it has to.

matslina (changes requested):
Thanks for the PR! Looks great. Just need a bit of caching for these http calls and we should be good to go.

apps/feelslike/feels_like.star
skalum: Agreed, I just know that weather can change very quickly... I will check how often they update their data.

Also, I didn't necessarily need this to be a public app, just made it for my wife who requested it! There's no way to create a private app, correct? So you don't need to spend time approving my prs?

matslina: We can unfortunately not host private apps right now, but that might change in the future. There's of course always the option of self-hosting and pushing graphics to your Tidbyt [via the API](https://tidbyt.dev/docs/integrate/pushing-apps).

apps/feelslike/feels_like.star
```diff
@@ -0,0 +1,118 @@
...
+                id = "location",
+                name = "Location",
+                desc = "Location for which to display weather",
+                icon = "locationDot",
+            ),
+        ],
+    )
+
+def main(config):
+    currentWeather = cache.get("currentWeather")
```
matslina, line 70: Bit of a problem here I'm afraid. The cache is shared across all "installations"of the app, so this cache record will be loaded and displayed by everyone running the app. The way around it is to include something in the key that's unique to the data stored, e.g. latitude and longitude.

matslina (changes requested):
Thanks for adding the caching logic!

Also wanted to ask if you could add a default graphic for the case where no api key has been provided? That would make it clearer to other users what to expect from the app, before they go hunting for their own API key. So basically, whenever `apiKey` is `""` or `"noKey"`, render the app with mock data in `currentWeather`.

apps/feelslike/feels_like.star
matslina: Wanted to mention that we recently added caching to the http client itself. I think that'll greatly simplify the cache handling in this app. You can read more about it [over here](https://discuss.tidbyt.com/t/http-caching-client/5129).

PR #1389, opened 2023-04-26
Update Tempest App

epifinygirl:
Updates to the Tempest.

Walkthrough:
Updated the rain inches so that it shows decimals
Updated the pressure steady icon so it's an arrow instead of a straight line
Added a drop-down schema option that allows the user to choose whether to display Feels Like temperature or Dew Point temperature.


codylerum:
Is it possible to have an option to only display the single temperature maybe as double height for those that only want a single temperature number shown?

rohansingh:
Looks like this replaces the entire app with a hello world or something like that.

jvivona:
@epifinygirl still problems with rain.   where you are getting it from - it's zero - but it rained here today,   from /better_forecast - notice the rain is 0
image

but if I use /observation/station/<station_id> - it's correct (it's in mm - but it's still correct.
image

which matches my web app: 
image

at https://tempestwx.com/station/24329/grid

jvivona:
in that /better_forecast - I think that's how much rain is forecasted to fall - not how much has fallen.  rain is done here for the day

jvivona:
also - you have hard coded inches for rainfall - can't assume that.

image

I'll see if I can look at it

epifinygirl:
@jvivona the rain accum field should be from the station’s observation and not the forecast. Was it still coming up as 0 after the update where the decimals were missing? If so, I can research and see why it would be coming up as 0. 

Good call out on the inches portion - the units should be there instead. Thanks for that catch.

jvivona:
it was still zero with the decimals from what I saw.   it's going to rain here again today - I'll keep an eye on it & check in again later.

jvivona:
you may have fixed this (but maybe didn't get pushed into the repo) - line 87 needs to be %g  instead of %d  to see the decimals.  But that also assumes the api is going to return 2 decimal places all the time in that element (which seems to be true).   
image
 this is what it looks like with %g
image
and with same test data using %d
image

No - it didn't rain 9.42 inches here yesterday :-)

epifinygirl:
You're right - I had thought I updated the %d but must have missed it. I submitted a PR for that change, along with a couple other items to update/fix [1396](https://github.com/tidbyt/community/pull/1396)

jvivona:
looks good.   I just spent an hour messing with the oauth stuff on my end because I was just missing something simple.  now that I've got it - & I see your changes just pushed in.    Don't know why my device was reporting 0 rainfall yesterday - when the web app was clearly showing it.   maybe i fat fingered the device ID or something.      Thanks for the changes..

PR #1401, opened 2023-04-30
Tetris Clock: Actually fix the movement for real this time

MarkGamed7794:
# Description
So it turns out for the last three months the pieces only have a 2 in 5 chance to move whenever it decides a movement should happen? And it only ever chose to move right or rotate clockwise?

This fixes that.


MarkGamed7794:
I did not sync the files right

PR #1406, opened 2023-04-30
Add MTA Bus widget

kevineder:
# Description
I'd like to add a widget that will folks track when the next bus is arriving at any MTA bus stop in New York City. Here is the widget I built demoing when the next B69 bus will stop outside my apartment:

mta_bus

## Configs
This widget takes two configs:
* `key`: The API key for the MTA's bus time API. Anyone can request one [here](https://register.developer.obanyc.com/).
* `stop`: This is used to identify the bus stop. Anyone can look up bus stop IDs [here](https://bustime.mta.info/m/routes/).

I chose not to bundle the API key as a secret because I'm concerned about rate limiting if multiple people use this widget. If you all feel strongly that the API key should come included, I'm open to experimenting with that.

Thanks for taking a look at this!

# Copilot

### Summary
📝🚍📦

Added a new applet `mtabus` that displays real-time bus arrival information for a given stop and route in New York City. The applet consists of two files: `manifest.yaml` for metadata and `mta_bus.star` for the main logic and graphics. The applet uses the MTA Bus Time API to fetch and parse the data.

> _`mtabus` applet_
> _Metadata and modules added_
> _Tidbyt shows the spring_

### Walkthrough
*  Add a new applet for displaying MTA bus arrival times ([link](https://github.com/tidbyt/community/pull/1406/files?diff=unified&w=0#diff-1b578a7dec1fd8e12e6c4a6634f32bf8ac159ea4a5f746e9fa030df91759b3d2R1-R8), [link](https://github.com/tidbyt/community/pull/1406/files?diff=unified&w=0#diff-a74fbfb2b0a5f100b4c68a115aa9bd471696e33091819369fe08cec5fe001eb2R1-R147))


kevineder:

kevineder:
Recheck

apps/mtabus/mta_bus.star
```diff
@@ -0,0 +1,152 @@
...
+"""
+BUS_ANIMATION_DURATION = 200
+
+def main(config):
+    if not config.get("key"):
+        return render.Root(
+            child = render.Text("No API key set."),
```
matslina, line 25: It would be great if this case could result in a default graphic with mock data. Just to give users a sense of what the app will look and feel like if they move on to signing up for their own API key.

matslina (changes requested):
Thanks for the PR! I know a few people in NYC who'll be very happy to have this!

Just the one request about adding a default graphic before we can merge and publish the app.

apps/mtabus/mta_bus.star
kevineder: Oooh I didn't realize! I will update, thanks.

kevineder: Smart. I'll update.

kevineder: Updated to handle cases where the user hasn't set the API key or stop ID, as well as handling the error cases. Let me know your thoughts on these various states.

### No API key set
mta_bus_no_key_set

### No stop ID set
mta_bus_no_stop_set

### No stop results returned
mta_bus_no_results

### API returns a non-200 response
mta_bus_api_error

kevineder: I realize you were suggesting a "default" graphic, not handling these states. To show something more-inviting to new users, do you think I should show `Next Bus ?? minutes` for the first 60 seconds, then go to the above behavior of letting them know they need to set the API key / stop ID after that?

matslina: I think this looks great! Love the bus passing by. =)

matslina (approved):

PR #1415, opened 2023-05-03
Fix Sky News app

meejle:
# Description
The API I was using has (sort of) closed so I'm now self-hosting the XML-to-JSON script.

# Copilot

### Summary
🔄🆗📰

Fixed the Sky News app by changing the XML-to-JSON converter service URL in `skynews.star`. This restores the functionality of fetching and displaying the news headlines.

> _`URL` changes_
> _XML-to-JSON converter_
> _Fall of old service_

### Walkthrough
*  Update the XML-to-JSON converter service URL to fix the Sky News RSS feed parsing ([link](https://github.com/tidbyt/community/pull/1415/files?diff=unified&w=0#diff-8c9210a64e9feb757641beeedfabab0710627095d7c95d1ad0f56e174fffc7f1L25-R25))


jvivona:
@meejle - have you thought about using the RSS feed direct instead?  that's what I am doing for the Fox/CNN/FinTimes/Guardian/Wall Street Journal apps.    I had similar problems where the APIs were getting dinged, changed, etc - but those RSS feeds were working just great & consistent (even when Guardian had their hack, their API went down - but RSS kept going).    Just my 2 cents.

meejle:
> @meejle - have you thought about using the RSS feed direct instead?

I... don't know how to do that 🙈 But I'll have a look at your apps later and see if I can figure it out, thanks.

This was only really meant as a stopgap tbh, I hate to have a dead app in the 'store' haha.

jvivona:
here's the code I use.  call the RSS feed direct via the URL   and then parse it:
`def get_cacheable_data(url, articlecount):
    key = url
    data = cache.get(key)
    articles = []

    if data == None:
        res = http.get(RSS_STUB.format(url))
        if res.status_code != 200:
            fail("request to %s failed with status code: %d - %s" % (url, res.status_code, res.body()))
        data = res.body()
        cache.set(key, data, ttl_seconds = CACHE_TTL_SECONDS)

    data_xml = xpath.loads(data)

    for i in range(1, articlecount + 1):
        title_query = "//item[{}]/title".format(str(i))
        desc_query = "//item[{}]/description".format(str(i))
        articles.append((data_xml.query(title_query), str(data_xml.query(desc_query)).replace("None", "")))

    return articles`

PR #1450, opened 2023-05-12
[Planetarium] Add option to hide the clock.

mtimkovich:
# Description
Added an option to planetarium called "static" to hide the clock and only show the planets.

cc: @dinosaursrarr 

# Copilot

### Summary
📱🌐🛒

Added a new app for displaying the current phase of the moon on the tidbyt. The app uses the `lunardate` module to calculate the moon phase and the `PIL` module to draw the moon image. The app code is in `apps/moon_phase.py` and the app config is in `apps/moon_phase.json`.

> _`is_valid` checks_
> _refactored for clarity_
> _autumn leaves falling_

### Walkthrough
*  Add a new app called `weather` that displays the current weather and forecast for a given location (weather.py, weather_test.py)


dinosaursrarr:
Looks fine to me.

My only question is whether it makes sense to make showing the clock a separate boolean option in the schema? Seems like your static mode = image_set mode but without the clock. But it seems like someone might also want an animated mode without the clock. Seems like it might be easier to accommodate with a second option rather than doubling the number of modes.

mtimkovich:
Thanks! Good suggestion

dinosaursrarr:
Looks fab 👍

PR #1480, opened 2023-05-17
strava: Convert to new HTTP cache.

betterengineering:
# Description
This commit Strava to use the new HTTP cache. In addition, it updates the CACHE TTL to instead be 36 hours to help alleviate API rate limits until we can get them increased. cc @rgkimball 

# Copilot

### Summary
🌐🧹🚀

Refactor Strava app to use `HTTP.cache` and remove unused code. This improves the app's performance, readability, and compliance with the Strava API terms.

> _`HTTP.cache` used_
> _Simpler code, fewer calls_
> _Winter of excess_

### Walkthrough
*  Remove unused and unnecessary code related to rate-limiting and debugging ([link](https://github.com/tidbyt/community/pull/1480/files?diff=unified&w=0#diff-7cc751532edf305a47e0bcc935eb429a2913f05523ddd5beec97283d56bb4ad1L26-L28), [link](https://github.com/tidbyt/community/pull/1480/files?diff=unified&w=0#diff-7cc751532edf305a47e0bcc935eb429a2913f05523ddd5beec97283d56bb4ad1L141-L145), [link](https://github.com/tidbyt/community/pull/1480/files?diff=unified&w=0#diff-7cc751532edf305a47e0bcc935eb429a2913f05523ddd5beec97283d56bb4ad1L195), [link](https://github.com/tidbyt/community/pull/1480/files?diff=unified&w=0#diff-7cc751532edf305a47e0bcc935eb429a2913f05523ddd5beec97283d56bb4ad1L214), [link](https://github.com/tidbyt/community/pull/1480/files?diff=unified&w=0#diff-7cc751532edf305a47e0bcc935eb429a2913f05523ddd5beec97283d56bb4ad1L1156-L1177))
[...]

PR #1482, opened 2023-05-17
Revert "strave: Convert to new HTTP cache. (#1480)"

betterengineering:
This reverts commit 8fd24937df78fcd6583f3e16f96a947e62f437cb.

# Description

# Copilot

### Summary

No summary available (Limit exceeded: required to process 69364 tokens, but only 50000 are allowed per call)

### Walkthrough
No walkthrough available (Limit exceeded: required to process 69364 tokens, but only 50000 are allowed per call)


betterengineering:
This didn't work with the new caching client - the issue is, we are keying the request based on the headers. It works! But any time we get a new access token, it would bust the cache, which does not work

PR #1506, opened 2023-05-24
New App: Tides & Temp

sudeepban:
# Description
Tides & Temp app for displaying predicted tides from NOAA tide stations and water temperature from NDBC buoy stations. See https://tidesandcurrents.noaa.gov/tide_predictions.html for NOAA tide station IDs. See https://www.ndbc.noaa.gov/obs.shtml for NDBC buoy station IDs. Wave graphic moves up and down based on estimated current tide calculated from the previous and next low / high tides.

tides_and_temp

# Copilot

### Summary
🌊🌡️📄

This pull request introduces a new applet for tidbyt that shows tide and water temperature data from NOAA and NDBC stations. The applet is implemented in `tides_and_temp.star` and uses render widgets and a schema object. The applet is also registered in the `manifest.yaml` file with its metadata.

> _`Tides & Temp` app_
> _Shows the sea's mood and warmth_
> _Manifest for spring_

### Walkthrough
*  Add the Tides & Temp applet, which displays predicted tides and water temperature from NOAA and NDBC stations ([link](https://github.com/tidbyt/community/pull/1506/files?diff=unified&w=0#diff-3cbc1c97c6acd593129b4b6ef0055bdae1d7e25000c5ef1831eaecd594bb7fccR1-R8), [link](https://github.com/tidbyt/community/pull/1506/files?diff=unified&w=0#diff-eea0cf481e8f71702f3c2d4583034be91ee9c98cbaa419a8e9f75bf29f1ada86R1-R202))


sudeepban:

apps/tidesandtemp/tides_and_temp.star
```diff
@@ -0,0 +1,202 @@
...
...AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==""")
```
matslina, line 102: Given the size of this block, I'm thinking it might be a good idea to scale the graphic down ahead of time?

apps/tidesandtemp/tides_and_temp.star
```diff
@@ -0,0 +1,202 @@
...
+        schema.Option(display = "China (+8)", value = "Asia/Shanghai"),
+        schema.Option(display = "Japan (+9)", value = "Asia/Tokyo"),
+        schema.Option(display = "Australia Eastern (+10)", value = "Australia/Sydney"),
+        schema.Option(display = "New Zealand (+12)", value = "Pacific/Auckland"),
+    ]
+
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Dropdown(
```
matslina, line 158: A common pattern for getting timezone data is to use `schema.Location`. That allows the user to select a specific location, and gives you the associated timezone in the location object.

apps/tidesandtemp/tides_and_temp.star
```diff
@@ -0,0 +1,202 @@
...
+    noaaTidesStationID = config.str("noaaTidesStationID", str(NOAA_TIDES_STATION_ID_DEFAULT))
+    ndbcBuoyStationID = config.str("ndbcBuoyStationID", str(NDBC_BUOY_STATION_ID_DEFAULT))
+    lowTideColor = config.str("lowTideColor", LOW_TIDE_COLOR_DEFAULT)
+    highTideColor = config.str("highTideColor", HIGH_TIDE_COLOR_DEFAULT)
+    waterTempColor = config.str("waterTempColor", WATER_TEMPERATURE_COLOR_DEFAULT)
+
+    now = time.now().in_location(TIMEZONE_GMT)
+    prevdate = (now - time.parse_duration("12h")).format(DATE_FORMAT)
+
+    resp = http.get(NOAA_TIDES_URL.format(begin_date = prevdate, station_id = noaaTidesStationID))
```
matslina, line 42: Would it be appropriate to add a `ttl_seconds` flag to this call? Not sure if they provide cache-control headers?

apps/tidesandtemp/tides_and_temp.star
matslina: Also, you automatically get the device's configured location in the magic '$tz' config variable. Not sure if that's good enough or if it's important for this to be fully configurable.

matslina (dismissed):

apps/tidesandtemp/tides_and_temp.star
sudeepban: I didn't see anything at TTL or cache-control headers at https://api.tidesandcurrents.noaa.gov/api/prod/ - let me know if you see anything relevant I could incorporate

apps/tidesandtemp/tides_and_temp.star
sudeepban: I thought about using the `schema.Location` option or the magic `$tz` config variable, but I thought it may be advantageous for the user to have more direct control over setting the timezone.

apps/tidesandtemp/tides_and_temp.star
sudeepban: Could you elaborate?  I also thought the size of the string was a little out of control but wasn't sure what I could do to improve it. It's already a small image, only 6x15, but it does have a decent number of frames to tie out the different waves so it cycles nicely.

sudeepban: I imagine one option would be to get the animated GIF properly hosted somewhere so I could reference that way, but wasn't sure if Tidbyt offers any such option.

apps/tidesandtemp/tides_and_temp.star
matslina: Got it. In that case I still think `Location` is a good option to consider. Gives you handling of e.g. daylight's savings for free, and is likely more intuitive for the user (most people don't think much about their timezone's current offset from UTC for some reason) =). But it's your call!

apps/tidesandtemp/tides_and_temp.star
matslina: When I decode that base64 block, I get a 1590x636 GIF. It looks like it's a 15x6 scaled up by a factor of 106.

sudeepban: Ohh interesting, let me take a closer look at that, thanks

apps/tidesandtemp/tides_and_temp.star
sudeepban: I think you're right, I will switch to using Location

sudeepban: Switched to using `schema.Location` now

apps/tidesandtemp/tides_and_temp.star
sudeepban: I tried rescaling to 15x6, and it actually made the animated GIF look not quite right. Rescaling to 45x18 looked fine, so it still reduces the size of the base64 block substantially.

apps/tidesandtemp/tides_and_temp.star
matslina: The simplest thing is probably for you to decide how long the data retrieved is good for. So for instance, if it's ok to use the buoy data for (say) 1 hour, you could do `http.get(NDBC_BUOY_URL.format(station_id = ndbcBuoyStationID), ttl_seconds = 3600)`. This means subsequent requests for the exact same URL will not hit the NOAA server, but will instead be served from our cache. After 3600 seconds (=1h) the cached data will get thrown out and the next request will refresh the data from the API again.

apps/tidesandtemp/tides_and_temp.star
sudeepban: Great, thanks

apps/tidesandtemp/tides_and_temp.star
sudeepban: I see, that makes sense. My preference would be for it to be up to date, with the latest water temperature reading as soon as it's made available. It seems like adding any non-zero `ttl_seconds` would introduce some potential arbitrary delay.

sudeepban:
@matslina is this change good to merge at this point?  Let me know if there's any other feedback I can address, thanks for your review!

apps/tidesandtemp/tides_and_temp.star
matslina: Totally understand wanting up to date data. That makes sense. Unfortunately we do have to cache responses to keep request volumes at a resonable level.

I took a look at the data served on the buoy url, and it includes cache-control headers requesting 600s caching of responses, so that's the minimum we should do there. In fact, that's what'll happen automatically when `ttl_seconds` is left out.

There doesn't seem to be such a header on the other endpoint, but if I understand this correctly, it only serves tide predictions. These likely do not change much in the short term, so an hour of caching shouldn't hurt I imagine?

For reference, the existing apps nooatides and nooabuoy cache responses for 4 hours and 10 minute respectively.

sudeepban: Thanks for the details.  I accepted your proposed changes, this should be fine!

sudeepban:
@matslina is there any other feedback to address or can this be merged now?

sudeepban:
> @matslina is there any other feedback to address or can this be merged now?

@betterengineering @matslina anything I can do to move this along? Really wanted to make this available via the mobile app for the beach season!

wilbah:
I would like to see / use this!

matslina:
Just got back from vacation. Sorry for the delay!

sudeepban:
> Just got back from vacation. Sorry for the delay!

Ahh no problem, thank you! Hope you had a great vacation

sudeepban:
@matslina I was curious, what is the process that makes a new app like this available via the Tidbyt mobile app?

sudeepban:
> When we merge a PR, CI takes over and releases the app. It should be available in the mobile app within minutes. It won't get categorized automatically, i.e. you won't see it under "what's new", but it should be avaialable when you search.

That's awesome, I see it now! Thank you

PR #1520, opened 2023-05-30
Update Jacket to use Openweather API

noahpodgurski:
# Description
Update Jacket to use Openweather API

# Copilot

### Summary
🧥🌤️🛠️

This pull request adds a new app called `jacket`, which tells the user whether they need a jacket or not based on the weather. The app uses the OpenWeather API to get the current temperature and conditions for a given location. The app code is in `jacket.star` and the app metadata is in `manifest.yaml`.

> _`jacket` app code_
> _fetches weather from the sky_
> _shows if you need warmth_

### Walkthrough
*  Add the jacket app, which displays whether the user needs a jacket or not based on the weather data ([link](https://github.com/tidbyt/community/pull/1520/files?diff=unified&w=0#diff-d06671f05ad7cc2a2e0e666d7cd24ea413131151e3471527049cbdfdeae86a36R1-R248), [link](https://github.com/tidbyt/community/pull/1520/files?diff=unified&w=0#diff-54215f3b495a6ce4de7cfa87c86a2abe6d6ac786aa49e7359b59a46057131a69R1-R8))


apps/jacket/jacket.star
```diff
@@ -182,60 +186,53 @@ def get_weather_data(config):
...
-        if application_key == None:
-            print("Missing application_key")
+        if location == None:
+            print("Missing location")
             return SAMPLE_STATION_RESPONSE
 
         print("Getting new weather data")
+        location = json.decode(location)
+        query = "%s?exclude=minutely,hourly,daily,alerts&lat=%s&lon=%s&appid=%s" % (OPEN_WEATHER_URL, location["lat"], location["lng"], api_key)
         res = http.get(
```
matslina, line 196: Note that the http module now has caching built in. You can pass `ttl_seconds` to `http.get()` and it'll all be handled for you!

matslina (approved):
This change will break all existing installations of the app, as these are already set up to pass in an ambient weather key, and won't pass in a location. It might be OK in this case, since the app is very new and there likely aren't many installations of it, but I wanted to mention it so you're aware before we merge.

apps/jacket/jacket.star
noahpodgurski: @matslina this is updated

PR #1525, opened 2023-06-02
New App: LA metro transit line stop times

ConnorWasHere:
# Description
Use the transitapp API to get departure times for LA metro trains at a specific stop.

la_metro_stop

# Copilot

copilot:all

### DIFF
--- a/apps/lametrostop/la_metro_stop.star
+++ b/apps/lametrostop/la_metro_stop.star
@@ -1,0 +1,144 @@
+"""
+Applet: LA Metro Stop
+Summary: LA metro times for stop
+Description: Get departure times for LA metro train stop.
+Author: connorwashere
+"""
+
+load("cache.star", "cache")
+load("encoding/json.star", "json")
+load("http.star", "http")
+load("render.star", "render")
+load("schema.star", "schema")
+load("secret.star", "secret")
+load("time.star", "time")
+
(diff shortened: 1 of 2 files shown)


ConnorWasHere:

ConnorWasHere:
recheck

apps/lametrostop/la_metro_stop.star
```diff
@@ -0,0 +1,144 @@
...
+                        render.Text(get_displayed_times(route_info[stop_name], 2), font = "5x8", offset = 1, color = "FFB033"),
+                    ],
+                ),
+            ],
+        ),
+    )
+
+def get_times(stop_id, api_key):
+    # Hits AC Transit's prediction api if there are no cache hits
+    cached = cache.get(stop_id)
```
matslina, line 95: If you haven't already, check out the http module's built in cache functionality. You can pass `ttl_seconds` straight into `http.get()` to have a request cached. Should simplify the code here a bit, and also future proof as we may start changing up the cache module a bit in the not too distant future.

matslina (approved):

apps/lametrostop/la_metro_stop.star
ConnorWasHere: I cant seem to find the documentation or code for that http cache, is the cache key just a hash of the URL, params, headers and body of the request?? I'd be happy to implement it using that cache instead!

ConnorWasHere: I've updated the code, please let me know if I used the http cache correctly!

matslina: Oh wow, yeah I think we need to update some docs. Thanks for the change!

PR #1534, opened 2023-06-04
Update/Modify: Big Clock App

rohzino:
Added amber color to the app


rohzino:

rohzino:
recheck

matslina:
@joeyhoer Ok with you?

Might be a good idea to switch the config field over `Color`, with the current options in the `palette` field?

PR #1545, opened 2023-06-07
Update http to do the caching

rs7q5:
# Description
update the github apps of mine to do the caching in the http module. Also removed the TODO comment in newapps as it is not eligible to be cached in `http.get` because call the same command twice with the same url.

@betterengineering to be aware of why the #TODO was removed in `new_apps.star`

# Copilot

copilot:all


matslina:
ty!

PR #1587, opened 2023-06-16
roomba: don't fail on missing config

matslina:
# Description

# Copilot

### Summary
🛠️🌐🧹

Fixed a bug in `roomba.star` that caused the app to crash when the server IP was not configured. The app now returns an empty list in the `config` function to display a blank screen instead of an error message.

> _No server IP_
> _`return []` before `fail`_
> _Blank screen in winter_

### Walkthrough
*  Prevent app from crashing when server IP is not configured by returning an empty list before failing ([link](https://github.com/tidbyt/community/pull/1587/files?diff=unified&w=0#diff-e30ae99e3ce331353e74da47892999add51030b44eb57b6beab2815ac34b9187L123-R128))

### DIFF
--- a/apps/roomba/roomba.star
+++ b/apps/roomba/roomba.star
@@ -126,7 +126,6 @@ def main(config):
 
     if not serverIP:
         return []
-        fail("Server IP must be configured")
 
     if type(int(serverPort)) != "int":
         fail("Server Port must be an integer")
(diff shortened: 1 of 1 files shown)


matslina:
@noahpodgurski CI now requires apps to run without error when no config is present. This change is to unbreak CI, but ideally the app would produce a default graphic in this case that can be used to generate previews, so that users can get a sense of what the app will look like once it's fully set up.

PR #1660, opened 2023-07-10
Marvel Facts: App Update

GothamGamer117:
# Description
This update fixes an issue where the app only displays "None" due to an incorrect API URL.

# Copilot

### Summary
📝🗓️🆕

Updated the docstring of the `marvel_facts.star` applet to reflect the latest changes and comply with the documentation standards.

> _`applet` docstring_
> _updated with date and notes_
> _tidbyt guidelines_

### Walkthrough
*  Update the applet docstring with the last updated date and description ([link](https://github.com/tidbyt/community/pull/1660/files?diff=unified&w=0#diff-c15c018fafe13d95e29a059577efe45a373695c54f463ab5b0329aeb44ea1048L1-R129))


matslina:
@GothamGamer117 I think the formatting issue was due to unix vs windows style line break, i.e. `\n` vs `\r\n`. I pushed a commit that gets past that particular problem. The next issue is that secrets can't be properly decrypted in CI. Bit of a pain this, but you basically need to make the app render something in this case instead of erroring out. When it's been merge it'll run in our backend environment, where the secrets can be decrypted properly.

GothamGamer117:
> @GothamGamer117 I think the formatting issue was due to unix vs windows style line break, i.e. `\n` vs `\r\n`. I pushed a commit that gets past that particular problem. The next issue is that secrets can't be properly decrypted in CI. Bit of a pain this, but you basically need to make the app render something in this case instead of erroring out. When it's been merge it'll run in our backend environment, where the secrets can be decrypted properly.

I just fixed the secrets issue, and it now displays properly locally. I seem to still have the formatting issue though (probably due to new code I added), and I am not sure how to fix that. Once that is fixed, it should all be working as expected.

PR #1661, opened 2023-07-10
New App: Octoprint

noahpodgurski:
# Description
View octoprint server status
octoprint

# Copilot

copilot:all

### DIFF
--- a/apps/octoprint/octoprint.star
+++ b/apps/octoprint/octoprint.star
@@ -1,0 +1,306 @@
+"""
+Applet: Octoprint
+Summary: View Octoprint status
+Description: Display current print's status and time remaining.
+Author: noahpodgurski
+"""
+
+load("animation.star", "animation")
+load("http.star", "http")
+load("math.star", "math")
+load("render.star", "render")
+load("schema.star", "schema")
+
+REFRESH_TIME = 60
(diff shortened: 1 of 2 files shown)


apps/octoprint/octoprint.star
```diff
@@ -0,0 +1,303 @@
...
+        )
+
+def get_schema():
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Text(
+                id = "serverIP",
+                name = "Server IP",
+                desc = "Ex: (192.168.1.123)",
```
matslina, line 281: Since apps run on Tidbyt servers, the app won't be able to reach local networks. I suspect that may render this app unusable in most cases?

noahpodgurski: This (like the roomba app), again requires you to port forward your octoprint port and use the public IP. I'll change the desc here so it's not misleading

matslina: Got it. I should also mention that these field descriptions strings unfortunately aren't exposed to users in the mobile app, so it might be good to mention it in the app description as well.

noahpodgurski: Fixed

PR #1665, opened 2023-07-11
New app: Avatars in Pixels

danielsitnik:
# Description
This app generates a random pixel art character using the [Avatars In Pixels API](https://www.avatarsinpixels.com/) and displays it on the Tidbyt using a nice animation:

ezgif-1-c30d151a4a

More details are provided in the `README.md` file.

# Copilot

copilot:all


matslina:
Thanks for the PR. If possible, it'd be great if you could use the built in cache functionality of the http module. I realize it might be tricky due to that cookie passed to the second request, but still wanted to mention it. We're hoping to minimize the use of the cache module, and possibly even remove or rescope it.

danielsitnik:
> Thanks for the PR. If possible, it'd be great if you could use the built in cache functionality of the http module. I realize it might be tricky due to that cookie passed to the second request, but still wanted to mention it. We're hoping to minimize the use of the cache module, and possibly even remove or rescope it.

Thanks. I considered using the http module cache, but I preferred the old cache module since it's not scoped by installation. I know that every Tidbyt device will display the same image, but this was intentional so the avatars in pixels API is not hammered, and there's less risk of rate limiting and IP blocking.

matslina:
Got it, good thinking. However, the http cache is actually not scoped by installation, but is instead shared among all users of the app. We may change the scope of `cache.star` in the future though, and that could become per-installation. TBD.

PR #1676, opened 2023-07-13
update: SailGP added large format standings display

jvivona:
# Description
 added scrolling, large format standings display with Flags

# Copilot

copilot:all


jvivona:
:shipit:

PR #1703, opened 2023-07-24
airthings : guard against missing pm25 value in sensor data

tavdog:
# Description
If the airtag does not report pm25 then the script crashes when trying to access.  This fix check for the presence of the data and if it's missing set it to -1 and then display n/a when rendering.

# Copilot

copilot:all


matslina:
Unrelated to this particular change, but it'd be great if someone could update the app to output an illustrative default graphic when no configuration has been provided. For previews and such.

PR #1725, opened 2023-08-08
update for new http caching

jmanske:
# Description
Updated code to use new http caching instead of old method

# Copilot

### Summary
🚀🛡️🧹

Use HTTP cache for leetcodestats app. This change updates the `leetcodestats` app to use the built-in HTTP cache feature instead of the external `cache` module, which improves the app's efficiency and code quality.

> _`HTTP cache` shines_
> _Storing and fetching stats data_
> _App is fast and smooth_

### Walkthrough
* Enable HTTP cache feature for stats data ([link](https://github.com/tidbyt/community/pull/1725/files?diff=unified&w=0#diff-c4d1f160dc1daecadb3874dbc0de58f370ed9133b65aaeba0e6256aac9b1d5d9L176-R158))

### DIFF
--- a/apps/leetcodestats/leetcodestats.star
+++ b/apps/leetcodestats/leetcodestats.star
@@ -6,6 +6,7 @@ Author: Jake Manske
 """
 
 load("encoding/base64.star", "base64")
+load("encoding/json.star", "json")
 load("http.star", "http")
 load("math.star", "math")
 load("render.star", "render")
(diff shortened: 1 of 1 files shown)


jmanske:
According to `pixlet profile`, almost all the time is spent on the http request. I can work to try to have a different GraphQL query to be faster but I wanted to update the app in its current state to use the new http caching.

PR #1739, opened 2023-08-12
Fix destination row splitting

dja852:
# Description
When the user is not using one of the display format the includes the destination, if there are multiple destinations being served by same route, they will split into multiple lines. e.g.:

K 4, 28
K 16

This corrects that by ignoring the destination name when it is not being displayed and the above would instead be rendered as K 4,16,28 which is more useful to the user.

# Copilot

### Summary
🚏🌈🔵

This pull request enhances the SF Next Muni app by adding more options for displaying bus arrival times and improving the appearance of route tags. It modifies the `apps/sfnextmuni/sf_next_muni.star` file.

> _Sing, O Muse, of the skillful coder who enhanced the app of Muni_
> _With new formats of prediction, showing the time of each arrival_
> _And adjusted the circle rendering, making the route tags clear and bright_
> _Like the stars that guide the sailors on the dark and stormy night_

### Walkthrough
*  Modify `titleKey` variable to group predictions by route and destination for different formats ([link](https://github.com/tidbyt/community/pull/1739/files?diff=unified&w=0#diff-edab8caf454dbc5b959d38e45974840dcf6e2cd3be2ecdcda9ff989a7c62eb45L393-R393))
*  Display only first character of route tag in circle to save space and avoid overlap ([link](https://github.com/tidbyt/community/pull/1739/files?diff=unified&w=0#diff-edab8caf454dbc5b959d38e45974840dcf6e2cd3be2ecdcda9ff989a7c62eb45L563-R565))

### DIFF
--- a/apps/sfnextmuni/sf_next_muni.star
+++ b/apps/sfnextmuni/sf_next_muni.star
@@ -390,7 +390,7 @@ def getPredictions(api_key, config, stop):
         seconds = predictedTimes[0] - time.now().unix
         minutes = int(seconds / 60)
 
-        titleKey = routeTag if "short" == config.get("prediction_format") else (routeTag, destTitle)
+        titleKey = (routeTag, routeTag) if config.get("prediction_format") in ("short", "medium", "two_line_four_times") else (routeTag, destTitle)
         if titleKey not in prediction_map:
             prediction_map[titleKey] = []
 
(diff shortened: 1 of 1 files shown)


matslina:
@juniormonkey Any thoughts on this one?

juniormonkey:
LGTM. You're right, splitting the lines makes no sense in this case. Thanks for the fix.

PR #1758, opened 2023-08-21
feat: add an app that shows you how much you've made today

klar-kuo:
# Description
Created a simple app that takes in paycheck and workday info and calculates how much money you've made so far that day. A nice little visualization of why you may be doing it

# Copilot

copilot:all


klar-kuo:

PR #1812, opened 2023-09-06
Hotfix for Critical Chicken app

meejle:
# Description
Had to rework it again as I'm having trouble with `marquee` getting cut off even though I used `show_full_animation = True`. Fingers crossed this will sort it. 😁

# Copilot


meejle:
I dunno, this looks _perfect_ in Pixlet, but the preview above is still ending early... 💀 Either way it's 4am so I'm just gonna leave this here and see what it looks like once it's merged. 😅

PR #1822, opened 2023-09-08
New App: AccuWeather

sudeepban:
# Description
AccuWeather forecast app for displaying AccuWeather three day forecast based on AccuWeather location key. To request an AccuWeather API key, see https://developer.accuweather.com/getting-started. To determine AccuWeather location key, search https://www.accuweather.com for a location and extract trailing number, e.g. 2191987 for https://www.accuweather.com/en/us/lavallette/08735/weather-forecast/2191987.

AccuWeather Forecast

# Copilot

### Summary
🌤️📝🛍️

Add a new app for displaying AccuWeather forecasts on the tidbyt device. The app requires an AccuWeather API key and location key, which are explained in the app description in `apps/accuweather/manifest.yaml`.

> _`AccuWeather` app_
> _Metadata for tidbyt store_
> _How to get keys? Read_

### Walkthrough
* Add the metadata for the AccuWeather app in `manifest.yaml` ([link](https://github.com/tidbyt/community/pull/1822/files?diff=unified&w=0#diff-6f60e304c0aa293c35b037dd937a08bf0f3869bd62a8e47fe67f3ac7ad095135R1-R8))


sudeepban:
I'm seeing the "Build and Test / check (pull_request)" check is failing with 401 error because the AccuWeather API key and AccuWeather location key need to be specified.  Can you advise on the right approach to making the check pass?

sudeepban:
@matslina is there a way I can specify schema params for the "Build and Test" check? It's failing because there is no default AccuWeather API key or location key in the app, both need to be supplied by the user.

sudeepban:
> The most common way of handling this is to handle the case of missing configuration by generating an illustrative default graphic. This has the added benefit of providing a preview graphic in the Tidbyt mobile app for users who haven't yet gone out and gotten an API key. Might want to overlay "SAMPLE" (check out the stack widget) in this case to make it especially clear to the user that they shouldn't expect the sample data in their particular location. =)

Thanks for the tip, I have updated the PR to display sample data with a red "SAMPLE" overlay in the event that the user-specified AccuWeather API key or location key are unavailable.

sudeepban:
@matslina @betterengineering anything else needed from me before this can be reviewed?

PR #1861, opened 2023-09-17
Update Tessie colors

inderpal:
Update Tessie colors for low Tesla battery states


inderpal:
@matslina @betterengineering - can you review this update? thanks!

PR #1945, opened 2023-10-14
New App: Lions Gate Bridge Status

aidenmitchell:
# Description
Created a new app to display the traffic and lane status of Vancouver's Lions Gate Bridge.

# Copilot

### Summary
📥🔑🚗

This change adds a new app `lionsbrgstatus` that displays the status of the Lions Gate Bridge in Vancouver. It imports the necessary modules and defines the constants for the app logic, including the images and the custom API for the counterflow arrows.

> _`base64`, `http`_
> _modules for app logic_
> _winter road safety_

### Walkthrough
* Create a new app called `lionsbrgstatus` that displays the status of the Lions Gate Bridge counterflow system in Vancouver, Canada (F0, F1, F2)
* Load the base64, http, render, and secret modules from the Starlark standard library ([link](https://github.com/tidbyt/community/pull/1945/files?diff=unified&w=0#diff-6ef5a5ffbc9c4cab5206f17f5268f8f701f9f77e4ad0b813de7b3ea70e7f306cR1-R86))


aidenmitchell:
Either the CI checks can't read secrets (like the local CLI) or I did something wrong 😛 
Please let me know if I need to fix something!

apps/lionsbrgstatus/lions_gate_status.star
```diff
@@ -0,0 +1,86 @@
...
+    sign_text = decoded_text.replace("[pt30o2]  ", "")
+
+    # Build lane icons list iteratively
+    lane_icons = []
+    for direction in arrow_directions:
+        lane_icons.append(render.Image(src = get_image_src_for_direction(direction), width = 10))
+        lane_icons.append(render.Box(width = 11))
+
+    # Remove the last gap after the last icon
+    lane_icons.pop()
```
jmanske, line 61: If `arrow_directions` array is empty (which it can be if you get anything other than 200 response code from the aws API call), then lane_icons will be empty and `.pop` will throw `index -1 out of range: empty list` exception. Might be worth making this more robust or failing in a more expected way if the API call to get the lane directions fails.

PR #1979, opened 2023-11-01
Fix links and station key in citibike-by-mk app

glenrobertson:
# Description
Lyft must have deprecated the citibikenyc GBFS feed, as well as the legacy_id key for station id.
This change updates the URL and key so that the feed works.
Users would have to set their station id again after this change.

# Copilot

copilot:all


glenrobertson:

glenrobertson:
recheck

PR #1992, opened 2023-11-05
Fixes retrieval method to be based on a 24hr random seed.

flynnt:
# Description
I had originally intended this to only show a character per day (to avoid rate limiting), but I realized later that the way I am using the `http` cache doesn't produce that outcome. That being said, the fact that I am still caching every response for 24 hours should probably keep the app from getting rate limited by the API. There are fewer than 1500 possible characters.

Anyway, this PR changes the app metadata to describe what it actually does, which is serve a new, random Marvel character whenever it renders.

# Copilot

### Summary
🎲🌟🔄

Updated the `marveloftheday` applet to show a random Marvel character instead of a fixed one per day. Changed the applet name, summary, and docstring in `manifest.yaml` and `marvel_of_the_day.star` accordingly.

> _`name` and `summary`_
> _changed to show random chars_
> _a lively applet_

### Walkthrough
*  Update applet name and summary to show random character ([link](https://github.com/tidbyt/community/pull/1992/files?diff=unified&w=0#diff-8ce0bec3ba03915672647b5b2454ab0629c62955d895a87e2bb25be9eef158dfL3-R4), [link](https://github.com/tidbyt/community/pull/1992/files?diff=unified&w=0#diff-955eba8fcec605510477c61b8554c0558f04f1c8c04ebbf3638a52b3da433756L2-R3))

### DIFF
--- a/apps/marveloftheday/manifest.yaml
+++ b/apps/marveloftheday/manifest.yaml
@@ -1,7 +1,7 @@
 ---
 id: marvel-of-the-day
 name: Marvel Characters
-summary: Randomized Marvel characters
+summary: Random Marvel characters
 desc: Shows the name and image of a Marvel Comics character using the Marvel API.
 author: flynnt
 fileName: marvel_of_the_day.star
--- a/apps/marveloftheday/manifest.yaml
+++ b/apps/marveloftheday/manifest.yaml
@@ -1,7 +1,7 @@
 ---
 id: marvel-of-the-day
-name: Marvel Characters
-summary: Random Marvel characters
+name: Marvel of the Day
+summary: A Marvel character a day
 desc: Shows the name and image of a Marvel Comics character using the Marvel API.
 author: flynnt
 fileName: marvel_of_the_day.star
--- a/apps/marveloftheday/manifest.yaml
+++ b/apps/marveloftheday/manifest.yaml
@@ -1,7 +1,7 @@
 ---
 id: marvel-of-the-day
-name: Marvel of the Day
-summary: A Marvel character a day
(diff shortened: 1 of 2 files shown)


matslina:
> I had originally intended this to only show a character per day (to avoid rate limiting), but I realized later that the way I am using the http cache doesn't produce that outcome. That being said, the fact that I am still caching every response for 24 hours should probably keep the app from getting rate limited by the API. There are fewer than 1500 possible characters.

I think you can get the originally intended behavior with a minor modification. Call `random.seed(time.now().unix // 86400` in `main()`. That'll make the RNG produce the same number over each 24h time window. The exact transition point from one graphic to the next is not entirely clear that way, but it'll do the trick I think. Also makes the app super easy for us to cache in the Tidbyt backend. =)

flynnt:
> > I had originally intended this to only show a character per day (to avoid rate limiting), but I realized later that the way I am using the http cache doesn't produce that outcome. That being said, the fact that I am still caching every response for 24 hours should probably keep the app from getting rate limited by the API. There are fewer than 1500 possible characters.
> 
> I think you can get the originally intended behavior with a minor modification. Call `random.seed(time.now().unix // 86400` in `main()`. That'll make the RNG produce the same number over each 24h time window. The exact transition point from one graphic to the next is not entirely clear that way, but it'll do the trick I think. Also makes the app super easy for us to cache in the Tidbyt backend. =)

Ah, thanks for that. I updated the code. It works locally, and now it's back to being in line with what I intended. I kind of like the advent calendar suspense of it being a once a day thing. :)

PR #2015, opened 2023-11-07
Fix caching issue related to use of dynamic query params.

flynnt:
# Description
Refactors caching in app to use the old school `cache` module as opposed to the native caching added to the `http` module. It appears as though the native option uses the full URL and its query params as the cache key, which in most cases is fine, but in cases where the query params contain some kind of timestamp or auth hash, this will lead to a situation where nothing really gets cached, and, in my case, quickly led to being rate limited by the API. This refactor should take care of that situation as well as eliminate the need to manually set a random seed.

Additional context here:
https://discuss.tidbyt.com/t/http-caching-client/5129/2?u=tflynn

# Copilot

### Summary
🚀🗄️🔐

The `marvel_oftheday` app is updated to cache and encode API responses. This improves the app's speed and stability.

> _Oh we are the `marvel_oftheday` crew_
> _And we love to code and cache and encode too_
> _We'll reduce the API requests with a clever trick_
> _And make our app more speedy and more slick_

### Walkthrough
*  Import modules for caching and encoding API responses ([link](https://github.com/tidbyt/community/pull/2015/files?diff=unified&w=0#diff-955eba8fcec605510477c61b8554c0558f04f1c8c04ebbf3638a52b3da433756R8-R10))
*  Implement caching logic for character ID and random character data, using base64 to handle binary data ([link](https://github.com/tidbyt/community/pull/2015/files?diff=unified&w=0#diff-955eba8fcec605510477c61b8554c0558f04f1c8c04ebbf3638a52b3da433756L31-R50), [link](https://github.com/tidbyt/community/pull/2015/files?diff=unified&w=0#diff-955eba8fcec605510477c61b8554c0558f04f1c8c04ebbf3638a52b3da433756L92-R116), [link](https://github.com/tidbyt/community/pull/2015/files?diff=unified&w=0#diff-955eba8fcec605510477c61b8554c0558f04f1c8c04ebbf3638a52b3da433756R126))

### DIFF
--- a/apps/marveloftheday/marvel_of_the_day.star
+++ b/apps/marveloftheday/marvel_of_the_day.star
@@ -5,6 +5,9 @@ Description: Shows the name and image of a Marvel Comics character using the Mar
 Author: flynnt
 """
 
+load("cache.star", "cache")
+load("encoding/base64.star", "base64")
+load("encoding/json.star", "json")
 load("hash.star", "hash")
 load("http.star", "http")
 load("random.star", "random")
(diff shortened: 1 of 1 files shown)


matslina:
Kinda looks like the ts parameters can be static?
> ts - a timestamp (or other long string which can change on a request-by-request basis)
If that doesn't work, then this type of caching is the only option at the moment. Would still suggest trying out passing a static (or very infrequently updated) `ts` parameter.

flynnt:
> Kinda looks like the ts parameters can be static?
> 
> > ts - a timestamp (or other long string which can change on a request-by-request basis)
> > If that doesn't work, then this type of caching is the only option at the moment. Would still suggest trying out passing a static (or very infrequently updated) `ts` parameter.

Good idea. Let me give that a shot.

flynnt:
The timestamp is now static, which should allow the cache key to remain stable for the specified `ttl`. Whew.

matslina:
Alright, let's give it a shot.

PR #2034, opened 2023-11-12
Updated multicountdown app to fix string > int comparison

BestDistress:
# Description
Updated multicountdown app to fix string > int comparison in configuring the number of events.

# Copilot

### Summary
🐛🔢🛠️

Fixed a bug in `multicountdown` app that could cause incorrect or invalid event display. Changed the `more_options` function to compare `eventCount` as a string instead of a number.

> _`eventCount` is string_
> _Compare with strings, not numbers_
> _Autumn bug is fixed_

### Walkthrough
*  Fix string comparison bug in `more_options` function ([link](https://github.com/tidbyt/community/pull/2034/files?diff=unified&w=0#diff-a23b99eeb8fbcbfe45317d5db9fbda5af1a7cf26f5bb25b64bcacfd4504c5975L113-R114))


BestDistress:
:shipit:

PR #2120, opened 2023-12-17
Mars clock - tweak am/pm suffix font

adam-aspect:
# Description
Pull the am/pm suffix into its own widget with separate font

# Copilot

copilot:all


adam-aspect:
:shipit:

PR #2128, opened 2023-12-19
Create Splatoon3Schedule app

Denton-L:
# Description
This app displays the current rotation of stages in Splatoon 3. To be honest, I didn't realise the other Splatoon 3 apps existed until I finished making this one. That being said, this app is different from those because it is more minimalistic. It will only display one mode's information, disappearing if the mode isn't active. In addition, it also supports Salmon Run.

Currently, this app doesn't support Splatfests, Big Runs nor Eggstra Work. I will implement those after this app is published.

splatoon3schedule
splatoon3schedule

# Copilot

copilot:all

### DIFF
--- a/apps/splatoon3schedule/manifest.yaml
+++ b/apps/splatoon3schedule/manifest.yaml
@@ -1,0 +1,8 @@
+---
+id: splatoon3schedule
+name: Splatoon3Schedule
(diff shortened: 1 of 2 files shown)


Denton-L:

PR #2130, opened 2023-12-21
Solarmanager_ch new ev logos

tavdog:
# Description
Add four new EV brand logos and schema options.  Open, Renault, Hyundai, Cupra

# Copilot

copilot:all


tavdog:
start over.

PR #2136, opened 2023-12-24
awair: add cloud API support via bearer token, improve layout, mock data view

flavorjones:
# Description

The big change in this PR is the addition of **support for the Awair Cloud API** via a Bearer token. (Unfortunately the Awair Cloud API quotas are quite low, so the network cache TTL is 5 minutes.) The schema uses `Generated` fields for the needed values, but the app should be backwards-compatible with the configs of anyone who's already installed and configured the Local API.

**Mock data is now rendered when the app is unconfigured**, so the app store webp is a bit more informational for users.

The **display has been improved** a bit: table numbers line up, with more consistent whitespace; and the score circle and font are larger. The color maps have been updated to match the values in the [official Awair docs](https://docs.developer.getawair.com/).

See the README for more information.

## Screenshots

For a sense of the visual layout changes:

|  | screenshot |
|---|---|
| before this PR |  |
| with this PR |  |
| unconfigured |  |
| api error |  |

/cc @tabrindle who wrote the original version of this app at https://github.com/tidbyt/community/pull/1762

# Copilot

copilot:all

PR #2137, opened 2023-12-24
awair: display bar graph

flavorjones:
# Description

NOTE: this PR is based off of #2136, please review that one first.

This PR introduces an alternative display configuration for a "bar chart" view (in place of the default data table view).

image

showing temperature, humidity, CO2, VOCs, PM2.5 in the same order as the Awair device and app.

# Copilot

copilot:all


flavorjones:
@tabrindle Any objections to this approach (now that I think #2136 is planned for merge)? Again I tried to make sure existing configs would continue to work.

PR #2139, opened 2023-12-24
Add: Chrono24

croossin:
# Description
Add a new Tidbyt app that shows watch market performance.

# Copilot

copilot:all


apps/chrono24/chrono24.star
```diff
@@ -0,0 +1,251 @@
...
+
+    # Check for cached data
+    cached_data = cache.get(timeframe)
+
+    if cached_data != None:
+        print("Hit! Displaying cached data.")
+        data = json.decode(cached_data)
+    else:
+        print("Miss cache!")
+        rep = http.get(BASE_URL + timeframe)
```
matslina, line 35: You can simplify this a bit by passing `ttl_seconds` to `http.get()`. It will cache the response accordingly. We may change the behavior of the `cache` module in the future, so using the http module's built in cache is a good choice.

PR #2164, opened 2023-12-31
Adds a ignore list to the live_on_twitch app

kevin-huff:
Modifies the live_on_twitch app to have an ignore list schema so you can exclude streams for the output. I did this because I follow several 24/7 always on streams and they are just cluttering up the display.


kevin-huff:

matslina:
@daltonclaybrook Thoughts?

apps/liveontwitch/live_on_twitch.star
```diff
@@ -277,7 +284,10 @@ def get_followed_streams(refresh_token, user_id):
         },
     )
     if res.status_code == 200:
-        return res.json()["data"]
+        streams = res.json()["data"]
+
+        # Filter out streams in the ignore list
+        return [stream for stream in streams if stream["user_name"] not in ignore_list]
```
daltonclaybrook, line 290: Does `ignore_list` ever get converted from the comma-separated string to a list? I suppose it probably doesn't matter because `not in` would still work with a substring. My one other small suggestion would be to use `.lower()` on both the `ignore_list` and `stream["user_name"]` to make the check case-insensitive. Otherwise, I think this looks good.

kevin-huff: Agreed. I've updated the pr with those changes.

daltonclaybrook (commented):
Looks good to me! Thanks for this addition.

PR #2179, opened 2024-01-06
New app: Maya Glyphs

dinosaursrarr:
# Description
Displays a random glyph from the Maya script together with the suspected pronunciations.

Glyphs are sourced from the Mayan Epigraphic Database Project http://www2.iath.virginia.edu/med/. It may not be the most up to date source, but it looks good on tidbit and anyway, this writing system went largely undeciphered for about 500 years until the 1970s.

Anyone who enjoys this app might also want to read _Breaking the Maya Code_ by Michael D Coe. It's a cracking read!

The app is large but it's mostly just static data pasted at the bottom. I've not included the script used to process the images and generate this dict, but only because it blocked the automatic review approval.

# Copilot

copilot:all


dinosaursrarr:
:shipit:

PR #2180, opened 2024-01-06
Maya Glyphs: add prep script

dinosaursrarr:
# Description
I added a new app in https://github.com/tidbyt/community/pull/2179

It relies on a large dict of static data, prepared from the original source. This is the script I used to generate that dict, in case the app ever needs to be updated. I've split it out just because having a python file in the app dir blocked automatic submission.

# Copilot

copilot:all

### DIFF
--- a/apps/mayaglyphs/prep.py
+++ b/apps/mayaglyphs/prep.py
@@ -1,0 +1,92 @@
+# Downloads and prepares glyph data in a format that can be pasted into a starlark app.
+
+import base64
+import io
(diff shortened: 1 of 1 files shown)


dinosaursrarr:
Ah - today I learned. The problem wasn't having a random python script included. It was that originally I had it in a subdirectory. Now it's at the top level and all is hunky dory.

dinosaursrarr:
:shipit:

PR #2181, opened 2024-01-06
New app to display WeatherSTEM data

imaginuts:
# Description
This is a new application to display current conditions provided by the WeatherSTEM API. Free registration with WeatherStem is required to aquire an api token. This allows current weather conditions to be displayed for a nearby weather station, or a favorite location users may want to monitor.

weatherstem

# Copilot

copilot:all

### DIFF
--- a/apps/weatherstem/manifest.yaml
+++ b/apps/weatherstem/manifest.yaml
@@ -1,0 +1,8 @@
+---
+id: weatherstem
+name: WeatherSTEM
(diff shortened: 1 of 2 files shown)


imaginuts:

matslina:
Would be great if the app could display a default graphic showing users what it'll look like once it's set up. That way we can show previews in the mobile app, and folks can get a sense of if it's worth the effort of going out to get their own API keys and so on.

PR #2183, opened 2024-01-07
TF2-Tidbyt-Ticker

FrontierProphet:
# Description
Got a tidbyt for christmas, this is the first app I've attempted. This keeps track of the key price on the tf2 market measured in "refined" via backpack.tf. Not expecting much, just seeing if it works out

Thank you backpack.tf!

# Copilot

copilot:all


FrontierProphet:

FrontierProphet:
If there are any issues, I assume it's with line 12. I am hoping to hide my backpack.tf api key using pixlet encrypt command, but it's difficult to check the success of that client side. App runs and passes all checks when line 12 is my personal API key

FrontierProphet:
don't add this yet I am going insane, but I think my initial problem lies in ignoring config

PR #2186, opened 2024-01-07
Shrink static data by converting to greyscale

dinosaursrarr:
# Description
The images are already monochrome, so we can save a fair bit of space by throwing away the unnecessary colour info. Saves a megabyte or two. No visible changes.

# Copilot

copilot:all


dinosaursrarr:
:shipit:

PR #2187, opened 2024-01-07
Fix: make hieroglyphs random again

dinosaursrarr:
# Description
Ooops, I added a hardcoded override to make some interesting screenshots. Forgot to take it out. So it's always showing the same one. Easily fixed.

# Copilot

copilot:all


dinosaursrarr:
:shipit:

dinosaursrarr:
:shipit:

PR #2185, opened 2024-01-07
New app: Hieroglyphs

dinosaursrarr:
Similar to yesterday's https://github.com/tidbyt/community/pull/2179 but for a different ancient script.

Thanks to [Wikipedia](https://en.wikipedia.org/wiki/List_of_Egyptian_hieroglyphs) and [Wikihiero](https://www.mediawiki.org/wiki/Extension:WikiHiero) for providing the content in a useable format and under a free licence.

Displays a new glyph every 15 seconds, showing:

- Gardiner sign number (teal)
- Pronunciation, in Manuel de Codage convention (white)
- Description (gold)

Manuel de Codage isn't the nicest transliteration format, but it is the easiest to work with Tidbyt's pixelated fonts. Good luck using the more typical ꜥ, ꜣ, š, ṯ, ḥ, ḫ, ẖ, and the rest.

# Description
Replace this with a description of your changes.

# Copilot

copilot:all

PR #2190, opened 2024-01-08
New app: Cuneiform

dinosaursrarr:
Similar to https://github.com/tidbyt/community/pull/2179 and https://github.com/tidbyt/community/pull/2185 but for another ancient script.

Shows a new cuneiform sign every 15 seconds, plus it's name (according to Unicode) and transliteration in Sumerian. I wish I had the sources or knowledge to also support the Akkadian language, but I leave that as an exercise for the avid user.

There is some text mangling to make it display on Tidbyts. Notably, subscripted numbers aren't subscripts and `ḫ` is displayed as `ħ`. But I'm impressed as I expected many fewer of the necessary characters to work in any of pixlet's fonts.

# Description

# Copilot

copilot:all


dinosaursrarr:
:shipit:

PR #2192, opened 2024-01-08
Add provider and country filtering to rocket launch data

kevin-huff:
# Description
This PR introduces a new feature that allows users to filter the rocket launch data based on the provider and country. The filtering is done in the `filter_rocket_launches` function, which takes the rocket launch data and comma-separated strings of providers and countries to filter for.

If no provider or country is specified, all data is returned. If all results are filtered out due to the applied filters, a message "All launches filtered.." is displayed. This ensures that users are informed when their filters result in no data.

This feature enhances the usability of the rocket launch data by allowing users to focus on the launches that are most relevant to them. It also provides feedback when the applied filters are too restrictive.

# Copilot

copilot:all

### DIFF
--- a/apps/launchcountdown/launchcountdown.star
+++ b/apps/launchcountdown/launchcountdown.star
@@ -126,13 +126,38 @@ def get_rocket_launch_json():
 
                 # TODO: Determine if this cache call can be converted to the new HTTP cache.
                 cache.set(ROCKET_LAUNCH_CACHE_NAME, json.encode(rocket_launch_data), ttl_seconds = cache_time_seconds)
+                # Filter out any providers in ignoredProviders
 
     return (rocket_launch_data)
+
+def filter_rocket_launches(rocket_launch_data, filter_for_providers_string, filter_for_countries_string):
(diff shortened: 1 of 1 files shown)


matslina:
@rjison

rjison:
@matslina I'm cool with these changes. 

@kevin-huff These are nice options. 

I think I'll add the following later on though:
1) Allow folks to hide the display altogether if no upcoming launches (Not done now because there will probably be 5 upcoming launches, but maybe not after filtering). Personally, I don't like to see notifications of nothing on my Tidbyt :)
2) Provide an option to turn on instructions and provide examples of countries and 'providers'. (I had to look at the JSON to figure out 'United States' (wasn't sure if it was USA, or United States of America..) and The providers need to be spelled out exactly (Mitsubishi doesn't work, you'd need 'Mitsubishi Heavy Industries')

kevin-huff:
@rjison yeah I thought about updating the display to show the provider and country but there's already a lot being shown. I think most people are just gonna filter for United States or SpaceX anyway.

PR #2207, opened 2024-01-14
Add Multi-Calendar App

logandeal:
# Description
This adds a new multi-calendar app. The app displays the next event (within the next 24 hours) on one or more Google calendars. By default, it grabs the next event from all of a user's calendars when no Google Calendar IDs are specified. Optionally, a user may specify a comma-delimited list of Google Calendar IDs that the app will use when displaying the next event. If there is no event in the next 24 hours, it simply tells the user that there are no events in the next 24 hours. When the current time is more than half past the current event's duration, the next event will be displayed.

This app also supports 12 and 24-hour time.

# Copilot

copilot:all

### DIFF
--- a/apps/multicalendar/manifest.yaml
+++ b/apps/multicalendar/manifest.yaml
@@ -1,0 +1,8 @@
+---
+id: multicalendar
+name: Multi-Calendar
+summary: Next event from calendars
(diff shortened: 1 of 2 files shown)


logandeal:

PR #2208, opened 2024-01-14
Add Multi-Calendar App

logandeal:
# Description
This adds a new multi-calendar app. The app displays the next event (within the next 24 hours) on one or more Google calendars. By default, it grabs the next event from all of a user's calendars when no Google Calendar IDs are specified. Optionally, a user may specify a comma-delimited list of Google Calendar IDs that the app will use when displaying the next event. If there is no event in the next 24 hours, it simply tells the user that there are no events in the next 24 hours. When the current time is more than half past the current event's duration, the next event will be displayed.

This app also supports 12 and 24-hour time.

# Copilot

copilot:all

### DIFF
--- a/apps/multicalendar/manifest.yaml
+++ b/apps/multicalendar/manifest.yaml
@@ -1,0 +1,8 @@
+---
+id: multicalendar
+name: Multi-Calendar
+summary: Next event from calendars
+desc: Displays next event from multiple Google calendars.
+author: Logan Deal
+fileName: multicalendar.star
+packageName: calendar
--- a/apps/multicalendar/multicalendar.star
+++ b/apps/multicalendar/multicalendar.star
@@ -75,7 +75,7 @@ def get_earliest_events_from_calendars(calendars, window_24h, token):
     earliest_list = []
     i = 0
     for calendar_id in calendars:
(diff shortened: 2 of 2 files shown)


logandeal:

logandeal:
@matslina Is there an issue with the code?

logandeal:
Sample Gif
Tidbyt

matslina:
@logandeal Thanks for the PR, and apologies for the delay. We've had a very busy couple of weeks, and all PRs have suffered as a consequence. Especially less trivial ones.

And I'm afraid this PR is not entirely trivial. I'll do my best to test it out over the next few days. We'll need to migrate the app to use Tidbyt's oauth2 credentials, which (sometimes) cascades into a painful review process with Google. Now that I look at it, I think we already have these scopes covered, which means it might be fairly straightforward. Stay tuned, and again, apologies for the delay.

logandeal:
@matslina Any updates? I would really like to have this app get released so that we can all use multiple calendars!

logandeal:
Keeping this open.

kaffolder7:
Can we please keep this open? This would be _super_ helpful for us. @logandeal has done a lot of the legwork. 🥳

PR #2209, opened 2024-01-14
Splatoon 3 Stages: Rewrite internals, add Salmon Run support

MarkGamed7794:
# Description
Rewrote the code to (hopefully) be cleaner and easier to follow. Also added Salmon Run support, and individual toggles for each pane of Splatfest information.

Also, I'd like to shout out the author of the other Splatoon 3 schedule app on their incredible work, which I wish no harm against.


Denton-L:
Cheers!

PR #2211, opened 2024-01-15
fix: (apps/awair) work around generated fields without default values

flavorjones:
# Description

This is an attempt to work around an issue in the mobile app described at https://discord.com/channels/928484660785336380/928638975298650172/1182047678914367638

The app hangs when configuring the app's generated fields. Poking around on discord, I found the above response from @matslina that describes an issue with generated fields when they don't have default values declared.

In this PR, I'm setting the default value to an empty string, but I'm not sure if this is sufficient to fix the issue (since the web dev environment isn't a complete reproduction of the mobile app's config screen).

# Copilot

copilot:all

### DIFF
--- a/apps/awair/awair.star
+++ b/apps/awair/awair.star
@@ -52,6 +52,7 @@ def api_connection_options(api_connection_type):
                 name = "Public IP address and port of Awair device",
                 desc = "Requires a public IP address with port forwarding to an Awair device configured for Local API access.",
                 icon = "computer",
+                default = "",
             ),
         ]
     elif api_connection_type == "cloud_token":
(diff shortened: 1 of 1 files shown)


flavorjones:
@matslina I'm curious if you think this is an appropriate workaround for the bug you described in Discord?

matslina:
Hey. So sorry for the radio silence. Been heads down on Gen 2 firmware for the past few weeks, and PRs have unfortunately suffered. Happy to merge this. If there are issues, hit me up on Discord and I'll roll up my sleeves and figure it all out.

flavorjones:
@matslina Thank you! App is working better, I can configure it. The generated fields don't seem to work properly if I edit an already-configured app (I need to re-enter them) but otherwise all good. :+1:

PR #2215, opened 2024-01-17
Create "Quote of the Day" App

taylorchasewhite:
# Description
quote_of_the_day

Quote of the Day allows users to see an inspiring, funny, or otherwise wise quotes throughout the day. There are more than 70 quotes currently, with more on the way. 

Users can configure the following:

`Duration of Quote`:
_How long the same quote will show for the app._
image

`Randomize Colors`
image

Whether to randomly select from the palette, or to use the user's color choices.

`Quote and Author Colors`
_The colors for the author and quote text_
image

image

# Copilot

copilot:all

PR #2232, opened 2024-01-29
Update: Quotetopia (fka A Quote A Day)

taylorchasewhite:
# Description
Updates to #2215:
- [x] Adds additional quotes
- [x] Changes the .gitignore to ignore Visual Studio generated files (.vs) for the repo (no one should be committing these).
- [x] Changes the name of the applet from A Quote a Day to Quotetopia (more fun 🤪)

quotetopia

Various color options and defaulting are all still available:
image
image

quotetopia

# Copilot

copilot:all


matslina:
Hey, thanks for the PR!

Changing app ID like this is a bit problematic. It'll break any existing installations of the app, unless we put time into doing some sort of migration. I would suggest you keep the current ID, as well as the current path (i.e. under `apps/quoteoftheday/`) and only changing the other metadata in the manifest file. E.g., you can totally keep this current ID, but change the `name` field in the manifest file.

taylorchasewhite:
> Hey, thanks for the PR!
> 
> Changing app ID like this is a bit problematic. It'll break any existing installations of the app, unless we put time into doing some sort of migration. I would suggest you keep the current ID, as well as the current path (i.e. under `apps/quoteoftheday/`) and only changing the other metadata in the manifest file. E.g., you can totally keep this current ID, but change the `name` field in the manifest file.

Ah, thanks, okay, I can do that, I'll take a look in a few days (out of town at the moment!).

PR #2235, opened 2024-01-31
New App: Melbourne Trains

bendiep:
# Description
This is a Train Transit app for Melbourne, Australia. 
It retrieves real-time data from the official train network API in Melbourne (PTV - Public Transport Victoria) and displays it the user.
Specifically the app displays;  Departure time in minutes, Departure direction, Departure station and Departure platform number. 
It supports the entire train network in Melbourne, so users can select train lines / stations that are relevant to them.

# Screenshot
melbourne_trains

# Copilot

copilot:all


bendiep:
recheck

bendiep:
recheck

bendiep:
Hi Tidbyt fam

I'm not quite sure why the Preview generated above can't read my API credentials via. the use of `secret.decrypt()`

bendiep:
recheck

matslina:
Hey @bendiep . Thanks for the PR.

The secrets issue here is annoying. Our CI set up can't decrypt secrets, so it'll always fail in this stage. Once the app makes it to the production environment, secrets can be encrypted. I'll merge and hopefully it all works out.

To answer your questions about the encrypt command: yes, the app ID should be the same as in the manifest file, and whether or not to quote the secret data strictly speaking depends on your shell, but in general the answer is yes.

bendiep:
Hi @matslina 

Thanks for the review!
It doesn't seem to working on my Tidbyt device, I just see a `Cloud Icon`.
(I'm not sure why though, because I have a default render image to display for errors)

**In my latest commit, I changed the key encryption command:**
I used:
```
pixlet encrypt <packageName value> "<key>"
```
Instead of:
```
pixlet encrypt <id value> "<key>"
```

Should I re-raise another PR to correct the encryption keys?

PR #2239, opened 2024-02-01
splatoon3schedule: support null boss object for Eggstra Work

Denton-L:
# Description

As it turns out, in Eggstra Work, the boss object is null. Support this by not displaying a boss string in case the corresponding object is null.

# Copilot

copilot:all


Denton-L:
Rendered:

splatoon3schedule

PR #2245, opened 2024-02-03
New Applet: Virtual Pets

frame-shift:
# Description
This will create a new applet called 'Virtual Pets.' Users can select and name one of 10 pets to follow. The environment changes with the user's current season and time of day.

# Copilot

copilot:all


matslina:
Thanks for the PR. Looks great!

One thing though: could you do 1 PR per app going forward? Just makes it a bit easier for us to overview and review.

frame-shift:
> Thanks for the PR. Looks great!
> 
> One thing though: could you do 1 PR per app going forward? Just makes it a bit easier for us to overview and review.

Will do!

PR #2246, opened 2024-02-03
New App: Billboard Top 10

rjison:
# Description
Displays the Top 10 songs from either US, Japan, or Global.
Pick your colors for the display text.
Uses RapidAPI Key - Programmed to pull data < 20 times a month which is the free level.

Request some help with the API Key, hoping Tidbyt can create their own RapidAPI Key in place of mine.

If you guys are willing to pay for $5 a month, I could rewrite the caching logic to check more often, but cache for a full 7 days once a new list is detected. That should make the updates more timely. Otherwise, no worries, close enough! :)

# Copilot
copilot:all


matslina:
Hey, thanks for the PR!

I'd prefer not to pick up the RapidAPI bill. No so much because of the cost, as the hassle of keeping track of yet another recurring payment. Paperwork's needed for each and every one of them... =\

PR #2250, opened 2024-02-04
New App: Park Wait Times

hx009:
# Description
Displays theme park ride wait times at various theme parks.

# Copilot

copilot:all

### DIFF
--- a/apps/parkwaittimes/manifest.yaml
+++ b/apps/parkwaittimes/manifest.yaml
@@ -1,0 +1,8 @@
+---
+id: park-wait-times
+name: ParkWaitTimes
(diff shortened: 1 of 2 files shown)


matslina:
Not sure if it's sufficient attribution, but most API providers are content with it being given in the app description, as opposed to in the graphic itself. Might make for a cleaner look.

hx009:
> Not sure if it's sufficient attribution, but most API providers are content with it being given in the app description, as opposed to in the graphic itself. Might make for a cleaner look.

Not a bad idea! I didn't even think of that.  I'll have to get a PR in for that soon.  Thanks!

PR #2251, opened 2024-02-04
create app for displaying real time departures of viennas public tran…

Impeergator:
# Real Time Departures for Public Transport Stops in Vienna

With this app, users can choose up to for stops of Viennas huge public transport system for which real time departures will be displayed.
The app features an animation where each stop is featured with a seperate display.

On each display, the following is shown:

- Vienna Public Transport Service (Wiener Linien) Logo 
- Name of the stop the user wants the departures of
- A row for each departure at the stop which is given by line, destination and departure times

In Vienna, a stop can have multiple departures. In the example we chose the stop 'Stephansplatz' which features 2 metro lines (U1/U3) in two directions.
Hence, a total of 4 departures is given.
To ensure that all departures are visible, a maximum of 3 departures is shown per display.
If the number of departures exceeds this, a new display is rendered.

If the user chooses to display more than one stop, a new display is rendered for the new stop.

Furthermore, the user can choose how fast the animation should jump to the next disply.

Note that texts and instructions in for the user configuration is in German since it is anticipated that mainly people living in Vienna - hence able to speak at least basic German - will use this app.


Impeergator:

Impeergator:
Regarding the 'Usage of http.star requires review' bullet point mentioned by the bot:
The Vienna Public Transport Company mentions that it urges user to limit API requests to 4 per minute. In this app, when obtaining the realtime data, ttl_seconds was set to 30, i.e. only upstreaming twice per minute which is well within the requirement.
Source: https://www.data.gv.at/katalog/dataset/522d3045-0b37-48d0-b868-57c99726b1c4
It's in German but a simple google translate will do the job, I guess.

matslina:
While an individual user of the app (more specifically, an individual combination of stop IDs) is unlikely to exceed the 4 requests per minute limit, it seems likely that the app will exceed the limit as soon as multiple users have it installed. Not a violation of the spirit of the rate limit, but possibly something they'll crack down on anyway.

PR #2257, opened 2024-02-08
mastodonfollows: switch from ActivityPub API to Mastodon API V1

smallsco:
# Description
The Mastodon Follows app currently does not work with instances that have "authorized fetch" enabled. This is because it is using the raw ActivityPub API, and not signing its request.

Example: https://oldbytes.space/users/smallsco/followers
Returns HTTP 401 with a body of `{"error": "Request not signed"}`

I have updated the app to use the higher-level Mastodon API V1 instead, which does not require request signing.

Example: https://oldbytes.space/api/v1/accounts/lookup/?acct=smallsco
Returns a JSON blob with all public info about my account, including followers in the `followers_count` key.

I have also updated the default user and instance, as that user has recently changed her name and moved accounts (the old account is a redirect).

# Copilot

copilot:all

### DIFF
--- a/apps/mastodonfollows/mastodon_follows.star
+++ b/apps/mastodonfollows/mastodon_follows.star
@@ -47,12 +47,12 @@ irBThl5EZKG0l5xrlQL/Lw2O4kHOCHzzH6J6eeTaYhipO9wa5uBYgNMHfs2MKyLvrjk=
 """)
 
 def main(config):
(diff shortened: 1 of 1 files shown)


smallsco:

matslina:
@drudge LGTM, LGTU?

drudge:
Looks good! Thanks for updating it @smallsco!

PR #2262, opened 2024-02-08
Hot fix: Virtual Pets timezone

frame-shift:
# Description
Now calculates the current time based on the user's timezone (previously was always UTC time)

# Copilot

copilot:all


matslina:
Note that you can also get users timezone (or their device's, more specifically) from the magic `$tz` config variable.

PR #2264, opened 2024-02-09
added a new app to track multiple CTA train arrival times at a station;

imcnaugh:
# Description
Adds a simple app to tidbyts that displays up to 4 Chicago CTA train arrival times at a given station, my biggest use case is having this displayed in the mornings to know when i need to leave the house to catch what train.

Destinations scroll past in the color of the train line, with the minutes to arrive at the selected station on the right.

# Copilot

copilot:all


imcnaugh:

imcnaugh:
@matslina is the delay between when i use `pixlet encrypt` to set the api key, and call `secret.decrypt(` significant? The build and test process seems to be having trouble finding my api key.

PR #2265, opened 2024-02-09
Cta train app

imcnaugh:
# Description
Adds a simple app to tidbyts that displays up to 4 Chicago CTA train arrival times at a given station, my biggest use case is having this displayed in the mornings to know when i need to leave the house to catch what train.

Destinations scroll past in the color of the train line, with the minutes to arrive at the selected station on the right.

# Copilot

copilot:all


imcnaugh:
Yeah I'm at a loss, the build and test step can not seem to find the api key i uploaded using `pixlet encrypt cta-train-tracker <api_key>`, I dont know if there is a delay in it showing up after adding it with the pixlet command. or if I have something wrong in my manifest, either way ill come back to this tomorrow.

imcnaugh:
@matslina sorry to call you out but any advice would be appreciated, the build step is failing to render the app because the secret key for the api im using is not found. 

I understand I need to run `pixlet encrypt {appId} {secret}`, but I must be missing something. 

Im assuming my appId is coming from the manifest file, so my command looks like

`pixlet encrypt cta-train-tracker {secret}` but still, the render process cant find the key.

any advice would be appreciated.

matslina:
The issue is most likely in our CI setup. It's separated from our production environment, and does not have access to the keys required to decrypt a secret, so the call to `secret.decrypt()` will always return `None` here. Very annoying... =(

If you add a special case for the `None` case, we should be able to get it through.

apps/cta_train_tracker/cta_train_tracker.star
```diff
@@ -0,0 +1,732 @@
...
+
+    return output
+
+def main(config):
+    train_dir = config.str("directions", DEFAULT_DIRECTION)
+    map_id = int(config.get("mapId", DEFAULT_MAPID))
+    api_key = secret.decrypt(ENCRYPTED_API_KEY)
+
+    if api_key == None:
+        fail("api key not found")
```
matslina, line 100:
```suggestion
        return []
```

imcnaugh:
Ah i see, in that case i might need to make one more change to handle a bad request to the api, but that helps a bunch, thank you so much!

PR #2271, opened 2024-02-11
Fix: Wiki Page Today

sch0lars:
- Added logic to retrieve previous article if current article is not available
- Added logic to print a failure message if no featured article can be found
- Added default thumbnail to address missing key error
- Added language feature
- Added font color feature

# Description
The current app was not displaying featured articles, oftentimes failing because of unhandled exceptions. I have added fixes for these issues, including:
- Adding a default thumbnail when the featured article does not contain one
- Checking for the previous day's featured article when the current day's article cannot be found
- Adding a failure message when neither of the aforementioned articles can be found

I have also added two additional features:
1. Language options to display the featured article in different languages
2. An option to change the font color

# Copilot

copilot:all


matslina:
@theUnBurn Thoughts on this one?

theUnBurn:
Looks good to me! Thanks for updating this, and appreciate you taking the time to do this!

PR #2275, opened 2024-02-13
Fix: Melbourne Trains

bendiep:
# Description

Use the correct `encryption command` for the encrypted keys used in the app.

**before:**
```
pixlet encrypt <manifest packageName value> "<key>"
```

**after:**
```
pixlet encrypt <manifest id value> "<key>"
```

# Copilot

copilot:all


bendiep:
Hi @matslina 

I've updated the encryption command used to encrypt the keys for this app.

The app is working perfectly locally.
However in the Tidbyt app store, my app just shows a `cloud icon` with a cross through it.
Can you give me some details on how I can debug such an issue?

matslina:
Looks like this PR fixed the problem? I'm at least seeing departures now, although I'm admittedly not up to speed on Melbourne local transit. Would love to go there sometime though.

PR #2276, opened 2024-02-13
Create CSV Viewer applet

mokolodi1:
# Description
This creates an app called CSV Viewer that allows for the viewing of CSV files available online. I'm using this personally to display published Google Sheet data on my Tidbyt.

The README has documentation on how exactly to format the text in the CSV file for changing text and background colors. Once this PR is merged, I'll link to that README from the [published CSV file (linked)](https://bit.ly/tidbyt-csv-viewer) in order to allow folks to discover the docs.

# Copilot

copilot:all


mokolodi1:
In order to facilitate testing of this applet, I'm documenting some of the testing I did below:

1. **Testing with a Google Sheet** (for color interpretation and for my own purposes): Create a Google Sheet with data you'd like to test (similar to [the default](https://bit.ly/tidbyt-csv-viewer), for example). Then do `File` => `Share` => `Publish to web` and select `Comma-separated values (.csv)`. Use the link that it gives you as input for the CSV URL and watch data and colors change as you change the Google Sheet. (Published sheets update every 5 minutes on Google's side and is separately cached on the Tidbyt with a TTL of 5 minutes, so it'll take up to 10 minutes to update.)
1. **Testing with an online data source other than Google Docs**: I found a list of online CSV files published [here](https://people.sc.fsu.edu/~jburkardt/data/csv/csv.html). I tested with [the first one that contains several addresses (csv file linked)](https://people.sc.fsu.edu/~jburkardt/data/csv/addresses.csv).
   1. There should be a 6 by 6 grid with scrolling text in each box representing the data in the CSV file
   2. Change the `Row count` and `Column count` to 2 each, and check that that matches the data in the top left 2x2 square of the CSV file.
   3. Keeping the other options the same, select the `Avoid scrolling text` option and check that the data stops scrolling as it can now be fit within the bounds of the now larger boxes. 
   4. Keeping the other options the same, change the `Row count` to 3 and check that only the data that cannot be fit horizontally scrolls (bottom left cell). 
   5. Keeping the other options the same, change the `Row count` to slowly increasing values and check that the fonts change size appropriately. At 8 rows, the text will stop being able to be rendered within the rows but will be shown nonetheless. (**Note: this is a known limitation of this app, so perhaps it's better to display a warning instead of showing cut-off text, but I don't think that's an important fix given it should be relatively clear what's happening.**)

Known issues:
- As noted above, funny things happen with more than 8 rows when displaying text, and there is a similar issue with many columns.
- Large and malformatted CSV files are not accepted and will fail without a good error message. I could theoretically add error checking for this, but I don't see it as a huge issue personally... For example: https://people.sc.fsu.edu/~jburkardt/data/csv/cities.csv 
   - ```
     Traceback (most recent call last):
        csvviewer.star:312:52: in main
        csvviewer.star:75:28: in get_data_rows_cols
        csvviewer.star:30:24: in get_csv_data
     Error in read_all: parse error on line 1, column 9: bare " in non-quoted-field```

mokolodi1:

matslina:
Thanks for the PR!

PR #2302, opened 2024-02-25
BtcFagi: Fix font size on 'extreme' in classification

PMK:
# Description
Fixes the font size if the classification contains 'extreme'. The font will be smaller to print all text.

# Copilot

copilot:all


PMK:
recheck

PR #2303, opened 2024-02-25
StocksFagi: Fix font size on 'extreme' in classification

PMK:
# Description
Fixes the font size if the classification contains 'extreme'. The font will be smaller to print all text.

@jondkelley, please approve.

# Copilot

copilot:all


matslina:
@jondkelley Any thoughts on this one? I'm inclined to merge unless we hear back from you soon,

PR #2304, opened 2024-02-25
Accuweather: cache response per location for 1h

glenrobertson:
# Description
Update the accuweather app to cache responses per location for 1 hour, to prevent from hammering the API unnecessarily.

# Copilot

copilot:all


glenrobertson:
@sudeepban thoughts on adding a 1h cache?

matslina:
Hey @glenrobertson. Sorry for the very delayed review here.

I'm not sure this will do anything to lower the request rate, as the http module already caches requests. The existing `ttl_seconds=3600` should handle this. I might be missing something though?

glenrobertson:
Yea np we can close it. I ended up using nginx caching for my custom pixlet server since the way I'm using it couldn't make use of the cache anyway. Thanks for taking a look though!

PR #2311, opened 2024-02-28
Initial version of app to display dinner specials at Selland's restaurants

matthewkent:
# Description
Show weekly dinner for two specials at Selland's restaurants in the Sacramento area.
sellands_specials

# Copilot

copilot:all


matthewkent:
Thanks for flagging. I noticed this and have been trying to figure out another approach but I think I'm out of luck. The app is quite simple, just makes one GET request and then parses some HTML, but I think the web site is too slow to reliably return under 500ms.

PR #2332, opened 2024-03-15
Add chart_name to Datadog Charts app to select dashboard chart

gabe-ochoa:
# Description
Allows people to select which chart they want to display from a Dashboard. Also adds in some better error handling if the response payload is too large. 

# Copilot

copilot:all


apps/datadogcharts/datadog_charts.star
```diff
@@ -29,13 +29,14 @@ def main(config):
...
     CHART_TIME_RANGE = config.get("chart_time_range") or "1h"
+    CHART_NAME = config.get("chart_name")
 
     ## APIs
     DD_DASHBOARD_API = "{}/dashboard".format(DD_API_URL)
     DD_METRICS_QUERY_API = "{}/query".format(DD_API_URL)
 
     if DD_API_KEY == None or DD_APP_KEY == None:
-        return redener("Set Datadog API and APP Key", fake_chart_data())
+        return renderer("Set Datadog API and APP Key", fake_chart_data())
```
gabe-ochoa, line 39: Fixed the typo :)

apps/datadogcharts/datadog_charts.star
```diff
@@ -97,6 +107,15 @@ def main(config):
...
         return render.Root(child = child)
+    elif query_response == None:
+        # The Datadog API sometimes returns > 50MB of data which causes a context deadline exceeded error
+        print("Error fetching data. Possibly too large of a response.")
+        print(query_response)
+        child = render.Marquee(
+            width = 48,
+            child = render.Text(content = "Error fetching data. Possibly too large of a response.", font = "6x13", color = "#fff"),
+        )
+        return render.Root(child = child)
```
gabe-ochoa, line 118: This was a bit hard to track down. Just kept getting "context deadline exceeded" with no other information. Payload size was too large from the DD API I think for some charts depending on the time range.

apps/datadogcharts/datadog_charts.star
```diff
@@ -174,7 +195,7 @@ def redener(display_name, datapoints):
 def get_schema():
     dd_site_options = [
         schema.Option(
-            display = "US",
+            display = "US1",
```
matslina, line 198: Note that these changes can break existing installations of the app. One way around that is to add a translation step in code. We unfortunately don't have a way of automatically migrating existing app configuration yet.

gabe-ochoa: Ah, ok. Just fixing some confusion but we can remove this line for now.

matslina: Oh, actually, this one isn't a problem. =) But the ones changing the value attribute would be.

gabe-ochoa:
@matslina Can this be merged since there's no issue with breaking changes?

matslina:
Sure thing. Just to be clear, by "the ones changing the value attribute", I meant the `"chart_time_range"` schema field. Existing installations will still pass in the former values, like `"1d"` instead of the new `"24h"`. Guessing the API handles both?

PR #2344, opened 2024-03-22
CFL Scores App

cmarkham20:
# Description

# Copilot

copilot:all


apps/cflscores/manifest.yaml
```diff
@@ -0,0 +1,8 @@
+---
+id: cfl-scores
+name: CFL Scores
+summary: CFL football scores
+desc: Creating multiple app instances is no longer necessary. Just install once and remove any other instances. You can set the rotation speed of scores in the settings.
```
matslina, line 5: This description might be redundant right now? Could be useful to explain what CFL is, etc?

PR #2346, opened 2024-03-23
CFL Scores App

cmarkham20:
# Description

# Copilot

copilot:all


jmanske:
Feel like there needs to be some "grandfathered" in way to allow fixes/updates to apps that do not make the runtime overall worse even if they are violating the `pixlet check` wall-clock runtime length limit.

cmarkham20:
Yeah I agree - or at least until we figure out a way to localize the logos as that takes most of the render time.

jmanske:
This gives me an idea to start a repo of base-64 encoded strings of logos of all teams across all sports. I already have NFL and MLB....

PR #2352, opened 2024-03-28
Update jacket.star; use updated API endpoint

WinslowJosiah:
# Description
OpenWeather's One Call API 2.5 [will be deprecated on April 15, 2024](https://openweathermap.org/one-call-transfer). In anticipation of this (and because this app is telling me I don't need a jacket in 28°F weather), I changed the API version from 2.5 to 3.0. This appears to be just a simple drop-in replacement, for this use case.

# Copilot

copilot:all


matslina:
@noahpodgurski Cool to merge?

PR #2353, opened 2024-03-28
visualizercoffee: update token path

matslina:
# Description

# Copilot

copilot:all


matslina:
@jeffbean This seems to fix the issue. I can now connect the app to visualizer.coffee, and have verified that the token works as expected.

PR #2354, opened 2024-03-28
Update LeetCode Stats - Update histogram max length calculation

bendiep:
# Description
Updated the calculation for `histogram maximum length` to prevent the `solved count` from overflowing beyond the Tidbyt screen width. 

This problem seems to occur when the distribution of solved problems is heavily skewed towards one LeetCode difficulty-level (in my case Easy).

# Copilot

copilot:all


jmanske:
I'll take a look tonight or tomorrow morning. My only hesitation would be that it's reducing the size of all the histograms because one might be too long. Thanks to @bendiep for contributing the fix and I will review it soon!

jmanske:
Hey @bendiep,

Thanks again for taking the time to submit a fix! I'm happy that you like the app and that it's useful for you. 

What you've shown me though is we actually have a fair number of edge cases; if a user has _only_ one type of problem solved, they'll still get pushed off the screen!

Let's do this:
pass the number solved into histogram attributes function on line 78:
```python
attr = get_HistogramAttributes(type, pct, statsObj.HistogramMaxLength, number_solved)
```
update the signature of the method to match on line 107:
```python
def get_HistogramAttributes(type, percentage, total_length, number_solved):
```
then replace the `bar_width` calculation with this code to make sure no one ever goes off the screen again:
```python
    number_solved_len = len(str(number_solved))
    proposed_width = math.ceil(float(percentage) * int(total_length))
    bar_width = proposed_width if proposed_width <= 64 - 5 * number_solved_len else 64 - 5 * number_solved_len
```

That should do it! I tested both with your user and some other ones (funnily enough, "bendi" was a user that has only solved easy problems and showed me we still had another edge case).

If you make those changes, I'm on board. Thanks again!

bendiep:
Thanks for the review @jmanske!

I've applied your suggested changes and reverted my previous change.

And yes I love this tidbyt app! I'm new to LeetCode and this app is an amazing way to stay motivated!
Thank you for creating it👏🏻🎉

jmanske:
Awesome, looks great. Thanks once again for the contribution!

@matslina .shipit

PR #2367, opened 2024-04-01
Update: MiLB Scores

M0ntyP:
# Description
Changed the names on the league selection dropdown

# Copilot

copilot:all


M0ntyP:
recheck

jmanske:
@matslina any thought to being able to grandfather in apps like this past the 1000 ms rendering limit? If you can show the changes do not effectively increase the rendering time, there is not a good reason I think to deny the PR when it fixes a real bug.

#2346 has the same problem.

M0ntyP:
recheck

matslina:
Hey, sorry for the tremendous delay on this.

Yeah, I think we need something like that. Plenty of apps have acceptable performance, except on the very first run before cache is warmed up.

matslina:
Temporarily bypassed the runtime check for this one.

PR #2375, opened 2024-04-06
Please help I don't know what I'm doing haha. I just wanted to update the code for the "afl" app

henrygg98:
# Description

# Copilot

copilot:all


jmanske:
To be able to contribute, you need to sign the CLA (see that post by the tidbyt bot above).

The build and test check is failing because you need to run `pixlet format` on your code.

However, I think your build and test check will still fail because the app takes too long to render. One of the reasons it takes so long is because it gets the data for every team then filters down to the chosen team instead of only pinging the API for the team that is selected. I can fix it and submit a PR that includes your new schema option, or you're welcome to try your hand at it too!

henrygg98:
**_I have read the CLA Document and I hereby sign the CLA_**

henrygg98:
recheck

PR #2388, opened 2024-04-13
Disabled intermediary color step for multicountdown

BestDistress:
# Description
Disabled intermediary color step for multicountdown.

# Copilot

copilot:all


BestDistress:
:shipit:

PR #2389, opened 2024-04-13
New App: San Diego Trolley

serriere:
# Description
This adds an applet for displaying real-time arrivals for the [San Diego Trolley](https://www.sdmts.com/transit-services/trolley) 🚋.
It can be configured to show arrival times for any two stops in the Trolley system.
Each station usually has two stops, one for each direction.

trolley

# Copilot

copilot:all


serriere:

PR #2409, opened 2024-04-23
New App: Next Event

mattcaruso:
# Description
This new app displays the next date in an iCalendar (ics) calendar when an event will occur. The other calendar options available from the community all focused on daily schedules and would often show "No more events today" (or similar) if there were no events in the connected calendar left on that day. For my use case, I want to display the next day when there's an event coming up, such as a calendar feed of concerts, shows, birthdays, or holidays!

next_event

Thanks to @quesurifn whose Universal iCal app heavily inspired the architecture of this app. 

(PS: The code for the Lambda function referenced in this Pixlet app will also be open sourced soon. I'll submit a new PR when I can link to it directly in the code comments for future creators to see how I'm handling the server-side functionality). 

# Copilot

copilot:all

### DIFF
--- a/apps/nextevent/manifest.yaml
+++ b/apps/nextevent/manifest.yaml
@@ -2,7 +2,7 @@
 id: next-event
 name: Next Event Date
 summary: Displays next event only
-desc: Displays the next calendar event, no matter how far away, from iCal (.ics) files.
+desc: Displays the next calendar event, no matter how far away, from iCal (.ics) files. Note: This app will process your iCal data via an external service outside of the Tidbyt ecosystem.
 author: mattcaruso <Matt Caruso>
 fileName: next_event.star
 packageName: nextevent
--- a/apps/nextevent/manifest.yaml
+++ b/apps/nextevent/manifest.yaml
@@ -1,0 +1,8 @@
+---
+id: next-event
+name: Next Event Date
+summary: Displays next event only
+desc: Displays the next calendar event, no matter how far away, from iCal (.ics) files.
+author: mattcaruso <Matt Caruso>
+fileName: next_event.star
+packageName: nextevent
--- a/apps/nextevent/manifest.yaml
+++ b/apps/nextevent/manifest.yaml
@@ -2,7 +2,7 @@
 id: next-event
 name: Next Event Date
 summary: Displays next event only
-desc: Displays the next calendar event, no matter how far away, from iCal (.ics) files. Note: This app will process your iCal data via an external service outside of the Tidbyt ecosystem.
+desc: Displays the next calendar event, no matter how far away, from iCal (.ics) files. Note; This app will process your iCal data via an external service outside of the Tidbyt ecosystem.
 author: mattcaruso <Matt Caruso>
 fileName: next_event.star
 packageName: nextevent
--- a/apps/nextevent/next_event.star
+++ b/apps/nextevent/next_event.star
@@ -2,6 +2,7 @@ load("encoding/json.star", "json")
 load("http.star", "http")
 load("render.star", "render")
 load("schema.star", "schema")
+load("secret.star", "secret")
 load("time.star", "time")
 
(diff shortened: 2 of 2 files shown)


mattcaruso:

quesurifn:
@mattcaruso I'm glad I inspired this. One quick note I haven't been able to resolve, even though it seems my code works, the way the Tidbyt updates (or doesn't) means it only shows the correct next event once. If you have the same issue, or if you figured out a solution, I'd love to hear! I haven't had the time to look into it.

mattcaruso:
@quesurifn Thanks for calling that out. Because of the way Tidbyt prevents you from "private publishing" outside of manually pushing updates via the API, I haven't been able to QA the app long enough to see that behavior. I'll keep an eye out for it and will think on it if it occurs.

matslina:
@mattcaruso @quesurifn Could you guys elaborate a bit on these Lambda functions? Are they just there to aid in parsing ics?

@quesurifn Could you provide more detail on the issue you mentioned about how "Tidbyt updates (or doesn't)"?

Also, @mattcaruso, private app hosting is available, but requires a subscription.

quesurifn:
@matslina The lambda parses the ICS. Unfortunately for starlink, even though it's implemented in Go, we cannot add our own modules. So for more complex functionality, things like Lambdas are required.

matslina:
> @matslina The lambda parses the ICS. Unfortunately for starlink, even though it's implemented in Go, we cannot add our own modules. So for more complex functionality, things like Lambdas are required.

Pixlet is FOSS, and I think something like an ics parser would be a great addition. Unless there's some technical limitation I'm unaware of, it should be easy to add.

> In terms of updating, the UniversalIcal app has a bug where only the first status is showed. It doesn't update after. I'm not sure if that's due to the latency of the lambda on a cold start or because of the way the tidbyt updates.

Not sure about the lambda, but Tidbyt apps are run each time they're displayed, and they run in an environment that's very similar to what you have locally with Pixlet. There are some differences in caching that might come into play, but not sure how that would happen exactly. I can try to take a look later today.

matslina:
Free and Open Source Software.

matslina:
Just submit a PR! Of course always a good idea to get some input on design and such before investing tons of time into it, just in case there's some issue that prevents us from merging. Discord is a great place for connecting with other community devs, and there's plenty of folks there who have enough experience with pixlet etc to give quality feedback.

mattcaruso:
@matslina Here's the Lambda function code used in my app, related to this PR. 
https://github.com/mattcaruso/tidbyt-nextevent-lambda

apps/nextevent/manifest.yaml
```diff
@@ -0,0 +1,8 @@
+---
+id: next-event
+name: Next Event Date
+summary: Displays next event only
+desc: Displays the next calendar event, no matter how far away, from iCal (.ics) files.
```
matslina, line 5:
```suggestion
desc: Displays the next calendar event, no matter how far away, from iCal (.ics) files. Note: This app will process your iCal data via an external service outside of the Tidbyt ecosystem.
```

matslina:
Took a quick look at iCal, and wow, yeah, adding support to Pixlet would not be entirely trivial. Might need some caching to get acceptable performance, as these ics files can get pretty large. My personal Google Calendar export is almost 1MB, reaching back to 2015.

For now, I'd like to at least add the suggested warning about how the data's processed in the app description. And @quesurifn would you mind doing the same on your app? I want it to be clear to users that their data is being processed by an external service.

apps/nextevent/manifest.yaml
mattcaruso: A colon was breaking yaml syntax and causing the build to fail. I added another commit after this to remove it from the note and replace with a semicolon.

quesurifn:
@matslina I'll update it when I get a chance.

mattcaruso:
@matslina Hello! Any other actions needed to get this one released?

PR #2432, opened 2024-05-01
Added IOB/COB options to display, updated to v2 API

JimAdkins:
Added ability to view Insulin on Board or Carbs on Board in place of Clock. 
Updated to use the v2 API. 
Updated to call properties endpoint for iob/cob/direction/current/delta, and only call entries when graph is shown to help minimize calls.


IsThisPaul:
I have reviewed these updates with Jim and we've both thoroughly tested. This has my blessing.

PR #2433, opened 2024-05-02
Submit new app Custom Quotes

vipulchhajer:
Users can store their favorite quotes in a google sheet and use this app to display them on tidbyt.


vipulchhajer:
Hi @matslina May I bother you for a PR review? Thanks!

vipulchhajer:
Hi @matslina and @rohansingh May I please bother you for a PR review? Thanks!

apps/customquotes/custom_quotes.star
```diff
@@ -0,0 +1,153 @@
...
+def get_image():
+    image = cache.get("image")
+    if not image:
+        response = http.get("https://random.imagecdn.app/500/250")
+        if response.status_code != 200:
+            fail("Failed to retrieve image: %d - %s" % (response.status_code, response.body()))
+        image = base64.encode(response.body())
+
+        cache.set("image", image, ttl_seconds = TTL_SECONDS)
+    return image
```
rohansingh, line 134: Please use `ttl_seconds` in `http.get` instead of using cache directly.

Example at: https://discuss.tidbyt.com/t/http-caching-client/5129

apps/customquotes/manifest.yaml
```diff
@@ -0,0 +1,8 @@
+---
+id: custom-quotes
+name: Custom Quotes
+summary: Display custom quotes
+desc: Display quotes from a GSheet like this https://t.ly/rI5ZT.
```
rohansingh, line 5: This link has "expired" and doesn't work. If possible please use a direct link rather than a shortener since those can stop working.

rohansingh (changes requested):

apps/customquotes/custom_quotes.star
vipulchhajer: Thanks. I made your suggested changes.

apps/customquotes/manifest.yaml
vipulchhajer: Good catch. Thanks, updated it with direct link.

vipulchhajer:
@rohansingh I have made the requested changes based on your feedback. Please take another look when you have a chance.

PR #2437, opened 2024-05-03
Emergency Bug Fix

JimAdkins:
Needed to fail gracefully on all properties.json attributes.


IsThisPaul:
This is a quick bug fix around some potentially missing variable values that Jim has corrected. This is good to go.
The sooner, the better.

JimAdkins:
We are still troubleshooting this. Closing PR for now. Will open when we are sure it's working the way we want. Thanks!

PR #2452, opened 2024-05-07
Nightscout: Urgent Bug Fix

IsThisPaul:
Box spacing to fix broken layout caused by colored border in last version


IsThisPaul:
@matslina Thanks for pushing the new version so quickly. Looks like I had broken a display with some new borders. This PR fixes that broken layout. Appreciate a quick push if possible.

PR #2467, opened 2024-05-16
New app: Noise

rektdeckard:
This PR adds a configurable Simplex Noise generator app, which makes pretty, ambient, animated textures.

noise

### DIFF
--- a/apps/noise/README.md
+++ b/apps/noise/README.md
@@ -1,0 +1,8 @@
+# Noise
+
+Pleasing visuals of colorful, generated Simplex noise.
+
+<img src="./noise.gif" width="640" height="320">
+
+Configurable `Foreground`, `Background`, `Scale`, `Speed`, and an optional `Random Seed` to generate a new pattern.
+
--- a/apps/noise/manifest.yaml
+++ b/apps/noise/manifest.yaml
@@ -1,0 +1,6 @@
+---
+id: noise
+name: Noise
+summary: Animated digital noise
+desc: Set the vibe with beautiful, complex, colorful animated noise produced with math!
+author: rektdeckard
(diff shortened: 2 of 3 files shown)


rektdeckard:

rektdeckard:
Is the 1 second execution limit really necessary? It runs consistently under 1 second on my machine, and 1.4 seconds on this runner. There's a fair amount of math going on here, and I feel the constraints are quite limiting. In order to get this below 1 second on a runner (very difficult to test from a local machine), I will have to either:

1. Make the (already short) animation even shorter. I have already done some tricks to this effect, like only generating half of the animation frames, then mirroring to form a loop that effectively doubles the anim length
2. Somehow optimize a program that is almost entirely unavoidable arithmetic. No matter what, I have to run a per-pixel-per-frame computation.

Wondering if there is any flexibility here.

jmanske:
FWIW, it is consistently over a second to render on my laptop. From the point of view of hundreds and thousands of apps rendering with different settings on the servers constantly, it's more an ecosystem-wide scalability problem than it is a single app problem.

For ideas to make it overall faster: Can you use any caching at all? Can "common" calculations be looked up instead of redone on the fly? I haven't looked at the code yet but I am happy to give a review if you are interested.

rektdeckard:
Thanks @jmanske, there isn't anything cacheable. The generative nature of the app means we WANT it to be different every run, and each pixel of each frame is pseudorandom.

I understand the reason for the limit, I just feel a little short-sold on what we're permitted to do here. I honestly wouldn't mind if local apps had access to schemas and could run code, like maybe on your phone, instead of being a glorified GIF.

jmanske:
Have you looked into private apps at all? It's $5 a month to be able to run whatever code on their server and works more or less like apps in the community repo. I do not think you're beholden to the 1 second rendering limit for those. I use private apps a ton for pre-releasing stuff so at least I can enjoy it before I'm ready to push or can work out kinks like you might be dealing with here. 

Definitely agree that the restrictions are sometimes a bummer.

PR #2508, opened 2024-06-15
[app] adds HASS Graph

vincentezw:
Hi Tidbyt team 👋. This PR adds an app to display historical values of a Home Assistant entity as a line graph. On the right hand side the app will show the highest. lowest and average value:


vincentezw:

PR #2518, opened 2024-06-24
feat: Add Berlin Transit app

flambeauRiverTours:
This commit adds the Berlin Transit app to the repository. The app provides information about upcoming train and bus departures for a given Berlin station. It uses VBB APIs to get real time departure info for Berlin transit. It's modeled off of several other transit-related apps, though primarily tube.star. It displays the station configured to track departures for at the top, and a rotating set of departures from that station below, with each departure containing the line name (e.g. U2, S41), destination of the departure, and the time until departure.

There are several parameters users can configure to change the behavior of the app. First and foremost is the station parameter, where users can look up stations via location services. This parameter . Second is a departure time offset, which can be used to filter out departures occurring in less than a given duration. For example, if you live a five minute walk from the a station, it's not helpful to see a train that's departing in two minutes. The rest of the parameters allow the user to filter to specific types of transport, such as regional trains, trams, busses, etc.

The rendered gif isn't super exciting - I wasn't able to find a way to provide a station when rendering using render, so it's just the default "No station selected" text.
berlin_transit

Here's a couple screenshots of what it looks like in action:
image
image

Some fun coordinates for stations for testing:
52.54947621473808, 13.415633950565859 - schonhauser allee
52.52154406318316, 13.411488625105003 - alexanderplatz
52.52534624541184, 13.369301134070499 - berlin hauptbahnhof

### DIFF
--- a/apps/berlintransit/berlin_transit.star
+++ b/apps/berlintransit/berlin_transit.star
@@ -1,0 +1,635 @@
+"""
+Applet: Berlin Transit
+Summary: Trains/Bus in Berlin
+Description: Provides upcoming train and bus departures for a given Berlin station.
+Author: flambeauRiverTours
+"""
+
+load("encoding/json.star", "json")
+load("http.star", "http")
+load("humanize.star", "humanize")
+load("math.star", "math")
+load("render.star", "render")
+load("schema.star", "schema")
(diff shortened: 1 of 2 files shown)


flambeauRiverTours:

PR #2522, opened 2024-06-25
add Purdue Basketball app

Griffinov22:
This app provides the standing wins and losses for Purdue. It requires an api that can easily be acquired from https://sportsdata.io/ (and it's free!).


apps/purduebasketball/purdue_basketball.star
```diff
@@ -0,0 +1,75 @@
...
+    return {"wins": 0, "losses": 0}
+
+def get_schema():
+    return schema.Schema(
+        version = "1",
+        fields = [
+            schema.Text(
+                id = "api_key",
+                name = "api key",
+                desc = "API key you can get for free at https://sportsdata.io/",
```
matslina, line 71: This description is currently not visible in the Tidbyt app. Would be great if you could add a note about the API key in the app description (in the manifest).

Griffinov22: I have updated the manifest file. If you would like any more changes, please don't hesitate to reach out. Thanks.

PR #2531, opened 2024-07-03
Gartner news fix

rjison:
RSS Feed call was failing. Added UserAgent and Accept headers, and it works again.


rjison:
Thanks. This works locally, but fails in production. Any chance you could tell me what the error is? Is the IP addressed blocked? Is there any response from the http request?ThanksOn Jul 3, 2024, at 5:09 PM, Mats Linander ***@***.***> wrote:﻿
Merged #2531 into main.

matslina:
Not seeing any errors in production for the past 6 hours. Running it locally on my machine triggers the check for `number_of_items == 0`, so it returns `[]`.

rjison:
Thanks for taking a look. The xml should have data, this is what it’s calling:Https://www.gartner.com/en/newsroom/rss You can get 0 items if the cached xml has 0 items, if the xml returned from the http get has zero items or if the http get call fails. Any chance you could tell me which case is happening? I can’t replicate getting zero items since adding some headers to the http call, so I don’t know how to fix this. Thanks again, sorry to be a pain. On Jul 4, 2024, at 9:37 AM, Mats Linander ***@***.***> wrote:﻿
Not seeing any errors in production for the past 6 hours. Running it locally on my machine triggers the check for number_of_items == 0, so it returns [].

PR #2535, opened 2024-07-08
AZ Highway Signs

cjsturgess:
This app allows a user to select any of the AZ Highway Message signs and display it's current message on their Tidbyt!

The AZ Highway Message signs are a collection of electronic roadway message boards to alert drivers to changing traffic info, detours, distance to key landmarks, etc - sometimes, they're even just fun messages!

Examples of real signs:
image
image

This app uses the AZ 511 API (https://www.az511.com/help/endpoint/messagesigns) to pull the current message from any sign, then formats it for display on a Tidbyt. If there are multiple messages (some signs rotate through a few) active for a sign, a random one will be displayed.

image


cjsturgess:

PR #2537, opened 2024-07-08
Congress Watch

rjison:
New App: Congress Watch

This application will display work that the congress is doing. Since there are many bills working through the house and the senate, you can choose to pick just the house, just the senate or both.

Since even in one day there could be dozens of bills working their way through, this app will pick one at random each time it comes up in the rotation. You can decide to pick an item from the past day, week, month or 90 days.


rjison:
If someone can help me figure out what I'm doing wrong with:
pixlet encrypt [app id] [top secret value]

And do we now have a way to read in local png files to avoid base64 encoding them? I saw in the comments on discord we did, but the only example I found in the community wasn't working locally for me.

PR #2539, opened 2024-07-09
goldpriceticker: disable plot, hopefully temporarily

matslina:


matslina:
@aaronbrace76 This app is performing a very large number of cache operations. For now, I'm disabling the time series component. There's a good chance other functionality is broken in the process. Could you take a look and see if the plot can be restored, while keeping down the number of cache operations?

aaronbrace76:
Yes, it looks like the longer periods are receiving more datapoints from the API I wrote than the Tidbyt can even display, several for each day and the more days that elapse obviously it keeps getting bigger. 

I can certainly cut it down by a factor of about 4 relatively easily, is that enough reduction or do I need to dig deeper?

I can also disable caching entirely and just return the full dataset every time. Caching was originally important when I was paying per query, but since I wrote my own API now its not that important.

Just trying to understand the scale of the cache issue.

matslina:
The key thing is to cut down caching to a small (and constant) number of calls. It should be sufficient to cache the HTTP calls (ideally by passing `ttl_seconds` to `http.get()`) and say one `cache.set` for the resulting time series (even that may not be necessary, unless computing it is very expensive).

PR #2545, opened 2024-07-11
Update holidays app to use timezone

czarandy:
I didn't add timezone as an option here, so the time it uses isn't correct. Adding an option now to allow configuring the timezone.


apps/holidays/holidays.star
```diff
@@ -22,7 +29,9 @@ def main(config):
         "Dec 31": CONFETTI,
     }
 
-    now = time.now()
+    location = json.decode(config.get("location") or DEFAULT_LOCATION)
```
matslina, line 32: Another option is to use the (undocumented and magic) `$tz` variable. It'll get passed in config when your app is running in the tidbyt production environment, and holds the device's selected timezone.

czarandy: Nice, that's better

czarandy:
Updated to use $tz variable

PR #2550, opened 2024-07-15
solaredgemonitor: add battery data and use solarmanagerch layout

IngmarStein:
This change adds storage (battery) information and switches the layout to that of the `solarmanagerch` app. This introduces multiple frames to cycle through (main stats, battery, production, consumption, summary, autarky).


IngmarStein:
CC @marcusb as the original owner.

IngmarStein:
@marcusb does the change look good to you?

apps/solaredgemonitor/solaredge_monitor.star
```diff
@@ -5,167 +5,671 @@ Description: Energy production and consumption monitor for your SolarEdge solar
...
+    content = json.decode(rep.body())
+    return render.Root(render.Box(render.WrappedText(str(rep.status_code) + " : " + content["String"], color = RED)))
+
+def get_energy_details(site_id, api_key, tz, interval):
+    now = time.now().in_location(tz)
+    start_string = ""
+    now_string = humanize.time_format("yyyy-MM-dd HH:mm:ss", now)
+    if interval == "day":  # from the start of today to now, just set time to 00:00
+        start_string = humanize.time_format("yyyy-MM-dd 00:00:00", now)
+        now_string = humanize.time_format("2006-01-02 15:04:05", now)
```
marcusb, line 188: test format string?

apps/solaredgemonitor/solaredge_monitor.star
```diff
@@ -5,167 +5,671 @@ Description: Energy production and consumption monitor for your SolarEdge solar
...
+        time_unit = "DAY"
+    elif interval == "24h":  # the last 24hr from now, use a now-duration
+        duration = time.parse_duration("24h")
```
marcusb, line 191: should be some way to get a time constant without parsing

apps/solaredgemonitor/solaredge_monitor.star
```diff
@@ -5,167 +5,671 @@ Description: Energy production and consumption monitor for your SolarEdge solar
...
+        start_time = now - duration
+        start_string = humanize.time_format("yyyy-MM-dd HH:mm:ss", start_time)
+        time_unit = "DAY"
+    elif interval == "month":  # from the start of the month to now, just set day to 01 and time to 00:00
+        month = now.month
+        if now.month < 10:  # pad a zero if less then 10
+            month = "0" + str(month)
+        start_string = "{}-{}-01 00:00:00".format(now.year, month)
```
marcusb, line 199: would `humanize.time_format("yyyy-MM-01 00:00:00", now)` work?

apps/solaredgemonitor/solaredge_monitor.star
```diff
@@ -5,167 +5,671 @@ Description: Energy production and consumption monitor for your SolarEdge solar
...
+            for v in m["values"]:
+                consumption += v["value"]
+    if consumption == 0:
+        return 100
+    return int(self_consumption / consumption * 100)
 
 def main(config):
     api_key = config.str("api_key")
     site_id = humanize.url_encode(config.str("site_id", ""))
+    tz = config.get("$tz", "Europe/Zurich")
```
marcusb, line 245: Seems like a little arbitrary as a default, how about defaulting to the system's time zone? Or `UTC`, if there isn't a way to get the system's time zone.

marcusb (commented):
Nice work! Left a few comments. I wasn't able to test because pixlet segfaults on my machine.

* Please also update `screenshot.png` or add a few screenshots for the README.

apps/solaredgemonitor/solaredge_monitor.star
```diff
@@ -5,167 +5,671 @@ Description: Energy production and consumption monitor for your SolarEdge solar
...
-        else:
-            dir = 0
-        flows.append(dir)
-    if o["LOAD"]:
-        points.append((HOUSE, o["LOAD"]["currentPower"]))
-        if {"from": "GRID", "to": "Load"} in connections:
-            dir = -1
-        elif {"from": "LOAD", "to": "Grid"} in connections:
-            dir = 1
+    if o["STORAGE"]:
```
marcusb, line 293: This crashes on installations without a battery. Use `if "STORAGE" in o` instead (also throughout the file)

apps/solaredgemonitor/solaredge_monitor.star
```diff
@@ -5,167 +5,671 @@ Description: Energy production and consumption monitor for your SolarEdge solar
...
-            el = render.Box(width = 10)
-        dots.append(
-            render.Stack(children = [render.Box(width = 21), el]),
-        )
-    return render.Root(
-        child = render.Stack(
+        ],
+    )
+
+    # CHARGE FRAME shows charge/discharge rate and state of charge percent
```
marcusb, line 481: protect this entire block including charge_frame with `if has_battery` to avoid crash on `o["STORAGE"]`

IngmarStein: Done. I've applied this to all frames so that each is only generated if needed.

apps/solaredgemonitor/solaredge_monitor.star
IngmarStein: Done.

apps/solaredgemonitor/solaredge_monitor.star
IngmarStein: Good idea. I've updated the PR to retrieve the time zone from the site details API and fall back to UTC if it's not available. However, it looks like the time zone is not part of the v2 schema (https://developers.solaredge.com/docs/monitoring/ed297a6bcaeb6-site-details), so that part will likely go away again, when this app is ported to the v2 API.

apps/solaredgemonitor/solaredge_monitor.star
```diff
@@ -19,153 +18,417 @@ URL = "https://monitoringapi.solaredge.com/site/{}/currentPowerFlow"
...
-        dots.append(
-            render.Stack(children = [render.Box(width = 21), el]),
-        )
-    return render.Root(
-        child = render.Stack(
+        ],
+    )
+
+    # CONSUMPTION FRAME
+    verbrauch_frame = render.Box(
```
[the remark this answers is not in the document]
IngmarStein: Done.

apps/solaredgemonitor/solaredge_monitor.star
IngmarStein: It would :)

apps/solaredgemonitor/solaredge_monitor.star
IngmarStein: Done.

apps/solaredgemonitor/solaredge_monitor.star
IngmarStein: Done.

IngmarStein:
Thanks for the feedback, @marcusb! I've applied your suggestions.

marcusb:
> Thanks for the feedback, @marcusb! I've applied your suggestions.

 :+1: Code looks good to me!

I think the new functionality is nice, but honestly I think the visuals are a step back. The solar panel is just a blue box. The grid towers don't look very recognizable. And the animated flow dots look less clear. That's just my preference though, I don't decide things over here :)

IngmarStein:
As you said, it's a matter of taste. I like `solarmanagerch`'s consistent design over the various screens.

I think you are the decider here: the repo owners won't merge a change to an existing app if the original owner doesn't approve.

If you object, we could also offer this variant as a separate applet to give users a choice. When doing so, maybe an upgrade to the v2 API would be a good opportunity because there doesn't seem to be a smooth migration path for apps using the v1 API. However, I don't see where the new API keys can be generated. I registered on http://developers.solaredge.com but didn't get a reply and the SolarEdge support only knows about the v1 keys…

marcusb:
I would prefer to add in the new functionality while keeping the design, so IMO it would be best to split this PR along those lines. Then we can consider changing the visuals separately.

I think the main screen will be the one most people use.

PR #2551, opened 2024-07-16
hassentitylist: render units for entities which have them

IngmarStein:
The app seems to have been designed for number entities which don't have units. When used with sensors, it's useful to also include units of measurement (e.g. "80 %" instead of just "80").


IngmarStein:
CC @jwoglom as the original author.

jwoglom:
@IngmarStein can we put it behind a preference?

IngmarStein:
@jwoglom Sure! Done.

jwoglom:
Don't have review permissions in this repo, but LGTM

PR #2561, opened 2024-07-22
hass-solar: add an option to hide the charging frame and a German translation

IngmarStein:
The option adds the ability to create instances of the app showing only the main frame or the charging frame (with animations).


IngmarStein:
Friendly ping @matslina

PR #2569, opened 2024-07-30
Update: Olympic Medals: add option to track a specific country

danielsitnik:
This update adds a second display option to the app where the user can track the medals of a specific country.

Preview:
olympic_medals


tim-hanssen:
We need this 🥇

PR #2588, opened 2024-08-13
CSV Viewer: a few small config updates

mokolodi1:
New features: 
- Option to allow letters to touch in adjacent cells
- Option to expand text right when possible if cells are blank
- Allow text to be slightly taller in certain circumstances


mokolodi1:
Welp, apparently these were merged in already!

PR #2600, opened 2024-08-23
Introduce Czechnews applet

solarisle:
This applet can show Czech news headline from seznamzpravy.cz or irozhlas.cz.


solarisle:

PR #2601, opened 2024-08-24
FPL Tracker:  Bug Fixes

DoubleGremlin181:
Fixed bugs with the previous week rank logic
1. Corrected rank change direction being reversed
2. Corrected prev_rank being a list of 1 value and not an int


DoubleGremlin181:
Thanks for the quick merge! ❤️

PR #2615, opened 2024-09-04
set max_age in tetris, life and fuzzy clock

matslina:


apps/fuzzyclock/fuzzy_clock.star
```diff
@@ -315,6 +315,7 @@ def main(config):
     texts = [render.Text(" " * i + s) for i, s in enumerate(fuzzed)]
 
     return render.Root(
+        max_age = 60*10,
```
matslina, line 318:
```suggestion
        max_age = 60 * 10,
```

PR #2639, opened 2024-09-10
New App: Snake Clock

paxos:
snake_clock

Adds a new app:
- shows a clock in the foreground
- background shows a snake moving around in a random pattern
- using global state to sync the movement for all users
- this means every user will see the same snake movements 🎉

### DIFF
--- a/apps/snakeclock/manifest.yaml
+++ b/apps/snakeclock/manifest.yaml
@@ -1,0 +1,6 @@
+---
+id: 62b55d8d-2741-4854-ae29-0cdcd6660bc8
+name: Snake Clock
+summary: Shows the time and a snake
+desc: Shows the time with a snake slithering around in the background.
+author: paxos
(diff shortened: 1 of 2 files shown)


paxos:

drudge:
This looks great! Would you mind adding support for 12-hour format?

Here is an example from another community app: 
https://github.com/tidbyt/community/blob/main/apps/underwaterclock/underwater_clock.star#L165-L170
https://github.com/tidbyt/community/blob/main/apps/underwaterclock/underwater_clock.star#L78-L83

paxos:
> This looks great! Would you mind adding support for 12-hour format?
> 
> Here is an example from another community app: https://github.com/tidbyt/community/blob/main/apps/underwaterclock/underwater_clock.star#L165-L170 https://github.com/tidbyt/community/blob/main/apps/underwaterclock/underwater_clock.star#L78-L83

For sure :) -  thanks for checking!

paxos:
12 / 24 hour setting added
Also made sure to show time for the correct timezone 🤦‍♂️

PR #2641, opened 2024-09-11
New App: Stock Spotlight

sethcottle:
### Description

The Stock Spotlight applet allows you to add up to 3 of your favorite stocks using [Finnhub.io](https://finnhub.io) for stock data. Users will need to use their own Finnhub API key for Stock Spotlight to work. I'm hopeful this implementation compared to some of the other stock apps on the Tidbyt right now can help provide a bit more "real-time" trading data—data is able to refresh every 60 seconds while staying within Finnhub free tier API limits (users should be mindful of having multiple instances of this applet to review that API rate limit).

### Screenshot
stock-spotlight


apps/stockspotlight/stock-spotlight.star
```diff
@@ -0,0 +1,197 @@
...
+            ),
+        ),
+    )
+
+def get_stock_data(api_key, stocks):
+    all_stock_data = []
+    for symbol in stocks:
+        if len(symbol) < 1 or len(symbol) > 5:  # Basic validation for stock symbol length
+            continue
+        cached_data = cache.get("stock_data_" + symbol)
```
matslina, line 68: You should be able to simplify caching a bit by relying on the http module's built-in cache functionality. Just pass `ttl_seconds` to `http.get()`. Would be great to see that change before merging!

sethcottle:
Thanks @matslina, I'll work on that this evening!

sethcottle:
@matslina didn't get to this until late this evening, but removed `load("cache.star", "cache")`, added `ttl_seconds` to the `http.get()` call in `fetch_stock_data` and removed the manual caching logic from `get_stock_data`.

matslina (approved):

PR #2657, opened 2024-09-13
Fixed Conferences

Greenieb32:
Removed Divisions from the Big Ten, SEC, and MAC to meet the 2024 Setup

### DIFF
--- a/apps/ncaafstandings/ncaaf_standings.star
+++ b/apps/ncaafstandings/ncaaf_standings.star
@@ -220,24 +220,16 @@ conferenceOptions = [
         value = "4",
     ),
     schema.Option(
-        display = "Big Ten - East",
-        value = "5&0",
-    ),
-    schema.Option(
-        display = "Big Ten - West",
-        value = "5&1",
+        display = "Big Ten",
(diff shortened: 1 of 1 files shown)


Greenieb32:

cmarkham20:
> @cmarkham20 PTAL?

Looks good to me.

matslina:
I see that this changes a couple of config values. Existing installations of the app will continue to pass the old values, until they get removed and reinstalled. Is that going to be an issue? If so, would it be possible to add a translation from old to new values to keep existing installations working?

Greenieb32:
I can change it so the old division selections (Big Ten - East) will just bring up the whole conference standings, but those selections aren't working right now. See image

Greenieb32:
IMG_2889

matslina:
If you could add a bit of code just to translate e.g. `8&1` and `8&0` to `8`, then I think that might give a decent experience for folks who already have the app installed. Correct me if I'm wrong!

Greenieb32 (commented):
Added Translations for Changed Values

PR #2662, opened 2024-09-13
splatoon3schedule: support Tricolor Turf War for Final Fest

Denton-L:
The data format has been changed for Final Fest. Update the app to handle this new format.


Denton-L:
splatoon3schedule

PR #2675, opened 2024-09-16
New app: Stonks!

harrisonpage:
Stock market app that uses Polygon.io API keys to fetch quotes.


apps/stonks/stonks.star
```diff
@@ -0,0 +1,239 @@
...
+    str_amount = str(amount).split(".")
+    dollars = str_amount[0]
+    cents = str_amount[1] if len(str_amount) > 1 else "00"
+    cents = (cents + "0")[:2]
+    return "${}.{}".format(dollars, cents)
+
+def get_stock_quote(url, symbol):
+    """Fetch quote from polygon.io or cache"""
+    cache_key = "stonks-{0}-{1}".format(symbol, VERSION)
+    cached_data = cache.get(cache_key)
```
matslina, line 82: You can replace these calls to the cache module by passing `ttl_seconds` directly to `http.get()`.

DanDobrick: Does that cache across all instances of the app like (I think) this cache does? I could see it nice to cache stock data across all installations since the data is universal.

matslina: Ah, I see what you're saying. Clever!

PR #2680, opened 2024-09-17
Fix enphase solar

### LINKED ISSUE #2661: [Enphase Solar] Enlighten API v4 sets energy_today in systems request to -1
# Summary

The Enphase Solar app currently always displays "**- 0.00 kWh generated**". The issue seemed to be experienced by [other users as well](https://discuss.tidbyt.com/t/enphase-solar-app/5947/2).

# Details

The application makes a request to the [Enphase API V4](https://developer-v4.enphase.com/docs.html) using the `/api/v4/systems` call, which returns a `energy_today` parameter for each system. The problem is, the `energy_today` parameter seems to always return -1.

What probably should happen is that the `system_id` parameter from the `/api/v4/systems` call should be collected, then a subsequent call to `/api/v4/systems/{system_id}/summary` should be made, where the `energy_today` parameter is set to the correct value.

pyropenguin:
Closes #2661

Adds Enphase API request to `/api/v4/systems/{system_id}/summary` to get `energy_today` after Enphase stopped putting `energy_today` in the original `/api/v4/systems` request.


pyropenguin:

matslina:
@laodaochong PTAL

laodaochong:
All checks are green now. So I do have permission to approve 😄

pyropenguin:
Sorry to derail the PR - I just noticed that the data isn't updating once per hour like I thought. I think maybe the cache timeout isn't working right. I'll work on a fix and put it back out to review shortly.

pyropenguin:
Nevermind -- I was seeing that the version I pushed to my Tidbyt was not updating. I didn't realize that pushing to my Tidbyt via the pixlet tool would display a static, non-updating image of my app. The PR is ready.

PR #2686, opened 2024-09-18
New App: Days Since

dtsch:
I saw the Accident Free Days app when setting up my new Tidbyt Gen 2 and thought that it would be more versatile if the 'accident' could be custom text too.

It also dawned on me that it would probably be an easy enough first Tidbyt app, something I've wanted to try my hand at ever since receiving my Gen 1 a few years ago.

So this app lets the user set a custom date and custom text, so they can let the world know how many days have elapsed since the thing happened.

days_since


dtsch:

PR #2696, opened 2024-09-19
New App: Homefiniti

luggage66:
This app shows live visitor stats for certain websites hosted by our company for our clients (homebuilders in the US). The user will get an easy-to-type API key from their admin dashboard. Shown below is sample output with and without an API key.

image

image


luggage66:

PR #2710, opened 2024-09-21
App hassgraph: show lowest/highest in local time

ChaosKid42:
App HASS Graph:

This PR makes the App show timestamps in local time instead of UTC

@vincentezw : Hope you like the idea.

### DIFF
--- a/apps/hassgraph/hass_graph.star
+++ b/apps/hassgraph/hass_graph.star
@@ -1,5 +1,6 @@
 load("animation.star", "animation")
 load("encoding/base64.star", "base64")
+load("encoding/json.star", "json")
 load("http.star", "http")
 load("render.star", "render")
 load("schema.star", "schema")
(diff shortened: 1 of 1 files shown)


ChaosKid42:
> Great idea, and thank you for the PR! Being in my timezone, I hadn't considered this issue before. The PR looks good overall. However, I do wonder if we might skip the in_location call when no timezone has been configured, rather than defaulting to New York. This change could help avoid unexpected results for existing users who haven't set a timezone. That's just a suggestion, though!

Yes. Your are right. Better to not change the app's beaheviour in case the location isn't set. I changed the PR accordingly.

What do you think?

vincentezw:
> @vincentezw: what do you think?

Awesome, thanks for adjusting that ❤️ . LGTM (we need to wait for Tidbyt to approve and merge though)

PR #2720, opened 2024-09-23
New App: Guild Wars 2 Fractals

ahlec:
This adds a new app to preview the [daily Fractal of the Mists](https://wiki.guildwars2.com/wiki/Fractals_of_the_Mists#Daily_Fractals) from the game Guild Wars 2.

gw2_fractals

### Brief overview (if necessary)
There are a collection of different dungeons in Guild Wars 2 called "fractals." Difficulty (known as "scale") ranges from 1 to 100, and is broken up into four difficulty tiers (Tier 1 = scale 1-25, Tier 2 = scale 26-50, etc). A specific scale is always the same dungeon (eg, scale 1 is always the "Volcanic" fractal). Within a specific tier, some fractals only show up once, and others show up twice (eg, in Tier 4, Swampland is both scale 77 and 89).

Every day, 3 fractals are the "daily fractals" and have increased rewards. You can complete this at any scale that features this fractal (eg, "Daily Tier 4 Swampland" can be earned by either 77 and 89). Endgame players will run fractals every day within the T4 difficulty tier; earlier tiers are typically seen as being gradual training to reach tier 4.

In addition to this, there are 3 fractals chosen as "recommended fractals" every day; these are specific scales, meaning you must specifically complete scale 19 (Volcanic) and not a different scale that features Volcanic. There is one "rec" from Tier 1, Tier 2, and Tier 3 (never Tier 4).

In short: You have 6 daily fractals. Three of them are associated with a specific scale, and three are not.

Additional to all of this (b/c this wasn't confusing enough), a subset of fractals have "challenge modes" ("cm" for short) which further increase difficulty and rewards. You can only activate challenge mode in Tier 4. Some players (but not all) will opt to do a fractal on challenge mode if it's part of that day's dailies.

There are a number of different playstyles for players:
* Daily T4s (w/ CMs if they show up in the rotation) + Recs
* Daily T4s (normal mode only) + Recs
* Daily T4s only
* Recs only
* The fixed set of all fractals that have challenge modes (unchanging day to day) + Daily T4s

### Implementation

Fractals are on a fixed 15-day cycle that repeats once completed. I've ported the implementation from the Guild Wars 2 Wiki for how this cycle is computed, and I've spot checked the implementation over the past two weeks.

The only time this data changes is when they release a new fractal, which right now happens once a year. New releases are known in advance. When released, which fractal is found at what scale is shuffled to make room. However, this does not affect the core "it's on a fixed 15-day cycle" logic.

I've attempted to support tailoring the app to match the different playstyles through config as best as possible.

### DIFF
--- a/apps/gw2_fractals/gw2_fractals.star
+++ b/apps/gw2_fractals/gw2_fractals.star
@@ -1,0 +1,412 @@
+load("animation.star", "animation")
+load("encoding/base64.star", "base64")
+load("math.star", "math")
+load("render.star", "render")
+load("schema.star", "schema")
+load("time.star", "time")
+
+DAILY_FOTM_ICON = base64.decode("""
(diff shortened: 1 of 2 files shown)


ahlec:

ahlec:
recheck

PR #2721, opened 2024-09-23
New app: Life Counter

danceforheaven:
An app that counts the months elapsed and remaining in your life. From the readme:

# Life Counter
> How long do you have left?

If you don't feel overwhelmed enough by your own mortality, download this app - and be reminded every day that your time is finite and depleting.

## What it does

Enter your date of birth in the app settings, and a grid of dots representing the months in your life will be displayed.

Each dot indicates one month, and each column of 24 represents two years. A red dot indicates a month that has already passed, a white dot is one that has not. A flashing dot marks the current month - it will flash faster as the month approaches its end.

The default number of years shown is 100. This can be changed, preferably to the average life expectancy of whichever demographic you happen to be in.


danceforheaven:

PR #2723, opened 2024-09-24
Update weather_snapshot.star to allow use of personal api key

elijahyoon:
Added personal API Key schema (config) value to allow more frequent updates (every 10 minutes vs every 6 hrs). Changed weather API to free tier that does not require a credit card. Changed AQI fetch url to use "current" endpoint vs "history" endpoint. Original functionality is maintained if no personal API key is provided (will still use the provided hashed API keys in file and cache/query every 6 hrs).

### DIFF
--- a/apps/weathersnapshot/weather_snapshot.star
+++ b/apps/weathersnapshot/weather_snapshot.star
@@ -1,11 +1,9 @@
 load("encoding/base64.star", "base64")
 load("encoding/json.star", "json")
 load("http.star", "http")
-load("math.star", "math")
 load("render.star", "render")
 load("schema.star", "schema")
 load("secret.star", "secret")
-load("time.star", "time")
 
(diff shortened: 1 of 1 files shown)


elijahyoon:

elijahyoon:
hey @arieltian would love to get your thoughts on these changes!

arieltian:
Hey, thanks for making these changes :)

Question about your changes to the AQI api: are you sure it’s safe to take out start and end from the query? iirc that was important both for the AQI calculation (which takes a weighted avg of pollution values over the last 24h), and also rounding down to the nearest cache_ttl_sec naturally kept the query consistent for 6h periods for the caching to work.

Lmk if any of that doesn’t make sense

elijahyoon:
Need to recreate this PR.

PR #2725, opened 2024-09-24
New App: PiHole Stats

siva801:
This app displays basic ad blocking statistics from PiHole instance. Config options include host name/ip and API Key.
Here is the rendered gif
pihole_screenshot
Here is the render before adding config
pihole_noconfig_screenshot


siva801:

siva801:
@DanDobrick pihole needs to be exposed using reverse proxy for this app to work. I tested with my pihole exposed via reverse proxy and it works.

IngmarStein:
Maybe also document that this is for Pi-hole v5 and won't work with the v6 beta.

PR #2726, opened 2024-09-24
Sleeperyearfix

ddebo531:
update to try and get the current nfl season so it works for every year and not just 2023


ddebo531:

matslina:
@jweier Thoughts?

jweier:
Looks good to me and would hopefully fix it for 2024.

PR #2731, opened 2024-09-25
feat: Support picking a team in ESPN FF

matthewfischer:
This lets you pick your own team (or a team you are interested in) and only show that score.


matthewfischer:
@jack-markle  hope you don't mind me adding this. I scored 151 this week and need to see my own glorious score only. I also do NOT know how to test the new setting, it may be broken. I can fix if so.

jack-markle:
> @jack-markle hope you don't mind me adding this. I scored 151 this week and need to see my own glorious score only. I also do NOT know how to test the new setting, it may be broken. I can fix if so.
> 
> Also no idea on that build issue!

Don't mind at all, this is awesome. I'll see if I can get around the build issue!

PR #2732, opened 2024-09-25
app that displays the stats of a Fenecon solar system

raven-rwho:
fenecon

This small app displays three values that are delivered from a Fenecon powered solar system.

On the left is a pie chart for the battery status. On the right is the consumption and the generated amount of energy.

Somehow the cli generated the fenecon.star file not under the apps/fenecon folder but instead on the root of the directory, which looks different in the documentation - should I change the location and make that directory myself?

Thanks!


raven-rwho:

matslina:
@raven-rwho Thanks for the PR. When CLA signing fails, it's usually due to commits being associated with an email address that's not on the Github account used to cut the PR. If that's the case here, opening a separate PR with the issue addressed might be the simplest path to a green build.

raven-rwho:
@matslina - I'm still wondering if it is correct that my code is not under the apps folder but in the root folder - as the cli generated it - is this correct?

Checks are all green now

matslina:
Oh, good catch. Yes, the app needs to go into a separate directory under `apps/`, just like all the others.

raven-rwho:
@matslina - moved the code - now I have the problem that the endpoint can only be reached locally (192.168.x.x) and I get this error message - ca I somehow check if the config is non null to render a default message?

raven-rwho:
Closed, because internal sources can't be reached - would be great to update the docs accordingly

PR #2741, opened 2024-09-26
First pass at Rachio app

matthewfischer:
This will show events from a Rachio sprinkler controller system.

rachio


matthewfischer:
Making a few changes and will reopen

PR #2750, opened 2024-09-28
playingoncapital app added

ibanyan:


ibanyan (commented):
It looks like my local fork was a little old which seemed to have caused the initial errors, however the new app is very simple and seems to have passed all tests. 

Just wondering when it will appear in the catalog to use.

.gitignore
```diff
@@ -24,5 +24,4 @@ BUILD.bazel
 
 # Pixlet
 *.webp
-*.gif
-*.tar.gz
\ No newline at end of file
+*.tar.gz
```
matslina, line 27:
```suggestion
*.gif
*.tar.gz
```

.gitignore
```diff
@@ -25,4 +25,4 @@ BUILD.bazel
 # Pixlet
 *.webp
 *.gif
-*.tar.gz
\ No newline at end of file
+*.tar.gz
```
matslina, line 28:
```suggestion
*.tar.gz```

PR #2751, opened 2024-09-28
Rachio Application

matthewfischer:
An app to show events from a Rachio sprinkler controller. Focused on showing when schedules begin or end and additionally showing weather related skips.


matthewfischer:
rachio

PR #2758, opened 2024-09-30
Update Air Quality app to add widget support and configurable display type

danielsitnik:
Updates the Air Quality app to add widget support (disables animation) and introduces a configurable display type.
The user can now select between the original forecast view and the new pollutants view where it's possible to see the value of a specific pollutant (eg: pm2.5).

Preview:
image


danielsitnik:
Build is failing because of the secrets module.
If you need an API key to validate locally I can provide one.

danielsitnik:
Fixed handling of API Keys so it does not break the CI anymore.

PR #2766, opened 2024-10-03
Use device $tz and convert epoch to UTC

### LINKED ISSUE #2663: [duolingo] Duolingo app only uses default timezone (London)
I just started using the app this week. I've noticed that it was saying I hadn't done any lesson that day when I had. I read your note that it should refresh every 20 minutes but that didn't seem to be it. I then changed to weekly view and it was flashing on Saturday when it is currently 6:30 PM on Friday evening for me (USA central daylight time).

I have my location set in my Tidbyt app preferences but apparently the Duolingo app isn't using that or unable to read that?

I see [where it _should_ be reading it and defaulting to the `DEFAULT_TIMEZONE`](https://github.com/tidbyt/community/blob/main/apps/duolingo/duolingo.star#L322) of "[Europe/London](https://github.com/tidbyt/community/blob/main/apps/duolingo/duolingo.star#L22)".

Just wondering if it's the app or maybe something else I didn't set. I'm a new Tidbyt user. This is on a Gen2, app is on Android.

dtseiler:
- Use device $tz if available to properly display things.
- Keep epoch time in UTC since Duolingo seems to specify midnight UTC of each day in the summary json.

Fixes #2663 (hopefully)

### DIFF
--- a/apps/duolingo/duolingo.star
+++ b/apps/duolingo/duolingo.star
@@ -319,7 +319,8 @@ def main(config):
     duolingo_streak_daystart_cached = cache.get(duolingo_cache_key_streak_daystart)
 
     # Get time and location variables
-    timezone = config.get("timezone", DEFAULT_TIMEZONE)
+    timezone = config.get("$tz", DEFAULT_TIMEZONE)
+    print("Using timezone " + timezone)
 
     # DEBUG
     # print("DEBUG WARNING: Duolingo Username is set to saltedlolly manually.")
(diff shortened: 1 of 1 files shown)


dtseiler:

dtseiler:
@matslina thank you for merging! I'm just curious now what is the process or ETA for the new version to be "deployed" to my device?

matslina:
@dtseiler It should already be live. Is it not?

dtseiler:
> @dtseiler It should already be live. Is it not?

Ah, not sure. The fix here doesn't come into play until later in the evening (when it's the next day in London or UTC). I'll check later. Thanks!

PR #2767, opened 2024-10-04
BGG last play update: display game cover art

DanDobrick:
This is what it looks like without the labels
bgg_last_play

"days since" label
bgg_last_play

"last play" label
bgg_last_play


DanDobrick:
@dtsch FYI I went ahead and update this app today

PR #2775, opened 2024-10-07
Update of Synth Solar SunDial app

maxalines:
We have made a big update to our Synth Solar SunDial app which we have tested and now want to role out to all of our customers.

You can see details on what it looks like here: https://synth.solar/introduction-to-the-new-synth-sundial/


maxalines:

maxalines:
Hi @matslina - I think we have resolved all conflicts here and should be good to go. There is a check that mentions an issue with rendering the new "6 x 10 rounded" font but I think this might be because the check is using an old runtime check version?

It be great to merge this soon as we have told our customers that this update is coming to their Tidbyts. Thanks!

maxalines:
Hi @matslina - I just wanted to see if any issues with merging this PR? We're keen to let our customers know that they're going to get an updated app and so be good to know if there is anything preventing this from being merged. 

Thank you!

PR #2792, opened 2024-10-10
Get random numbers across paths and revert to using Marquee

MichaelYagi:
Use caching to keep number static across paths


tobyxdd:

matslina:
Hey. Sorry for the late review.

Would you mind giving some more background on what you're trying to do with the cache module here? If all you need is a random number that stays the same over some time window, it might be easier to seed the RNG with current time in seconds, divided by some constant. Something along the lines of `random.seed(time.now().unix // 15)`. Generally prefer to keep usage of the cache module down as much as possible, as it's been very common for devs to introduce bugs when using it. Especially so when the cached data is potentially shared across multiple installations of the app.

MichaelYagi:
> Hey. Sorry for the late review.
> 
> Would you mind giving some more background on what you're trying to do with the cache module here? If all you need is a random number that stays the same over some time window, it might be easier to seed the RNG with current time in seconds, divided by some constant. Something along the lines of `random.seed(time.now().unix // 15)`. Generally prefer to keep usage of the cache module down as much as possible, as it's been very common for devs to introduce bugs when using it. Especially so when the cached data is potentially shared across multiple installations of the app.

Yes, that's the intent. But in addition to that, I'm tracking 10 different random numbers based on the label (could be less, maybe keep it to 5?). Right now I have 10 buckets, rand0 to rand9. I'm not sure how I would be able to implement that based on the RNG seed. "cached data is potentially shared across multiple installations of the app", is that on the same device?

matslina:
> Yes, that's the intent. But in addition to that, I'm tracking 10 different random numbers based on a label to be able to handle nested indexes. Right now I have 10 buckets, rand0 to rand9. I'm not sure how I would be able to implement that based on the RNG seed. "cached data is potentially shared across multiple installations of the app", is that on the same device? How is the TTL affected with the shared cache data? I'm not overly concerned since it helps keep random numbers consistent across paths and there are multiple labels.

The cache is scoped to your app, but not to devices or installations. I.e. doing `cache.set("foo", "bar")` in this app won't affect other apps reading `"foo"`, but it absolutely will affect other users of this app (i.e. their "installations" of the app.)

Not sure what bucket means in this context, but if you just need 10 random numbers that are static over a time period, you can do the seeding I suggested above, and then just call `random.number()` 10 times. With the same seed, the 10 numbers will be the same across the time window.

MichaelYagi:
> > ... but if you just need 10 random numbers that are static over a time period, you can do the seeding I suggested above, and then just call `random.number()` 10 times. With the same seed, the 10 numbers will be the same across the time window.

Is there a way to control the time window using that? eg 5s, 20s, etc?

matslina:
> Is there a way to control the time window using that? eg 5s, 20s, etc?

Yeah, so `time.now().unix` gives you current time since epoch in seconds. If you do integer division of that with e.g. 10, i.e. `time.now().unix // 10` you get a number that'll increment every 10 seconds, but be the same within that 10s window. Pass that into `random.seed()`, and the RNG will produce the same sequence of number on each run.

However, if you need different time windows per bucket, then you're out of luck. Can't immediately think of a way of doing that with the random module. Caching might work.

MichaelYagi:
> However, if you need different time windows per bucket, then you're out of luck. Can't immediately think of a way of doing that with the random module. Caching might work.

Yes, that is what I'd like to do. Can I leave the code as is? (I learned something new with how seeding works!)

matslina:
let's try it

PR #2801, opened 2024-10-14
Update usgs_earthquakes.star

BillyMcSkintos:
Added start_time param affected by lookback config item to allow users to configure how long events will display for.


matslina:
@csilv Thoughts?

apps/usgsearthquakes/usgs_earthquakes.star
```diff
@@ -60,7 +69,8 @@ DEFAULT_LOCATION = """
 }
 """
 DEFAULT_MAGNITUDE = "3"
-DEFAULT_RADIUS = "0"
+DEFAULT_RADIUS = "250"
```
csilv, line 72: My preference would be to keep the default radius at zero. While 250 makes sense for the area I live in, for many people this results in nothing to show for most of the time.

apps/usgsearthquakes/usgs_earthquakes.star
```diff
@@ -239,10 +255,11 @@ def fetch_earthquakes(lat, lng, radius, magnitude):
...
     features = resp.json().get("features")
-    if not features:
-        # buildifier: disable=print
-        print("missing features: %s" & resp.body())
-        return None
+
+    #2024-09-27# if not features:
+    # buildifier: disable=print
+    #2024-09-27#    print("missing features: %s" + resp.body())
+    #2024-09-27#    return None
```
csilv, line 262: suggest just deleting these lines if we don't need them.

csilv (commented):
Thank you for putting together this PR. Couple nits, but let me know if you disagree with either.

BillyMcSkintos:
Candidly this is my first PR (anywhere) and I am unfamiliar with
appropriate dev. procedures and protocol when affecting another's work.

csilv:
> Candidly this is my first PR (anywhere) and I am unfamiliar with appropriate dev. procedures and protocol when affecting another's work.

You're all good. This is one of my first code reviews on GitHub so we'll muddle through this together. ;-)

apps/usgsearthquakes/usgs_earthquakes.star
BillyMcSkintos: Will update and commit.

apps/usgsearthquakes/usgs_earthquakes.star
BillyMcSkintos: Yep

BillyMcSkintos:
@csilv Updated per your comments.

csilv:
Looks good to me. Thanks @BillyMcSkintos

BillyMcSkintos:
@matslina We're good now.

PR #2803, opened 2024-10-14
SolarEdge - Overview

markazali:
works with v1 API. Uses https://[monitoringapi.solaredge.com/site/[site]/overview?api_key=[apikey]

My site/api key did not work with the other apps currently available


markazali:

BillyMcSkintos:
@markazali How did you get the PR to work without storing your own API key and SiteID?

markazali:
> @markazali How did you get the PR to work without storing your own API key and SiteID?

I had a ton of help on the source. I believe it was a matter of this checking if the values are stored yet and using dummy data if now

   #before we go get the data - let's make sure we have an API key & Site ID
    if (config.get("apikey", "") == "" or config.get("siteid", "") == ""):
        # we don't have a key - so we're going to use the demo data
        json_data = V1DEMO_DATA
        status_code = 200
        message = "DEMO DATA"

PR #2805, opened 2024-10-15
remove next_codeforces

matslina:


matslina:
@vzsky Getting 403s from the API in CI, preventing other merged apps from getting released. Removing the app for now to unblock CI.

I got 403s locally as well with `pixlet render`. When I loaded the URL in a browser I got hit with a CAPTCHA. Guessing that's what's causing the issue.

vzsky:
I was logged in to the site while doing the dev so they didn't ask for my captcha. I tried with a fresh browser and it did. This is very strange. Why do you need a captcha for an API?

Anyway, let's just leave it at this point. I'm contacting the site but it could take ages. Thanks.

matslina:
They're using cloudflare. Should be a setting in there to disable recaptcha for `/api/`. FWIW.

PR #2807, opened 2024-10-16
fix: [App update] Cricket scores app functional now with CricBuzz source

adilansari:
- Changed the API source from Cricinfo to Cricbuzz as former now requires authentication. The app was non-functional because of auth requirement, it is back working again
- Use global cache for caching API results, there is no user specific data used in the API calls, just generic match schedules and scores


adilansari:
Hi @matslina the app is broken right now and is fixed by this change.

matslina:
Thanks for the fix.

Would be great if the app didn't rely on the `cache.star` module, but perhaps unavoidable here.

PR #2810, opened 2024-10-17
fix: Cricket app bug addressed when a player entry is missing

adilansari:


adilansari:
@matslina  apologies for bugging you again.

PR #2811, opened 2024-10-17
New app: OctoWatts 

TomForeman86:


TomForeman86:

apps/OctoWatts/octowatts.star
```diff
@@ -0,0 +1,480 @@
...
+    else:
+        # Not enough entries; set to current demand
+        demand_average = latest_telemetry["demand"]
+
+    demand_value = int(float(latest_telemetry.get("demand", 0)))
+    return demand_value, demand_average
+
+def fetch_telemetry(access_token, device_id, ttl):
+    # Check cache so we don't make requests more than every 10 seconds
+    cached_telemetry = cache.get("telemetry".format(device_id))
```
matslina, line 152: Wouldn't this just evaluate to "telemetry"? No `{}`in that string. That in turn means the same cache key is used for all installations of the app.

TomForeman86: Your right, I've pushed the update code.

PR #2818, opened 2024-10-20
Rachio Update

rjison:
New icon without white background, skip when nothing to report option, support device selection if you own multiple rachios, decide to show past and/or future events. Round timestamps to nearest 3 minutes -- this will allow the http cache, which has timestamps in the URLs, to leverage caching. Set to display full scroll, and option to adjust scroll speed.

@Matt-Pesce  I made the changes we discussed, also went ahead and supported multiple devices (Should work, even though I only have a single device.. how about you?)


matslina:
@matthewfischer Cool to merge?

matthewfischer:
Yes this is much cleaner!  Thanks

matthewfischer:
Also thanks @rjison

PR #2819, opened 2024-10-20
[New App]: MotoGP schedule for upcoming races

adilansari:
# MotoGP standings and schedules on your Tidbyt
- App lets you display the next event time in your local timezone. By default Free Practice, Qualifying, Sprint and Race schedule is included.
- User can optionally disable schedule screen for Practice and Qualifying sessions. Sprint and Races will always be displayed
- Also displays Top 4 riders in current championship season.

### Example: Upcoming racing event in your local time
schedule

### Example: Top 4 riders in current championship with points
standings


apps/motogp/motogp.star
```diff
@@ -0,0 +1,247 @@
...
+    for category in categories:
+        if "motogp" in category["name"].lower():
+            category_id = category["id"]
+            break
+    url = STANDINGS_URL.format(season_id = season_id, category_id = category_id)
+    data = get_cachable_data(url, 1 * ONE_HOUR)
+    return data
+
+def get_cachable_data(url, timeout):
+    cached_data = cache.get(url)
```
matslina, line 216: Would be great if you could switch this over to use the http module's built in caching instead? Just pass `ttl_seconds` to `http.get()`.

adilansari: Done

PR #2824, opened 2024-10-23
feat: candystatus app!

nbohling:
A simple animated image and marquee app that you can use to tell prospective trick-or-treaters whether you still have candy. If your neighborhood is like mine, you run out before the trick-or-treaters stop. This sign in your window can help stop the knocking when you've run out.

candystatus


nbohling:

PR #2828, opened 2024-10-26
splatoon3schedule: fix Tricolor Turf War again

Denton-L:
Turns out they updated the data model for Tricolor Turf War yet again. Update the app to handle this but, if `.data.currentFest.timetable` is ever not null again, make use of that instead. (Hopefully its data model hasn't changed as well.)


Denton-L:
splatoon3schedule

PR #2830, opened 2024-10-26
Rachio updates

rjison:
Fixed method to display the current zone that is watering at the moment.  Found way to know if there is a 'current' job running, or instead, a recent job to display. 

Allow user to display the device name on Tidbyt to help distunguish multiple rachio units.

### DIFF
--- a/apps/rachio/rachio.star
+++ b/apps/rachio/rachio.star
@@ -89,11 +89,7 @@ def main(config):
 
     # we have a selected device; otherwise, they've already been sent to the display_error_screen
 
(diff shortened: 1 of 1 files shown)


rjison:
@matthewfischer What I submitted last time didn't quite work with regards to 'next' scheduled display. So this addresses that mistake. Instead of trying to calculate the next scheduled event, I'll display the 'last' schedule, or the current schedule and the current zone.

matthewfischer (commented):
My poor casing all gone now ;)

PR #2833, opened 2024-10-27
Remove version pin in devcontainer

RobRoseKnows:
The version pin in the devcontainer means that `pixlet create` and `pixlet check` are unavailable. I removed the pinned version so that by default we install the latest pixlet.


RobRoseKnows:

PR #2843, opened 2024-10-30
ctaltracker: Add selectable destination and timing offset options

danlafeir:
This PR does two things: 
1. Add two optionally consumed features. Without any change, this app will continue to function as-is for current users
2. Refactor code structure, naming, caching utility and remove unnecessary API token usage* (You can elect to merge without this commit and the features will be added without issue).

I tested this using `pixlet serve`, disabled some caching and my own CTA Arrivals API token. Everything worked on my machine... and I believe it will work on yours. I tested each commit individually incase the refactor was not preferable.

I developed this independently but pivoted to this contribution. I wanted the Tidbyt to help with my commute to work. I live a 7 minute walk from the train station, so knowing the next two trains are arriving now and in 2 minutes does not help.

I am happy to independently publish an app if that would be easier or preferable (but there are already three-ish :P)

*I was able to hit `https://data.cityofchicago.org/resource/8pix-ypme.json` without an `app_token` so I removed it. It might be required but I was unsure how to locate whether it is required.

### DIFF
--- a/apps/ctaltracker/cta_l_tracker.star
+++ b/apps/ctaltracker/cta_l_tracker.star
@@ -5,21 +5,17 @@ Description: Shows the next two arriving CTA "L" Trains for a selected station. 
 Author: samshapiro13
 """
 
-load("cache.star", "cache")
-load("encoding/json.star", "json")
 load("http.star", "http")
 load("render.star", "render")
 load("schema.star", "schema")
 load("secret.star", "secret")
 load("time.star", "time")
 
+##### constants #####
+
 CTA_STATIONS_URL = "https://data.cityofchicago.org/resource/8pix-ypme.json"
 CTA_ARRIVALS_URL = "https://lapi.transitchicago.com/api/1.0/ttarrivals.aspx"
 
-L_STOPS_CACHE_KEY = "lstops"
-ARRIVALS_CACHE_KEY_PREFIX = "arrivals"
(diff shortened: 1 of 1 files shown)


danlafeir:

danlafeir:
recheck

apps/ctaltracker/cta_l_tracker.star
```diff
@@ -35,245 +31,217 @@ COLOR_MAP = {
...
+    ) for destination in DESTINATION_STATIONS]
 
     return schema.Schema(
         version = "1",
         fields = [
             schema.Dropdown(
                 id = "station",
                 name = "Departing Station",
-                desc = "The CTA \"L\" Station to get departure schedule for.",
+                desc = "The CTA \"L\" Station to you would depart from",
```
matslina, line 98:
```suggestion
                desc = "The CTA \"L\" Station you would depart from",
```

matslina:
@samshapiro13 Thoughts?

danlafeir:
Phoning @samshapiro13. Your car is illegally parked.


Psych, but while you're here do you have any thoughts?

samshapiro13:
These additions sound good to me!

danlafeir:
@matslina I believe I added your text changes into the PR and it should be ready to merge whenever you get a chance.

PR #2845, opened 2024-10-30
Update zmanim.star date and new times

shimonsavitsky:
adding in current date and ability to handle edge case prayer times around sabbath and holiday


shimonsavitsky:
Ready for review whenever @matslina

PR #2853, opened 2024-11-03
Simplify app description to explain what kind of google sheet it accepts

vipulchhajer:
Instead of displaying url to a sample google sheet, explaining the type of google sheet that's needed. Easier for users to understand.


matslina:
Thanks!
Note that this PR is changing two apps. In general, we prefer having separate PRs for separate apps.

vipulchhajer:
Thanks for the reminder Mats. My bad, will keep that in mind moving forward.

PR #2863, opened 2024-11-06
Added Victron app

jduncc:
Hello,

This is a new app that integrates with the Victron VRM API to show energy data for a given Victron installation.  API key is provided in the config to render actual data.  Here is a sample:
victron


jduncc:

PR #2868, opened 2024-11-07
Hanukkah App Quick Update for 2024

BryanSlavinBryter:
Including @jvivona here - this is the quick update for 2024.  It is the same as last year's app, just adapted for 2024's start and end dates.  

Once approved, this can go through on the holiday promo you had planned.

I plan to work on another release between now and Hanukkah with some enhancements. Stay tuned!


apps/hanukkah/hanukkah.star
```diff
@@ -1,166 +1,66 @@
...
+    }
+    hanukkah_first_day = time.parse_time(hanukkah_dates[year])
+    hanukkah_last_day = hanukkah_first_day + time.parse_duration("192h")
+    return hanukkah_first_day, hanukkah_last_day
 
 def main(config):
-    timezone = config.get("timezone") or "US/Eastern"
-    if (not time.is_valid_timezone(timezone)):
-        timezone = "US/Eastern"
+    timezone = config.get("timezone") or "America/New_York"
```
matslina, line 28: Are you looking for the `$tz` parameter here? It's by default set to the timezone of the device for which the app is run. Otherwise, a location object will have a `timezone` field, but that's not what's being used here.

matslina (changes requested):

danielsitnik (approved):
Author implemented suggestions.

PR #2869, opened 2024-11-08
Fix Thanksgiving day math for 2024

matthewfischer:
Also add a 2025 date. App was using the date of the 22nd previously which is incorrect


matthewfischer:
FYI @rohansingh  the original fix failed to copy in the variable you added to the calc part. I think this fix works.

Also FYI @jalexcooney

PR #2876, opened 2024-11-11
New app: Sidereal Time

danielsitnik:
This is a [Sidereal Time](https://en.wikipedia.org/wiki/Sidereal_time) clock that displays the current local sidereal time (LST) based on the user's location.

There are options to display the time using a blinking colon, to use 24h notation and to also display the local (standard) time along with the sidereal time. All colors are also configurable.

Preview:
sidereal_time


danielsitnik:
@matslina could you please approve this one? Been waiting for 12 days now and it's a simple clock. Everything runs locally, no http module involved.

On a side note, if you and @rohansingh are ok with it, I'm sure some of the more senior community developers wouldn't mind lending a hand with these approvals.

Thanks!

PR #2879, opened 2024-11-13
New app: qlock

craigerskine:
**qlock** - custom clock with time, binary, beats, and date; all with a custom font

screenshot


craigerskine:

danielsitnik:
@craigerskine very cool clock!
But it's showing the time in UTC, instead of the user's location.

You can get the device's timezone with `tz = config.str("$tz", "UTC")`.
And then convert the current time to it with `now = time.time().in_location(tz)`.

Another nice addition would be to add a Schema option to allow the user to choose how the date should be displayed: "Month - Day" or "Day - Month". 😉

craigerskine:
@danielsitnik Thanks.

I used the official clock app example as a base to get the timezone. Maybe I did it wrong. I wish there was a way to push the app to the TidByt for testing. It works on my local dev.

PR #2884, opened 2024-11-19
Add MealViewer

jed:
This app uses the [MealViewer](https://schools.mealviewer.com) API to display what's next being served for lunch at a participating school.

mealviewer


rohansingh:
For caching of HTTP requests, please pass `ttl_seconds` to `http.get` instead of using the cache directly.

PR #2892, opened 2024-11-27
Update Thanksgiving day

savdagod:
- Display "day" instead of "days" when there is one day left
- Display "Happy Turkey Day!" on Thanksgiving day


matslina:
@jalexcooney Cool to merge this?

PR #2886, opened 2024-11-23
Add 10 new mountains to Ski Report 

cmorrisseau77:
Adds Big Sky, Brighton, Cypress, Summit, Boyne Mountain, The Highlands, Loon, Pleasant Mountain, Sugarloaf and Sunday River to Ski Report through an additional API

PR #2898, opened 2024-12-09
Support metric units for ski-report

CoatedMoose:
Defaults to Fahrenheit for temperature, and inches for snowfall, but allows switching to Celsius and/or centimeters.

Additionally, add the option to hide the snowfall duration hint.

### DIFF
--- a/apps/skireport/ski_report.star
+++ b/apps/skireport/ski_report.star
@@ -73,10 +73,31 @@ def get_schema():
                 id = "resort",
                 name = "Ski Resort",
                 icon = "mountain",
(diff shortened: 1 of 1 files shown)


CoatedMoose:
The failing check is on the duration of the run. I don't think there is anything that can be done about this. After running the profiling, the slow spot is a network call that regularly takes > 1s. (Also, if I checkout the version without my changes, the check fails with the same issue.

What can be done about this?

CoatedMoose:
Will conflict with https://github.com/tidbyt/community/pull/2886. I'll follow up and update this if the other gets merged first.

CoatedMoose:
Closing this is a mistake. Please reopen and review.

PR #2899, opened 2024-12-09
Team Fortress Key Price

FrontierProphet:
no idea what Im doing, hope this works


theimpossibleleap:
Hello again, @FrontierProphet. As you can see, the application is not passing the Build and Test / check pipeline. A best practice, before pushing to GH and creating a PR, is to run `pixlet check`. I cloned your repo and, while it works, I ran a check and it shows the following errors:

```
✖ tf_key_price.star
  ▪️ Problem: app is not formatted correctly: formatting returned non-zero exit status: 4
  ▪️ Solution: try `pixlet format tf_key_price.star`
✖ tf_key_price.star
  ▪️ Problem: app has lint warnings: linting failed with exit code: 4
  ▪️ Solution: try `pixlet lint --fix tf_key_price.star`
✔️ tf_key_price.star
```

These are quite easy to resolve by running the commands in the solution: `pixlet format <APP_NAME>` and `pixlet lint --fix <APP_NAME>`. You can read some about the formatting [here](https://tidbyt.dev/docs/publish/FAQ#why-is-the-formatter-so-opinionated). When running the lint check, it shows an error on line 11 with the the parameter / variable `config` not being used. I believe this is because your app is not using a schema for in-app fields that users can configure, as shown [here](https://tidbyt.dev/docs/reference/schema). I removed `config` from `def main()`, ran lint again, and it worked. I then ran another `pixlet check` and it passed. 

I would try these, it should resolve the issues you're having. But the App works, congrats on that! 

Hope this helps.

FrontierProphet:
is this 30 line program actaully taking that long to locally render? could this be an api accessibility/availability issue?

jvivona:
if you do a pixlet profile you can get a feel for the timing in each area of the code.   Similar to this:  
image

FrontierProphet:
image
running locally the program seems quick?

PR #2922, opened 2025-01-28
Norway transit app

matsgrosvik:
This is an app for Norway's public transport. Search for your stop, and get the next departures. Choose between directions and different types of public transport. Everything from boats to buses.

image


matsgrosvik:

matsgrosvik:
recheck

PR #2926, opened 2025-02-05
New App: NYC ASP Status

A-Wozzie:
This app displays the current status of New York City alternate side parking (street cleaning) rules, which may be suspended due to inclement weather (such as snow removal operations) or because of observed legal or religious holidays.

screenshot

- The street cleaning logo on the left is red when ASP rules are in effect, and turns green when rules are suspended or not in effect.
- The following day's status is shown after 3PM EST/DST.
- The app can be configured to only be shown when alternate side parking rules are suspended or not in effect.

Data is sourced from the NYC 311 API (https://api-portal.nyc.gov/)


A-Wozzie:

apps/nycaspstatus/nyc_asp_status.star
```diff
@@ -0,0 +1,117 @@
...
+load("http.star", "http")
+load("render.star", "render")
+load("schema.star", "schema")
+load("time.star", "time")
+
+months = ["", "Jan.", "Feb.", "Mar.", "Apr.", "May", "Jun.", "Jul.", "Aug.", "Sep.", "Oct.", "Nov.", "Dec."]
+days = ["", "1st", "2nd", "3rd", "4th", "5th", "6th", "7th", "8th", "9th", "10th", "11th", "12th", "13th", "14th", "15th", "16th", "17th", "18th", "19th", "20th", "21st", "22nd", "23rd", "24th", "25th", "26th", "27th", "28th", "29th", "30th", "31st"]
+
+URL = "https://api.nyc.gov/public/api/GetCalendar?"
+API_KEY = "5e5d2e9e95d64fcfa0be86ac934ae41f"
```
drudge, line 18: Since this is an API key does it make sense to encrypt it with the secrets module so it's not here in plain text?

danielsitnik:
@A-Wozzie we're waiting for your input here so this can be approved.

A-Wozzie:
Good catch @drudge, thanks. I've implemented the encrypted key. Should work when deployed.

drudge:
> Good catch @drudge, thanks. I've implemented the encrypted key. Should work when deployed, although it fails the check here for returning NoneType.

Thanks, @A-Wozzie. The check is failing here because when shown in an app preview in the mobile app, no secret module is used so you'll get None back. You should fallback to an empty string or show a preview of what the app could look like if no API key is entered. 

Whatever you render when no config data is passed is what will show for users adding your app from the mobile app or viewing the available apps on the website.

A-Wozzie:
> Thanks, @A-Wozzie. The check is failing here because when shown in an app preview in the mobile app, no secret module is used so you'll get None back. You should fallback to an empty string or show a preview of what the app could look like if no API key is entered.
> 
> Whatever you render when no config data is passed is what will show for users adding your app from the mobile app or viewing the available apps on the website.

That makes sense, thanks for the explanation! Added a demo render for whenever the secret module isn't used. Should all be working now.

PR #2932, opened 2025-02-18
indexticker: fix rendering of 0.0x%

Denton-L:
Any number that's 0.0x%, for example 0.09%, ends up rendering out as `0.9%`. Fix this by using the humanize package's float() to do the hard work for us.

### DIFF
--- a/apps/indexticker/index_ticker.star
+++ b/apps/indexticker/index_ticker.star
@@ -12,7 +12,7 @@ v1.3 - Added NIKKEI, Europe 350, Global 100, Global 1200, NZX50 indices; support
 load("animation.star", "animation")
 load("encoding/json.star", "json")
 load("http.star", "http")
-load("math.star", "math")
+load("humanize.star", "humanize")
 load("render.star", "render")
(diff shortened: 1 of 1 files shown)


drudge (approved):

drudge:
@Denton-L On check I'm seeing a 429 error:

  Error in fail: fail: request to https://query1.finance.yahoo.com/v8/finance/chart/%!E(MISSING)axjo?metrics=high?&interval=5m&range=1d failed with status code: 429 - Edge: Too Many Requests

Denton-L:
From what I can tell, this uses the Yahoo Finance API which isn't really an official API so the rate limit is undocumented. Maybe we should bump the cache time from 60 -> 120 s? Or just retry until it passes lol

Looks like the MISSING thing is just purely a logging bug, the actual request should still be the entity code.

If it doesn't work after a couple retries, we should ask the original app author about increasing the cache time.

Denton-L:
Thanks for getting the community repo moving again btw!

PR #2940, opened 2025-03-03
Coffee app

rjison:


rjison:
Thanks for merging my PR. Nice to see this project is still active! 

My app isn't available in the Tidbyt app yet though, is there another step that needs to happen?

PR #2944, opened 2025-03-04
FIX: add a user-agent to prevent rate limiting

danielsitnik:


Denton-L:
This undoes the formatting fix that I put in earlier, was this intentional?

PR #2935, opened 2025-02-24
Initial implementation of anime next ep

brianmakesthings:
Create a Tidbyt app that shows when the next episode of an anime is airing by leveraging the AniList api

image

Implementation should follow the anilist usage guidelines

[AniList Rate Limiting](https://docs.anilist.co/guide/rate-limiting)

> The API is currently in a degraded state and is limited to 30 requests per minute. This is a temporary measure until the API is fully restored.

[AniLIst terms of use](https://docs.anilist.co/guide/terms-of-use)

> 1. Free for non-commercial usage. See our [commercial usage](https://docs.anilist.co/guide/terms-of-use#commercial-usage) section for more.
> 2. Using the AniList API as a backup or data storage service is strictly prohibited.
> 3. Hoarding or mass collection of data from the AniList API is strictly prohibited.
> 4. Applications or services must comply with our [naming guidelines](https://docs.anilist.co/guide/terms-of-use#naming-guidelines).
> 5. Prohibited from use within competing noncomplementary services of the same nature. This includes, but is not limited to Anime/Manga list/tracker services. Competing services may be authorized on request if they provide significant sustained integration/syncing with the AniList API and AniList user accounts.

---

#### Finished airing
image

#### Unknown ID
image

#### Next episode unknown

image

---

I don't think there's anything too novel about this app, maybe in the future an anilist user id can be passed in so we automatically show all the shows a user is currently watching rather than having to look up an id

PR #2945, opened 2025-03-05
[animeNextEpisode] Search anime by name instead of ID

brianmakesthings:
original PR https://github.com/tidbyt/community/pull/2935

As suggested by @danielsitnik instead of searching for anime by ID, search by name

### Screenshot

image

Note: [pixlet typeahead](https://github.com/tidbyt/pixlet/blob/c29059778792f1731822098aeb71f9bf5567b38b/src/features/schema/fields/Typeahead.jsx#L42-L54) is slightly broken in that if the result is not a substring of the options, then it won't be shown.

ex. if I search "Solo Leveling" the api will return "Ore dake Level Up na Ken" but there no options are shown because the Autocomplete component from mui filters this out.

I hope this is not the case in the mobile app

### DIFF
--- a/apps/anime_next_ep/anime_next_ep.star
+++ b/apps/anime_next_ep/anime_next_ep.star
@@ -11,15 +11,19 @@ load("re.star", "re")
 load("render.star", "render")
 load("schema.star", "schema")
 load("time.star", "time")
+load("encoding/json.star", "json")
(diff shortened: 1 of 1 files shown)


danielsitnik:
Great work @brianmakesthings!

PR #2950, opened 2025-03-06
indexticker: fix rendering of 0.0x%

Denton-L:
Any number that's 0.0x%, for example 0.09%, ends up rendering out as `0.9%`. Fix this by using the humanize package's float() to do the hard work for us.


Denton-L:
@danielsitnik in case the portion of #2944 reverting the formatting change was unintentional, this PR restores the formatting fix that was merged in #2932

danielsitnik:
@Denton-L purely unintentional. When we approved #2932 it blocked the Github Actions from deploying all other PRs because indexticker was being rate limited by Yahoo when the pipeline was running `pixlet check` against it.

So we just made a small change to add a User-Agent in there, not sure why it was created against a version without your changes.

Anyway, approving again.

PR #2953, opened 2025-03-09
Add FRC ranking app to community apps

dragid10:
This FRC ranking app will display the current ranking position of a team in the FRC competition at their active competion. The app will display the team avatar (if an avatar was provided to FIRST), team number, team name, and ranking position.

If the team is actively participating at an event, then the app will display their current rank at that event. If the team isn't actively participating at an event, then it will show "No Ranking" (because there's no ranking number if there is no event to be ranked at)

Renders:
Team 6908 (my team):
frc_team_rank-6908

Team 4829:
frc_team_rank-4829

### DIFF
--- a/apps/frcteamrank/frc_team_rank.star
+++ b/apps/frcteamrank/frc_team_rank.star
@@ -1,0 +1,345 @@
+"""
+Applet: FRC Team Rank
+Summary: Display FRC event ranking
+Description: Displays the ranking of an FRC team at the team's active event.
+Author: dragid10
+"""
+
+# Imports
+# load("encoding/base64.star", "base64")
+load("http.star", "http")
+load("humanize.star", "humanize")
+load("render.star", "render")
+load("schema.star", "schema")
+load("time.star", "time")
+
+## ==================== BEGIN CONSTANTS ====================
+# Cache and API settings
+TEAM_INFO_CACHE_SECS = 3600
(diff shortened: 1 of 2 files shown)


dragid10:
Since the app requires an api key, I think the build is going to fail. I added the encrypted API key to the code, but I'm not sure what else is required to allow the build to access it

dragid10:
@drudge I'm wondering what the next steps are for me with the encrypted secret

drudge:
> @drudge I'm wondering what the next steps are for me with the encrypted secret

If you can, please handle the absence of an API key and show some form of sample/preview data. This is what will show up in the mobile app when browsing for new apps.

dragid10:
> > @drudge I'm wondering what the next steps are for me with the encrypted secret
> 
> If you can, please handle the absence of an API key and show some form of sample/preview data. This is what will show up in the mobile app when browsing for new apps.

Thanks for the feedback, I've updated the app to show a default / demo view if no information is entered! 
frc_team_rank

dragid10:
Hey @drudge does the above change make it more app store friendly?

dragid10:
Out of curiosity, what types of things are you looking for in terms of http.star usage? Is there something dangerous about it?

dragid10:
@drudge sorry to ping again, but I'm wondering if there is anything else I need to do to submit the app to the store

PR #2957, opened 2025-03-10
Relay name change

radiocolin:
Fixing more references to the old "Relay.fm" name


radiocolin:
I'm not sure why those two old commits show as part of this PR... they aren't included in this branch. This only edits the Relay app.

radiocolin:
Ignore... will resubmit

PR #2963, opened 2025-03-11
[githubbadge] Fix error handling and use the HTTP cache

IngmarStein:
Before this change, the app stored an empty array in the cache in case `fetch_workflow_data` failed (for example for non-existing repositories). Subsequent requests would read the empty list, but not the error state and thus display the default `tidbyt/pixlet` repo.


IngmarStein:
@Cavallando (original author): does this look good to you?

Cavallando:
@IngmarStein thanks for the PR! This looks great to me!

PR #2964, opened 2025-03-12
[worldcapitals] display translated country names

IngmarStein:
- display translated country names
- add a setting to change how long a country is displayed


drudge:
@joshspicer Looks good to me. Thoughts?

PR #2968, opened 2025-03-15
Update PiHole app to support both v5 and v6 API versions

siva801:
## Changes Made
- Added support for both PiHole v5 and v6 API versions through version selection dropdown
- Implemented separate API handling functions for each version:
  - `get_pihole_stats()` for v5 using `/admin/api.php` endpoints
  - `get_pihole_v6_stats()` for v6 using session-based authentication and newer REST API endpoints
- Updated schema to include version selection dropdown
- Maintained backward compatibility with existing v5 installations

## API Details
- V5: Uses token-based auth with `auth` parameter
  - `/admin/api.php?summaryRaw` for statistics
  - `/admin/api.php?overTimeData10mins` for graph data
- V6: Uses session-based auth with `/api/auth` endpoint
  - `/api/stats/summary` for statistics
  - `/api/history` for graph data

## Testing
- Tested with both PiHole v5 and v6 instances
- Verified graph plotting works correctly with both API versions
- Confirmed statistics display is consistent between versions

## Notes
Users need to select their PiHole version in the app configuration. Defaults to v5 for compatibility with existing installations.

### DIFF
--- a/apps/pihole/pihole.star
+++ b/apps/pihole/pihole.star
@@ -13,27 +13,65 @@ load("schema.star", "schema")
 
 HOST = ""
 API_KEY = ""
+VERSION = "V5"
 GREEN = "#00cc00"
 RED = "#ff4136"
 PIHOLE_LOGO = base64.decode("""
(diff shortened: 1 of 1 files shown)


siva801:

PR #2974, opened 2025-03-18
Update PiHole to add support for HTTPS schema in hostnames

siva801:
This pull request enhances the Pi-hole app by adding support for HTTPS schema in hostnames. Previously, the app only supported HTTP schema. Backward compatibility is maintained by appending `http://` if no schema is detected in the hostname.


siva801:

PR #2979, opened 2025-03-22
Fix NYC ASP Status App

A-Wozzie:
NYC ASP Status app is not rendering on the community hub, nor in the Tidbyt cloud. This is probably caused by a mismatched app id, when I originally uploaded the manifest and .star files.

So, I created a new pixlet app and generated a new encrypted key using its id. Everything should match now.

I updated the manifest to reflect the new id, and updated the .star file with the updated encrypted key.

### DIFF
--- a/apps/nycaspstatus/manifest.yaml
+++ b/apps/nycaspstatus/manifest.yaml
@@ -1,5 +1,5 @@
 ---
-id: nyc-asp-status
+id: nycaspstatus
 name: NYC ASP Status
 summary: NYC alt side parking status
 desc: Displays whether New York City alternate side parking (street cleaning) rules are in effect today (or next day after 3PM).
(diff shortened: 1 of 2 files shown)


apps/nycaspstatus/manifest.yaml
```diff
@@ -1,6 +1,6 @@
 ---
-id: nycaspstatus
+id: nyc-asp-status
```
drudge, line 2: Since we've already published the app we don't want to change the ID.

drudge (changes requested):

apps/nycaspstatus/manifest.yaml
A-Wozzie: Got it, I reverted the app id in the manifest to match what's in the community repo. Also updated the encrypted key in the .star file to match this.

The app preview in the community hub shows an offline icon. The app itself doesn't render at all when added. It should render with a demo message when the secrets module isn't used, which it does when running locally.

IMG_4466

drudge (approved):

drudge:
Looks good now!

A-Wozzie:
Wonderful, thanks!

PR #2980, opened 2025-03-22
Update verge_taglines to correctly select text after site update

joevgreathead:
### What is this?

This is a bug fix to the Verge taglines app which recently started returning the default text ("THE VERGE") after [The Verge](https://theverge.com) updated their page layout and structure again so that the graphql endpoint previously used was no longer functional.

### How?

This changes the app back to a previous implementation method which searches the html content since the tagline is now, once again, present in the base HTML returned.

### Tasks

- [x] I have run `pixlet check` to verify my app passes.


joevgreathead:
Previously signed the CLA and build is green. Just awaiting a review.

PR #2996, opened 2025-04-02
MLB standings: Update abbreviation for Athletics

jmanske:
The Oakland A's are now without a city and ESPN updated their abbreviation. Fixed the dictionary to handle it.


drudge:
cc @cmarkham20 for visiblity

PR #3016, opened 2025-04-19
adding LAX Departures app

conor-mclaughlin-br:
In this PR, we add a new app called **LAX Departures** which shows the next six departing flights from LAX.

lax_departures

The data is shown across two screens, with three rows of data per screen.

- [x] Linted
- [x] Formatted
- [x] Checked

Screenshot from 2025-04-18 22-45-17


rjison:
Pretty cool. Think it would be possible to have a setting to pick your airport instead of just showing LAX?

conor-mclaughlin-br:
> Pretty cool. Think it would be possible to have a setting to pick your airport instead of just showing LAX?

I would love to do so! But have some API cost considerations for fetching schedules on the backend that have constrained me to just the one airport right now... perhaps if I can spend some time improving my caching layer, then this could be extended to flexible airport selection

rjison:
Just a thought, you could have a setting asking if they have their own api key. If they do, then you could trigger a list of supported cities. People with their own key wouldn't affect you sharing data with the LAX feed.

PR #3017, opened 2025-04-19
New App: Should I Deploy?

adaviscourt:
shouldideploy
shouldideploy
shouldideploy
shouldideploy

After developing this, I realized there is an [existing app](https://github.com/tidbyt/community/blob/main/apps/shouldideploy/manifest.yaml), but I'd still like to add this one. It adds some flair and configurability.

- Location will be based on `schema.Location` rather than input, to ensure it matches what the shouldideploy.today is expecting (has error handling too just in case).
- You can choose:
    - symbols: black background with ✅ or ❌
    - thumbs: red/green background with 👍 or 👎 

Passes checks:
image

### DIFF
--- a/apps/shouldideploytoday/manifest.yaml
+++ b/apps/shouldideploytoday/manifest.yaml
@@ -1,0 +1,7 @@
+---
+id: shouldideploy
+name: Should I Deploy?
+summary: The developers dilemma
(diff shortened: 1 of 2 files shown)


adaviscourt:

PR #3021, opened 2025-04-23
Weatherbard app

mgtkach:
This app takes the weather condition from openweatherapp and returns one or poetic lines that describe that condition.  It will randomly choose from many.


mgtkach:

danielsitnik:
@mgtkach this app has an OpenWeatherMap API key stored in plain text inside the file. You should be using Pixlet's [Secret Module](https://tidbyt.dev/docs/reference/modules#pixlet-module-secret) for this.

Also, I noticed that the http call to OpenWeatherMap is not using the [Cache Module](https://tidbyt.dev/docs/reference/modules#starlib-http). If the app is installed by many users this has the potential to hit the API rate limits.

mgtkach:
Issues have been addressed - thank you

mgtkach:
@danielsitnik Thank you. I have made the corrections.  I cannot seem to get the "waiting on author" tag to go away.

PR #3022, opened 2025-04-23
SlotMachineClock: Support for different timezones

Rob-Keys:
Small change but important, the app now gets the timezone from the config to adjust the time shown.


danielsitnik:
@Rob-Keys I'm not sure this is going to work.

Usually we read the current timezone from the special config `$tz`, and the default value can also be passed to the function as the second argument. So the code should actually be:
```python
timezone = config.get("$tz", "America/New_York")
```

danielsitnik:
Yes, sadly it has not been documented. :(
There's a PR (https://github.com/tidbyt/tidbyt.dev/pull/620) open to fix it since last year but it hasn't been merged by the Tidbyt team.

PR #3023, opened 2025-04-23
Bug fix: German Transit

kmartinez834:
Bug fix:
- Description of bug: No suggested stations appeared when location was submitted on live app
- Fix: Order of coordinates in the `name_sf` param of `fetch_stations` were switched

Improvement:
- Added default station so preview includes actual upcoming departures, rather than "No station selected"


drudge (changes requested):
Linter is complaining

drudge (approved):

PR #3026, opened 2025-04-26
Add Retrograde Planet app with clean history

tidbytdev:
Display when planets change from retrograde to direct and vice versa and current status. Green color indicates the planet is direct. Red color indicates the planet is current. This shows up as D or R at the bottom of the screen. In addition, the app also shows three dates. The most recent date in the past when it switched, and the next two dates in future when it will switch again. If a date is green, the planet went/will go direct on that date. If the date is red, the planet went/will go retrograde on that date. Finally, the app shows the name of the planet in the left bottom corner and shows a visual representation of the planet on the right using base64 images.


tidbytdev:

tidbytdev:
recheck

tidbytdev:
updated.

drudge (approved):

PR #3036, opened 2025-05-07
emojilingo: support for MS emojis + defaulting to English + more inline comments

cedricsam:
* adding support for Microsoft emojis (there are 1565 valid ones in English)
* changing default language to English (to make it more appealing to an international audience -- was Canadian French before)
* inline comments providing some context in the source code


cedricsam:
To spot check the MS emojis, you can change the [default vendor](https://github.com/cedricsam/tidbyt-community/blob/emojilingo-microsoft-emojis/apps/emojilingo/emoji_lingo.star#L20) locally to `microsoft` and let the randomizer pick emojis or pick one from the [full-emoji-list](https://unicode.org/emoji/charts/full-emoji-list.html) with their number (`№` / the first column) set for the variable `test_emoji` in `emoji_lingo.star`.

PR #3052, opened 2025-05-29
Improve Literature Clock

WinslowJosiah:
After looking at the code for Literature Clock, I found a few things that could be improved.

- The source of the data was [the author's fork](https://github.com/alyshakwok/literature-clock/) of a "Literature Clock" repo, and said fork doesn't appear to have been updated since then. ([The original repo](https://github.com/JohannesNE/literature-clock/) has had many updates, changes, and corrections in that time period.)
- The caching feature of the HTTP module was not yet used, even though it evidently existed at the time.
- An `offset_start` is set on the marquee, but not an `offset_end`, making the scrolling loop improperly.

This PR addresses all of these issues.

### DIFF
--- a/apps/literatureclock/literature_clock.star
+++ b/apps/literatureclock/literature_clock.star
@@ -5,8 +5,6 @@ Description: Displays the time using a quote from a piece of literature. Based o
 Author: Alysha Kwok
 """
 
-load("cache.star", "cache")
-load("encoding/json.star", "json")
 load("http.star", "http")
 load("random.star", "random")
 load("render.star", "render")


WinslowJosiah:
> @WinslowJosiah - There are some linting errors preventing this from merging.
> 
> @alyshakwok - As the original author, are you okay with these changes?

Noted. I had run `pixlet format`, but I forgot to run `pixlet lint --fix`. 😅 Should be okay now, it just removed two `load` commands which are now unnecessary.

alyshakwok:
> @WinslowJosiah - There are some linting errors preventing this from merging.
> 
> @alyshakwok - As the original author, are you okay with these changes?

I reviewed the changes and am okay with them. Thanks!

PR #3055, opened 2025-05-31
New app: PATH Times

DavidGoldman:
# Description

This implements a NJ PATH trains schedule, similar to the default/built in NY metro app. This uses the official PATH train website which publishes real time data as a simple JSON file - we only hit this API twice a minute.

# Demo

path_times


DavidGoldman:
recheck

PR #3078, opened 2025-07-01
New: lolstats - show League of Legends player data

thiagoss:
lol_stats

A new app that shows League of Legends player data. It contains 4 rows:

1. Shows player name and Rank
2. Shows total player stats: wins / losses and current League Points
3. Shows last match data: win/loss, champion played, KDA, Gold and Minions killed
4. Shows a list of W/L representing wins/losses on the last few matches

Configuration allows you to set the summoner name, tagline and region. If not set, it will pick from a small pool of famous players to display.

Requires setting a RIOT API key (free to get for development but only lasts a day). Still working through the process to obtain a RIOT production key API and will put a follow up PR with it in the project.

### DIFF
--- a/apps/lolstats/lol_stats.star
+++ b/apps/lolstats/lol_stats.star
@@ -29,8 +29,8 @@ DEFAULT_REGION = "NA1"
 ENCRYPTED_API_KEY = ""
 
 # standard components:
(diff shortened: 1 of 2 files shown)


thiagoss:

PR #3080, opened 2025-07-06
update noaa_buoy.star

tavdog:
pull from mobile page instead of rss feed.  Enable option for displaying wind swell (ground swell is default now). fix checkbox get logic.


tavdog:
don't merge yet, a few tweaks incoming.

PR #3089, opened 2025-07-22
feat: adding new app for basic visualcrossing sourced weather display

ryan-doucette:
I created this app so users can display weather data sourced from visualcrossing.

It simply displays the temperature in F and C (my wife and I use different units), the humidity, the current conditions, and a label based on the user's input. If the user does not input a valid address, it will display "Weather load error".

visualcrossing_weather


ryan-doucette:

PR #3104, opened 2025-08-14
USGS Water Level and Temp App

nalanbar:
This is a super basic app, written mostly with the help of Claude, to display USGS data. It polls the API once an hour, and shows a few different data codes - depth and temperature. The top marquee is the station name, the vertical column on the right calculates the high and low points for the past 30 days, and displays today's measurement as an X relative to those highs and lows.

This is my first attempt at submitting anything, anywhere, so please let me know what I should do better! (actually, this is my second attempt. I submitted the first pull request about ten minutes ago with the wrong GitHub username, because I am that new at this!)

usgs_water_levels


nalanbar:
recheck

PR #3105, opened 2025-08-14
Adding Hisui and Paldea regions

eric-pierce:


eric-pierce:
@drudge @forrestsyrett Author line updated

PR #3119, opened 2025-09-04
Add Pawcast - Dog Walking Weather App

JosephScharnweber:
A Tidbyt app that helps dog owners determine safe walking conditions based on current weather data. Pawcast displays temperature, humidity, heat index, and provides risk assessments for walking dogs in hot weather.

### DIFF
--- a/apps/pawcast/pawcast.star
+++ b/apps/pawcast/pawcast.star
@@ -1,41 +1,51 @@
 # Pawcast - safe dog-walk time from temp + humidity
 
-load("render.star", "render")
 load("math.star", "math")
+load("render.star", "render")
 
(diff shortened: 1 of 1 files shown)


JosephScharnweber:

danielsitnik:
Hey @JosephScharnweber, is this still a work in progress?

I'm asking because you have some `config.get` calls in the code, but you don't have the `get_schema()` function defined, so these calls aren't really doing anything.

From what I understood, the app doesn't use any APIs to get the weather info, the user should input the temperature and humidity in the configuration.. but since you don't have `get_schema` defined, it's always using the default value.

PR #3135, opened 2025-10-01
Fly The L: add handling for TBD opponents

jmanske:
When the playoffs are happening, some teams don't know who their next opponent is. Right now we crash on "key not found in dict" errors; this PR handles it more gracefully.


jmanske:
:shipit:

PR #3144, opened 2025-10-23
Weather Gear

rubencodes:
Inspired by the classic `jacket` project, expands to include info about whether an umbrella is needed, hydration reminders for hot days, shorts weather support, and snow info!

cold_dry_night
hot_day
hot_night
mild_day
mild_night
vcold_rain
vcold_snow
warm_day
warm_rain


rubencodes:
recheck

PR #3149, opened 2025-11-03
Fix metar

tmsteen:
Reopening #3129 as @avalys mentioned he would comment to see if this would push it through automation. 

Updated the metar app to utilize the new Data API from Aviation Weather. The old API was deprecated and fully decommissioned in Sept 2025.

This update also implements the HTTP module native caching mechanism. Tested with local render.

@avalys is the original author and may want to comment/approve this PR.


michaelsobota:
been waiting for this fix watching my tidbyt spin through empty METAR info for a year. LGTM!

PR #3155, opened 2025-11-26
added comments

Ujwal200707:
This script automatically checks all modified Pixlet apps and validates them using pixlet check.
It first cleans the TARGETS list, then loops through each app and runs the appropriate check.

Some apps take longer to render on a cold cache, so the script includes a list of runtime exceptions.
If a modified app is in this exception list, it uses a custom --max-render-time value.
Otherwise, it runs a normal Pixlet check.

If no apps were modified, the script exits immediately.

This ensures consistent, reliable validation of all updated apps in CI while allowing slower apps enough time to render safely.


Ujwal200707:
i want to sign on cla

Issue #3146, opened 2025-10-27
No updates to apps in over 30 days -- don't bother


tmsteen:
If you are thinking of updating an app, just don't. It has been over 30 days since they have merged an update and their Stale Bot is just autoclosing PRs at 30 days. 

It seems there is no longer any intention of updating the platform. Time to start looking at the community options available to keep this thing running.

tmsteen:
@avalys I opened a new PR (cannot reopen old one) if you want to give it a try. I will be attempting a Tronbyt build soon to overcome this obstacle.

tmsteen:
@danielsitnik appreciate the update. Any way to turn off the auto-closing of stale PRs until/if this is resolved?

drudge:
I've extended it to 120 days for now.

kaffolder7:
> [@avalys](https://github.com/avalys) I opened a new PR (cannot reopen old one) if you want to give it a try. I will be attempting a Tronbyt build soon to overcome this obstacle.

In case anyone is interested, I just finished some initial development on a Home Assistant app (aka. add-on) that wraps [Tronbyt Server](https://github.com/Tronbyt/server) locally. 🎉🚀

_I figured HA was a good place to host this as my Home Assistant device is always "on"._

[Tronbyt](https://github.com/Tronbyt) is pretty much an awesome, full ecosystem replacement for Tidbyt since it appears Tidbyt is now dead. 💀
Tronbyt Server provides a nice GUI that allows you to generate custom firmware to flash various ESP32/HUB75-based display hardware (e.g. Tidbyt Gen. 1 & 2, Raspberry Pi connected to a 64 x 32 matrix LED, etc.).

If you'd like to run this HA app, I’d suggest at least 512 MB RAM for your host device. _(For reference, I’m running it on a Home Assistant Green, and it’s been running smoothly.)_

Here’s the repo if you'd like to try it out (feedback is welcomed):
https://github.com/kaffolder7/ha-app-tronbyt-server

markgreinke:
Im just here to give the developers of Tronbyt props. Very active community.

[https://github.com/tronbyt](url)

PR #3180, opened 2026-02-20
Fix enphasesolar token refresh flow

jwlawrence:
## Problem

  The EnphaseSolar app's token refresh flow has several interacting bugs that prevent re-authentication when the cached access token expires, leaving the app in an unrecoverable error state:

  1. **`INIT_KEY` cached with no TTL** — The bootstrap block that seeds config tokens into cache is gated by an `INIT_KEY` flag cached permanently. Once set, it never runs again, so when token caches
  expire they are never re-seeded.

  2. **`check_response` crashes on 401** — `EXPIRE_MSG in response.json().get("details")` crashes when the `"details"` key is missing from the response, because the `in` operator fails on `None`.

  3. **`request_refresh_token` silently fails** — The function calls `render_msg()` on error but doesn't return the result. Callers have no way to know whether the refresh succeeded or failed.

  4. **Wrong HTTP method for token refresh** — The code uses `http.post()` but the Enphase v4 token endpoint expects GET with query parameters.

  5. **API functions crash when cache is empty** — `get_system_id` and `get_system_stats` read the access token from cache internally. When the cache is empty, `"Bearer " + None` crashes.

  6. **401 errors return "wait" instead of retrying** — On a 401 from an API call, the code refreshes the token but then returns a "wait for next call" message instead of retrying the request with the
  new token.

  ## Changes

  - **New `get_access_token` function** replaces `request_refresh_token`. Uses a cache-first pattern: returns the cached access token on hit, otherwise attempts a refresh using the cached refresh token
  (falling back to the config refresh token). Returns a status dict `{"status", "access_token", "error"}` so callers can handle failures. Uses `http.get()` to match the Enphase API spec.

  - **Hardened `check_response`** — Defaults the `"details"` field to empty string instead of `None`. Treats any 401 as an auth failure rather than checking for a specific message.

  - **API functions accept `access_token` as a parameter** instead of reading from cache internally, eliminating the `"Bearer " + None` crash.

  - **`main()` retries on 401** — When an API call returns 401, the cached access token is invalidated, `get_access_token` is called to refresh, and the API call is retried once with the new token.

  - **Removed `INIT_KEY`** — Cache seeding now uses a simple cache-miss check (both access and refresh token caches empty) instead of a permanent flag.

  ## Verification

  - `pixlet format` — no changes needed
  - `pixlet lint` — no warnings or errors
  - `pixlet render` — renders successfully, showing "Missing credential" message as expected when no config is provided
  - No changes to the schema or user-facing config fields; fully backwards compatible


jwlawrence:

jwlawrence:
Bump. Is there anyone I can ping to review this?

LukiLeu:
See here: [https://github.com/tidbyt/community/issues/3146#issuecomment-3482100171](https://github.com/tidbyt/community/issues/3146#issuecomment-3482100171)

jwlawrence:
So is Tidbyt effectively dead?

PR #3196, opened 2026-03-20
Bug-Fix: FRC Team Rank - Handle null qual data from The Blue Alliance API

dragid10:
## Summary
- The Blue Alliance API returns `"qual": null` when an event is active but qualification matches haven't started yet
- The existing null checks only verified key existence (`not in`), not whether the value was `None`, causing a crash on the `in` operator
- Added `== None` checks at each nesting level in `get_team_ranking()` so the app gracefully falls back to the default ranking message


dragid10:

dragid10:
Will this ever get merged?

PR #3198, opened 2026-03-27
Add F1 Driver Standings app

aadit-cpu:
The app will provide a view of the driver standings as the F1 races happen. You can configure how many drivers you want to see, how fas to scroll across drivers, and to avoid spoilers, you can delay refresh by unto a day so allow you to see recordings without knowing the results.


aadit-cpu:

PR #3221, opened 2026-06-06
Add LA Street Sweep app

ryantaylor16:
## LA Street Sweep

preview

A 64×32 app that counts down to your next Los Angeles street sweeping day. A cute side-view car is parked at the curb while the street-sweeper truck creeps closer each day (countdown chip goes green → amber → red); on sweep day the car is gone and the sweeper passes the freshly-cleaned spot.

### How it works
- The user sets only **weekday + week-pattern (1st&3rd / 2nd&4th / every week / non-posted) + time**, per side of the street (optional second side).
- Sweep dates come straight from the **City of LA's public Google Calendar feeds**. Those feeds are already holiday-adjusted and contain no phantom 5th-week dates, so resolving the next sweep is a pure feed filter — the earliest feed date on the user's weekday that isn't in the past. No address lookup, geocoding, or date arithmetic.
- Find your days: streets.lacity.gov/services/street-sweeping

### Checklist
- [x] `pixlet check` passes
- [x] `pixlet format` applied
- [x] App is in its own directory under `apps/` with a `manifest.yaml`
- [x] Caches the upstream feeds (12h TTL)
- [x] Includes an animated preview

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ryantaylor16:
Note for the reviewer on the `http.star` flag 👋

The app makes only read-only `http.get` calls to the **City of LA's public street-sweeping Google Calendar feeds** (iCal), e.g.:

`https://calendar.google.com/calendar/ical/<calendar-id>/public/basic.ics`

- Three public, unauthenticated feeds: 1st & 3rd weeks, 2nd & 4th weeks, and non-posted routes. No API keys, no user data sent.
- Each request is cached with `ttl_seconds = 43200` (12h); the feeds change rarely, so device-load on the upstream is minimal.
- The feeds are parsed for `DTSTART` dates only. They're already holiday-adjusted by the city, so resolving the next sweep is a pure filter (earliest feed date on the user's weekday that isn't in the past) — no scraping, geocoding, or address lookup.

Happy to adjust the cache TTL or anything else if you'd prefer different behavior. Thanks for the review!

PR #3223, opened 2026-06-11
Fix `carenewables` applet

jxxe:
The CSV URL currently in use doesn't exist anymore. I updated it with the new URL as used here: https://www.caiso.com/todays-outlook/supply


jxxe:

PR #3224, opened 2026-06-20
Add Civics Test app

hachibu:
## Civics Test

Unknown

Displays a random question and answer from the 128 USCIS civics test questions — great for studying for the U.S. citizenship test.

**Each render:**
1. Waving American flag animation intro
2. "QUESTION" title card
3. Question text (scrolls vertically if long)
4. "ANSWER" title card
5. Answer text

**Notes:**
- Question is randomized each render via LCG seeded from current timestamp
- Long text scrolls via `render.Marquee`; short text is vertically centered
- Flag animation uses a pixel-column sine wave with an integer lookup table (no math lib)
- No external API calls — all 128 Q&A pairs are bundled in the app


hachibu:

hachibu:
:shipit:

PR #3225, opened 2026-06-21
Several hassentitylist formatting fixes

joe714:
render_entity parsed the Home Assistant entity state with int(), which raises "invalid literal with base 10" for float-valued states (e.g. "0.23"), crashing the whole applet. Use float() instead, matching the parsing already used by the sibling hasssolar and hassgraph applets. count is only used numerically (sorting and color scaling), so a float value is fully compatible.

num_format inserted a thousands comma a fixed 3 chars from the end of the raw string, which mangled decimal values ("7.032" -> "7,.032"). Group only the integer part so decimals are preserved, and add a "Decimal places" setting: blank keeps the value as reported by Home Assistant, while a number rounds to that many places (both via humanize for correct thousands grouping).

### DIFF
--- a/apps/hassentitylist/hassentitylist.star
+++ b/apps/hassentitylist/hassentitylist.star
@@ -6,6 +6,7 @@ Author: James Woglom
 """
 
 load("http.star", "http")
+load("humanize.star", "humanize")
 load("render.star", "render")
 load("schema.star", "schema")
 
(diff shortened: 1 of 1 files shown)


joe714:
I hereby sign the CLA.

joe714:
I hereby sign the CLA

PR #3239, opened 2026-07-06
Refactor ESPN news fetching to use public API

travisressler60-alt:
Updated ESPN news fetching to use the public API instead of the frontend data layer, ensuring compatibility with changes in the data structure. Adjusted sports list and URL construction accordingly.


travisressler60-alt (commented):
I am just trying to find a way to fix the app... Could really use help please.

travisressler60-alt (commented):
Does this work?? I have no idea.

PR #3245, opened 2026-07-20
Add Federal Flag app

wise-guy07:
## What does this app do?

Federal Flag displays a pixel-art United States flag on a full-height flagpole. It automatically lowers the flag during active federal half-staff notices and lets users choose between a gentle wave and a still flag.

## Data source and caching

The app reads the public federal widget endpoint from HalfStaff.org and only treats responses with `type: federal` as active federal notices. Responses are cached for 15 minutes to keep request volume low. The rendered graphic has a 30-minute `max_age` to avoid displaying stale status indefinitely.

If the endpoint is unavailable, the app renders at full-staff instead of failing.

## Testing

- `pixlet check apps/federalflag` passes
- Tested both animated and still modes locally
- Tested the live animation on a physical Tidbyt


wise-guy07:

wise-guy07:
Preview GIFs attached (full staff). 
Federal-Flag-Still
Federal-Flag-Animated

PR #3250, opened 2026-07-31
qbittorrent: support HTTP 204 login responses

harababurel:
## Summary

- accept HTTP 204 as a successful qBittorrent login response
- preserve support for older versions that return `Ok.`
- continue rejecting unsuccessful login responses

qBittorrent now sends 204 when a successful WebAPI response contains no data. The app currently checks only for an `Ok.` response body, so valid logins are displayed as `Login failed :(`.

## Testing

- `pixlet check apps/qbittorrent/qbittorrent.star`


harababurel:

harababurel:
recheck

PR #3256, opened 2026-08-21
Fix HVV Departures: migrate from defunct transport.rest HVV API to Transitous

frank-dornberger:
## What's broken

`hvv_departures.star` has been fetching nothing but errors for a while: HVV shut off the HAFAS API that backed [transport.rest](https://transport.rest/)'s HVV-specific endpoint, so `v5.hvv.transport.rest` (hardcoded in this app) no longer resolves at all. The [`hvv-rest`](https://github.com/derhuerst/hvv-rest) project behind that endpoint is archived, its readme now reading *"Deprecated because HVV has shut off their HAFAS API."* There's no drop-in HVV-specific replacement — even Deutsche Bahn's own classic HAFAS API (a plausible fallback) has since been shut off too.

## The fix

Switches to [Transitous](https://transitous.org/), a free, no-API-key, community-run GTFS/GTFS-RT router that covers Hamburg (HVV) among many other regions. Verified live against the current API (S-Bahn, U-Bahn/Hochbahn, AKN, MetroBus/XpressBus, and ferries all confirmed working with real-time data).

This isn't a pure URL swap — a few things had to change:

- **Endpoints**: `locations`/`departures` → Transitous's `geocode`/`stoptimes`.
- **User-Agent**: added a descriptive one (`tidbyt-hvv-departures/2.0 (+link)`), since Transitous's [usage policy](https://transitous.org/api/) asks API consumers to identify themselves and to reach out if usage becomes non-trivial.
- **Line classification**: Transitous only exposes a coarse `mode` (SUBWAY, SUBURBAN, REGIONAL_RAIL, BUS, FERRY, ...), unlike the old API's dedicated per-line `product`. AKN and RB/RE all share `REGIONAL_RAIL`, and regular/Xpress buses both share `BUS`. Added client-side reclassification (`classify_stop_time`) based on line short name/agency, so the existing per-transport-type toggles keep working independently.
- **Dropped the optional "direction" filter** — Transitous's `stoptimes` endpoint has no equivalent parameter for filtering by a downstream stop, and there's no reasonable way to fake it client-side.
- **Station IDs changed namespace.** Old HVV station IDs were plain numbers (e.g. `"90"`); Transitous IDs look like `de-DELFI_de:02000:10950_G`. Existing installs will show a one-time "Please re-select your station in the app settings" message instead of a generic fetch error, since the old IDs can't be resolved anymore.

## Also fixed

A few related, previously-latent issues surfaced while testing against live data:

- The `DEFAULT_*_CONFIG` fallback dicts were missing keys their render functions read unconditionally, crashing for any line not already hardcoded in `LINE_CONFIG` — e.g. Hamburg's newly opened **S5** line.
- Ferries (and anything else `render_line_icon` doesn't have a dedicated icon for) crashed by returning `None` into a `render.Box` child slot instead of rendering an empty box.
- Ferries now get a real icon: HVV's actual trapezoid/keystone-shaped blue badge (`#0994c7`), sampled directly from [HVV's official ferry line 62 schedule PDF](https://www.hvv.de/resource/blob/146504/37274f2519032273e1bfd61dbc99ed89/hvv_linienfahrplan_F62.pdf), instead of a blank box.
- S-Bahn line colors and the line set itself were stale. As of HVV's 2025 network expansion, **S11/S21/S31 no longer exist** — those branches are now independent top-level lines **S5** (`#008abd`) and **S7** (`#cf7822`), each with their own official color, and S1 (`#0aa537`)/S2 (`#b71b43`)/S3 (`#652282`) were nudged slightly too. Updated `LINE_CONFIG` (and regenerated the small pill-shaped line badge images from these exact hex values) to match [HVV's current Schnellbahn/Regionalverkehr network map](https://www.hvv.de/resource/blob/120104/3ff58f395fa6e8d4818231b0573b0bc6/hvv_usar-plan.pdf).

## Testing

- `pixlet check apps/hvvdepartures/hvv_departures.star` passes.
- Rendered against live data for: default config (Hamburg Hbf), ferry-only, suburban-only (confirming S2/S3/S5's new colors), AKN-only (confirming RB/RE don't leak in despite sharing the same API `mode`), a legacy numeric station ID (confirming the new stopgap message), and "nothing selected" — all render correctly.
- Line colors/shapes were cross-checked pixel-by-pixel against HVV's official PDFs linked above, not guessed from memory.

cc @fxb — original author, would appreciate your eyes on this given you own the product direction here per the [contributing guide](https://tidbyt.dev/docs/publish/modifying-apps).


frank-dornberger:

fxb (commented):
Thanks for fixing both the API and outdated HVV lines :-) I didn't even notice as I currently don't have this in my Tidbyt rotation...

The PR looks good from my side!

Technically, using the Transitous API, we could cover a lot more than just Hamburg, but I'd say that is out of scope for now unless someone really wants it. Another option, which I use with Home Assistant, is to use the [official API](https://www.hvv.de/de/fahrplaene/abruf-fahrplaninfos/datenabruf), but that would require a user to manually request API credentials. Using Transitous makes it a lot easier to set up.

fxb:
@frank-dornberger You might actually want to consider going the [Tronbyt ](https://tronbyt.com/) route, as all active development has since moved there.

frank-dornberger:
@fxb That's my plan B if @drudge, @danielsitnik, or @matsgrosvik won't respond by end of week.

danielsitnik:
> @fxb That's my plan B if @drudge, @danielsitnik, or @matsgrosvik won't respond by end of week.

Sorry @fxb, the pipeline that published new and updated apps has stopped working ages ago. One of the certificates expired and sadly me and @drudge do not have access to the underlying infrastructure to try and fix it, we only have PR approval permission here on GitHub.

I contacted @rohansingh and @matslina a long time ago. Both of them said they would look into it but nothing happened.

There's nothing else we can do here, we are completely unable to publish new or updated apps. The best course of action is to do the migration to Tronbyt and then have it fixed and published there. :(

frank-dornberger:
Thanks for the update, @danielsitnik. I'll get it published then as a Tronbyt version and will run a server on y RasberryPi.

frank-dornberger:
Closing PR as it's unreasonable to expect this will move forward.

PR #3267, opened 2026-09-02
Fix Todoist apps for API v1

### LINKED ISSUE #3266: Fix Todoist apps for API v1 and current OAuth endpoints
## Problem

The `Todoist` and `Todoist Tasks` community apps still use the retired Todoist REST v2 task endpoint and old OAuth URLs. They also expect the legacy bare-array task response, so existing installations can show errors or malformed states now that Todoist API v1 returns paginated objects.

## Proposed fix

- Preserve both apps’ names, IDs, encrypted OAuth credentials, configuration fields, and existing feature scope.
- Fetch tasks from `https://api.todoist.com/api/v1/tasks/filter` using `query` and `limit=200`.
- Parse the current `{results, next_cursor}` response; show `200+` in the count app when another page exists.
- Keep the list app to one API page and sort that page by priority.
- Update authorization and token endpoints and include the redirect/grant parameters supplied by Tidbyt OAuth.
- Version cache keys so old REST v2 payloads cannot be reused.
- Correct empty, hidden-empty, unauthenticated, cached, and API-error rendering.
- Validate synthetic zero/one/many/200+/priority/long-text states with Pixlet and do not publish real Todoist content in preview artifacts.

Todoist API reference: https://developer.todoist.com/api/v1/

tlambrou:
Closes #3266

## Summary

- migrate `Todoist` and `Todoist Tasks` from retired REST v2 calls to `/api/v1/tasks/filter` with `query`, `limit=200`, and cursor-aware response parsing
- update Todoist authorization and token endpoints while preserving both apps’ encrypted legacy OAuth credentials and existing configuration scope
- version and hash cache keys, fix empty/hidden-empty/unauthenticated/API-error rendering, and keep the list app to one priority-sorted API page
- regenerate both preview GIFs using synthetic content only

## Validation

- `pixlet check apps/todoist apps/todoisttasks`
- synthetic Pixlet renders for zero, one, many, 200+, hidden-empty, unauthenticated, API-error, priority, and 500-character long-text states
- `git diff --check`


tlambrou:

PR #3271, opened 2026-09-06
Add GitHub Actions app

nrkramer:
## GitHub Actions

CI status for a set of repositories, one row per repo, readable at a glance
from across a room.

Green means the last workflow run passed, red means it failed, amber means a
run is in flight. The header carries the worst state across everything being
watched, so a single glance says whether anything needs attention, and `+N`
shows how many repos did not fit.

There is a `circleci` app and a `gitlabpipelinestatus` app, but nothing for
GitHub Actions, which is what prompted this.

### Configuration

- **GitHub token** — a fine-grained personal access token.
- **Repositories** — `schema.Generated`, sourced from the token: the app calls
  `/user/repos` and renders one checkbox per repository that token can reach,
  most recently pushed first. Nothing to type and nothing to misspell.
- **Other repositories** — optional free-text `owner/repo` list for anything
  past the 60-repo picker limit.
- **Only show problems** — hides repositories whose last run passed.

Only `Actions: Read-only` is required (plus the mandatory `Metadata`), and the
app never writes anything. Private repositories work.

### Rate limits

One request per watched repository cached for 60s, plus one request for the
repository list cached for an hour. With the display showing four rows and a
twelve-repo cap, this stays far inside the 5000 requests/hour budget.

### Display notes

- Rows are sorted worst-first so a failure is always visible.
- The owner is dropped from each row, and the owner's leading word is dropped
  from the repository name where it repeats it (`Acme-Corp/AcmeEngine` renders
  as `engine`). Without this, organization repos sharing a prefix all truncate
  to the same string at 64px and become indistinguishable.
- No scrolling. A vertical marquee was tried and removed: it leaves the display
  blank for part of its cycle, which defeats the purpose of a glanceable app.
- With no token configured the app renders a generic sample board for the store
  listing; with a token but nothing selected it says so rather than showing an
  empty screen.

### Checks

`pixlet check`, `pixlet lint`, `pixlet community validate-manifest` and
`pixlet community validate-icons` all pass. Rendered and run against real
public and private repositories, and displayed on hardware.


nrkramer:

PR #3276, opened 2026-09-16
p(doom) App

philweldon:
p(DOOM) counts down to local midnight as though midnight were the moment AI
  takes over. At midnight it shows `p(DOOM) = 1.00`, says SEE YOU TOMORROW, and
  starts again — the joke being that p(DOOM) happens every midnight, every day.

  Between countdowns it shows a real, attributed quotation from someone who has
  publicly discussed catastrophic or existential risk from AI. Nothing is
  invented: all fifteen live in `quotes.star` with the speaker, source, URL and
  date, and the excerpts shown on screen mark their cuts.

  Settings: location (for the timezone), quote frequency, and animation
  intensity. The countdown is baked into the animation at one second per four
  frames, so it ticks in real time for the whole render.

`pixlet check apps/pdoom` passes; renders in about 20 ms.


philweldon:

